PR #2, opened 2020-08-25
[bugfix] Fix parsing logic when handling empty/missing fields

andyndang:
* Add tests with data from Java library

### DIFF
--- a/src/whylogs/util/dsketch.py
+++ b/src/whylogs/util/dsketch.py
@@ -211,7 +211,7 @@ class FrequentItemsSketch:
         If x is an empty sketch, None is returned
         """
         if len(x) <= 8:
-            FrequentItemsSketch()
+            return FrequentItemsSketch()
         sketch = datasketches.frequent_strings_sketch.deserialize(x)
         return FrequentItemsSketch(sketch=sketch)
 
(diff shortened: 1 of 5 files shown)


ibackus (commented):
I'm not sure that this will work with dataset profiles serialized by the python code, unless there are previous changes I'm not aware of.  I believe we'll need to make those changes to all python writers.  I can test this.

If we are moving away from protobuf to delimited protobuf for serialization to disk, we should formalize that as our standard data format for all protobuf binary outputs.

I think there's probably value in centralizing this logic in utility functions for reading/writing delimited protobuf to avoid duplicate code for other protobuf classes.

andyndang:
@ibackus the Python serialized data set is still deserializable with `from_protobuf_string` method (no change here).

We should definitely standarize the format to delimited format, but that's the next step. This delimited logic only applies to the top level message - inner messages don't use delimited deserialization

src/whylogs/util/dsketch.py
```diff
@@ -211,7 +211,7 @@ def deserialize(x: bytes):
         If x is an empty sketch, None is returned
         """
         if len(x) <= 8:
-            FrequentItemsSketch()
+            return FrequentItemsSketch()
```
andyndang, line 214: @ibackus I think this was the bug in the existing code?

ibackus: That definitely seems to be a problem.

ibackus (dismissed):

andyndang:
@ibackus I added the delimited logic for DatasetProfiel (that's the only place we need delimited serialization)

PR #10, opened 2020-09-03
Ibackus/feature/docstrings

ibackus:
Add significant documentation to WhyLogs

## Changes

* Complete, reformat, and fix the doc strings.  Everything except the `cli` is well documented now.
* Update sphinx configuration
* Add code example gallery to docs using sphinx-gallery.

### Notes
* We cannot build the WhyLogs package in readthedocs, which therefore means we cannot run any code examples in sphinx-gallery.  Additionally, some internal auto-generated documentation examples may break (anything which calls/imports WhyLogs in sphinx to run code snippets)


docs/conf.py
```diff
@@ -22,21 +22,27 @@
 import sphinx_rtd_theme
 from sphinx.ext.autodoc import between
 
+
+# Add package paths to the python path to allow sphinx to import modules
 sys.path.insert(0, os.path.abspath("../src/"))
 
+import whylogs
```
andyndang: By importing WhyLogs here you require the doc build runtime to have all whylogs dependencies. This is breaking `readthedocs` build in Sphinx (I didn't wire it in correctly so it's not showing up in the CI but I'm working on fixing it)

A better way to do this is here: https://github.com/whylabs/whylogs-python/blob/mainline/setup.py#L103

The runtime for docs building is here: https://github.com/whylabs/whylogs-python/blob/mainline/docs/requirements.txt

andyndang (changes requested):

docs/conf.py
ibackus: Weird, it should find whylogs because of the line above where the whylogs path is added to the sys.path.  But I can wire something like that in.

.readthedocs.yml
```diff
@@ -16,8 +16,15 @@ sphinx:
...
 
+submodules:
+  include: all
+  recursive: true
+
 ## Optionally set the version of Python and requirements required to build your docs
 python:
   version: 3.7
   install:
+    - requirements: requirements-dev.txt
```
andyndang: I'm really uncomfortable with this. Building documentation shouldn't require installing the whole development dependency set. The reason you're doing this is because you're directly importing the package in the doc building step - which is antipattern. We should separate the responsibilities here

ibackus: I'm just messing around with readthedocs right now.  I'm trying to understand what can actually be done there and what can be built.

ibackus: Importantly, I was poking around to see if I could get readthedocs to import WhyLogs so we could run the example scripts to generate example documentation.  But I can't install protoc without more effort, so I'm scrapping that for now.

ibackus:
@dangitandy the docs are building, but with some minor issues.

A wonderful thing about sphinx is that we can dynamically generate content by calling python and interacting with the package.  This can be used to generate examples and capture the output (for example, the plotly docs: https://plotly.com/python/) and can be used to document dynamically generated content.

I've disabled executing the example scripts, but there are a fair amount of code snippets within the code which currently break.  I can disable those for now, but we are definitely limiting our documentation ability by not being able to interact with WhyLogs in the documentation building environment.

As I guess you've noticed, we cannot get WhyLogs working in the readthedocs environment due to the dependency on `protoc` for building WhyLogs.  I played around with installing `protoc` in the readthedocs containerized environment at build time, but without root permissions it wasn't going to be straightforward.

For now, I'll close this PR and re-open it once I've removed all code snippets.

I think we may want to explore building (and publishing) our docs natively in github.  Github pages will host the docs, and we could treat the doc building as another build stage.

andyndang:
> As I guess you've noticed, we cannot get WhyLogs working in the readthedocs environment due to the dependency on protoc for building WhyLogs. I played around with installing protoc in the readthedocs containerized environment at build time, but without root permissions it wasn't going to be straightforward.

The the Plotly example you showed (https://github.com/plotly/plotly.py/blob/doc-prod/doc/apidoc/conf.py) actually uses sphinx. They pull plotly into the doc via requirements.txt (so separate package building from documentation building) - which I guess is fine but will require some monitoring and coordination.

ibackus:
That seems like a reasonable approach.

andyndang (approved):
Superb!

PR #39, opened 2020-09-19
Fix bug where numpy dtypes were not recognized as numbers

ibackus:
This manifested as numpy int64 values NOT being treated as numbers,
therefore not being tracked, and therefore not generating quantiles.

### DIFF
--- a/tests/helpers/testutil.py
+++ b/tests/helpers/testutil.py
@@ -1,3 +1,6 @@
+TEST_BUCKET = "whylabs-isaac"
+
+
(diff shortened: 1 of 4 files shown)


tests/helpers/testutil.py
```diff
@@ -1,3 +1,6 @@
+TEST_BUCKET = "whylabs-isaac"
```
andyndang: intentional commit?

andyndang (approved):

PR #68, opened 2020-10-19
Enable MLFlow integration with whylogs for batch mode

### LINKED ISSUE #67: MLFlow integration: batch logging support
MLFlow is the most popular model deployment framework at the moment. We'll like to support WhyLogs logging in MLFlow.

* WhyLogs as model artifact: when running an MLFlow experiment run, we'd like to output WhyLogs as a model artifact using `mlflow.log_artifact(..)` call. This will enable WhyLogs to appear under `artifacts/whylogs` path of the model artifact storage system. - See PR
~* WhyLogs metrics as an experiment metric: when user trigger whylogs in mlflow, the library will then call into `mlflow.log_metrics()` to spit out whylogs metrics if enabled. ~ See #70
~* There will be a "detailed" metric mode where metrics of all columns will be reported. This mode is disabled by default as it can increase the workload of the MLFlow system~ See #70
~*  Analyzing whylogs with mlflow runs~ See #71
~* Production monitoring for `python_function` model deployments~ See #69

andyndang:
Integration with MLFlow by patching the module:
```
import mlflow
import whylogs

whylogs.enable_mlflow()

import numpy as np
import pandas as pd
pdf = pd.DataFrame(
    data=[[1, 2, 3, 4, True, "x", bytes([1])]],
    columns=["b", "d", "a", "c", "e", "g", "f"],
    dtype=np.object,
)

active_run = mlflow.start_run()

# log a Pandas dataframe under default name
mlflow.whylogs.log_pandas(pdf)

# log a Pandas dataframe with custom name
mlflow.whylogs.log_pandas(pdf, "another dataset")

mlflow.end_run()
```
Fixes #67 partially


ibackus (approved):
Okay, this all looks good to me.  I'd like to just jot down a couple notes for future work, but I think we can merge this PR
## Future work
* Examples/integration for accessing the MLFlow artifacts
* automatic logging of dataframes/data within MLFlow runs?  i.e. no need to call `mlflow.whylogs.log_pandas(df)`
* basic tests

andyndang:
> * automatic logging of dataframes/data within MLFlow runs? i.e. no need to call mlflow.whylogs.log_pandas(df)

I'm not sure about this since if there are multiple PDFs created during the run, some of them temporary then we can't distinguish them. We can make it simpler to log a PDF with something like `datafram.whylog("custom-name")`, but I'd err from implicit logging in this case

ibackus:
Again, I think we should merge this PR, but there must be some way to do automatic logging of dataframes/data if I understand MLFlow correctly.

I may be mistaken, but I thought MLFlow can automatically log model parameters and things like that, and they also offer logging of simple data statistics like min/max, stddev, etc.  I'm not sure how automatic that is when you use MLFlow, but if model parameters and statistics can be logged, we should conceivably be able to log our statistics in the same manner.

andyndang:
>  there must be some way to do automatic logging of dataframes/data if I understand MLFlow correctly.

MLFlow doesn't track dataframes automatically. You can enable automatic training metrics logging for model training frameworks like H2O, but that's a different thing since each framework emits a predefined number of metrics with a predefined number of steps. MLFlow doesn't log any data metrics automatically otherwise FYI - you'll have to call `log_metrics` explicitly if you want to track min/max etc...

How would you reason about?
```
df1 = ...
# renaming a column. Should it be tracked?
df2 = df1.rename({"old_name", "new_name"})
df3 = d2.concat(df2)
```

That would be:
1) Expensive to track for each single "views" created. The only way to do it is to hijack all Pandas transformation
2) Noisy from customer's experience. Either you have a new dataset name for each DF, or you merge them into a single data.

ibackus:
Oh yeah, 100% makes sense.

I guess I had been thinking more about piggy-backing onto MLFlow to manage integration with existing ML model training frameworks.  Hence future work, we'll obviously need integration with those frameworks separately.

PR #75, opened 2020-10-20
Replace broken GitHub link

ishotjr:
One of the first things that people do when looking at the docs is likely clicking the GitHub link to get their hands on some code - but the current URL is https://github.com/whylabs/whylogs which is a 404 (at least for public users; I'm guessing a repo was renamed as new languages were added?). The examples repo links to both libraries so is probably a good substitute. :bowtie: 🔎 🔗

### DIFF
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -7,7 +7,7 @@ Overview
 Introduction
 ===================================
 
-WhyLogs (https://github.com/whylabs/whylogs) is an open source data quality \
+WhyLogs (https://github.com/whylabs/whylogs-examples) is an open source data quality \
 library that uses advanced data science statistics to log and monitor data \
 for your AI/ML application. WhyLogs is designed to scale with your MLOps \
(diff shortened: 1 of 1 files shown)


andyndang:
Thanks for the the change! It should link to the main repo though (https://github.com/whylabs/whylogs-python) as the example repo doesn't provide sufficient context compared to the main repo. 

We definitely should link to the example repo from the documentation page as well though! But that's probably a different link name

docs/overview.rst
```diff
@@ -7,7 +7,7 @@ Overview
 Introduction
 ===================================
 
-WhyLogs (https://github.com/whylabs/whylogs) is an open source data quality \
+WhyLogs (https://github.com/whylabs/whylogs-examples) is an open source data quality \
```
andyndang, line 10: Should link to https://github.com/whylabs/whylogs-python instead

andyndang:
This is fixed by: https://github.com/whylabs/whylogs-python/pull/85

Issue #31, opened 2020-09-15
logging


ibackus:
We need to add lots more logging, especially at the debug level, all over WhyLogs.

bernease:
The irony!

jamie256:
We added targeted logging for errors that customers hit, around configuration and serialization. Can always add more, but closing in favor of putting in better operational logging in next version.

PR #87, opened 2020-10-29
Use Python logger for logging

andyndang:
Partially for this issue #31

### DIFF
--- a/src/whylogs/__init__.py
+++ b/src/whylogs/__init__.py
@@ -5,6 +5,11 @@ from .app.session import get_or_create_session
 from .app.session import reset_default_session
 from .mlflow import enable_mlflow
 
+# noinspection PyProtectedMember
+from .util.logging_utils import _configure_loggers
+
+_configure_loggers(__name__)
+
 __all__ = [
     "ColumnProfile",
     "DatasetProfile",
--- a/src/whylogs/app/config.py
+++ b/src/whylogs/app/config.py
@@ -199,6 +199,9 @@ class SessionConfigSchema(Schema):
         return SessionConfig(**data)
 
 
+logger = getLogger(__name__)
+
+
 def load_config():
     """
     Load logging configuration, from disk and from the environment.
@@ -219,7 +222,6 @@ def load_config():
     """
     import os
 
-    logger = getLogger(__name__)
     cfg_candidates = [
         os.environ.get("WHYLOGS_CONFIG"),
         WHYLOGS_YML,
--- a/src/whylogs/app/logger.py
+++ b/src/whylogs/app/logger.py
@@ -2,6 +2,7 @@
 Class and functions for whylogs logging
 """
 import datetime
+import logging
 import typing
 from typing import List, Optional
 
@@ -10,6 +11,9 @@ from pandas._typing import FilePathOrBuffer
 
 from whylogs.app.writers import Writer
 from whylogs.core import DatasetProfile
+
+
+py_logger = logging.getLogger(__name__)
 
 
 class Logger:
@@ -73,7 +77,7 @@ class Logger:
         Synchronously perform all remaining write tasks
         """
         if not self._active:
-            print("WARNING: attempting to flush a closed logger")
+            py_logger.warning("WARNING: attempting to flush a closed logger")
             return
 
         for writer in self.writers:
@@ -86,7 +90,7 @@ class Logger:
         :return: the result dataset profile. None if the logger is closed
         """
         if not self._active:
-            print("WARNING: attempting to close a closed logger")
+            py_logger.warning("WARNING: attempting to close a closed logger")
             return None
 
         self.flush()
--- a/src/whylogs/app/session.py
+++ b/src/whylogs/app/session.py
@@ -5,6 +5,7 @@ import datetime
 from logging import getLogger as _getLogger
 from typing import Dict, List, Optional
 from uuid import uuid4
+import logging
 
 import pandas as pd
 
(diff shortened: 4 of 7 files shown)


src/whylogs/__init__.py
```diff
@@ -5,6 +5,11 @@
 from .app.session import reset_default_session
 from .mlflow import enable_mlflow
 
+# noinspection PyProtectedMember
```
ibackus, line 8: Since WhyLogs is a library (not just an application) we should not configure the loggers during import.  Logging config and control should be handled by any running application, and should be do-able within the standard python logging framework.

My preference is to provide a convenience utility to easily enable logging for interactive workflows, something like:

`whylogs.display_logging()`

We could allow users to configure this to get called in their environment at whylogs import time, if we really want.

src/whylogs/app/config.py
```diff
@@ -199,6 +199,9 @@ def make_session(self, data, **kwargs):
         return SessionConfig(**data)
 
 
+logger = getLogger(__name__)
```
ibackus, line 202: Since whylogs is a library and we don't have control over when modules get imported/used, all `getLogger()` calls should be within functions or classes, not at the module level.   This makes dynamic configuration possible.   Importantly, this allows users to configure whylogs regardless of import order. This is standard with python logging for library code.

For application code which we can control the workflow of, this is not necessary, but it's probably for the best to just stick to this paradigm in whylogs.

src/whylogs/app/logger.py
```diff
@@ -12,6 +13,9 @@
 from whylogs.core import DatasetProfile
 
 
+py_logger = logging.getLogger(__name__)
```
ibackus, line 16: Move from module to function level.

src/whylogs/app/logger.py
```diff
@@ -73,7 +77,7 @@ def flush(self):
         Synchronously perform all remaining write tasks
         """
         if not self._active:
-            print("WARNING: attempting to flush a closed logger")
+            py_logger.warning("WARNING: attempting to flush a closed logger")
```
ibackus, line 80: WARNING text redundant.

src/whylogs/app/logger.py
```diff
@@ -86,7 +90,7 @@ def close(self) -> Optional[DatasetProfile]:
         :return: the result dataset profile. None if the logger is closed
         """
         if not self._active:
-            print("WARNING: attempting to close a closed logger")
+            py_logger.warning("WARNING: attempting to close a closed logger")
```
ibackus, line 93: WARNING text redundant.

src/whylogs/app/session.py
```diff
@@ -13,6 +14,8 @@
 from whylogs.app.writers import Writer, writer_from_config
 from whylogs.core import DatasetProfile
 
+py_logger = logging.getLogger(__name__)
```
ibackus, line 17: Move from module to function level.

src/whylogs/app/session.py
```diff
@@ -236,7 +239,7 @@ def close(self):
         Deactivate this session and flush all associated loggers
         """
         if not self._active:
-            print("WARNING: attempting to close an inactive session")
+            py_logger.warning("WARNING: attempting to close an inactive session")
```
ibackus, line 242: WARNING text redundant.

src/whylogs/app/session.py
```diff
@@ -259,10 +262,8 @@ def remove_logger(self, dataset_name):
 
         """
         if self._loggers.get(dataset_name) is None:
-            print(
-                "WARNING: logger {} is not present in the current Session".format(
-                    dataset_name
-                )
+            py_logger.info(
```
ibackus, line 265: This may be relevant.  Copied from [stackoverflow](https://stackoverflow.com/questions/9595009/warnings-warn-vs-logging-warning):

> warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning
> 
> logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted

src/whylogs/mlflow/patcher.py
```diff
@@ -8,7 +8,7 @@
 
 from whylogs.app.logger import Logger
 
-logger = logging.getLogger("whylogs.mlflow")
+logger = logging.getLogger(__name__)
```
ibackus, line 11: Move from module to function level.

src/whylogs/util/logging_utils.py
```diff
@@ -0,0 +1,42 @@
...
+import logging
+import logging.config
+
+# Logging format example:
+# 2018/11/20 12:36:37 INFO whylogs.cli: Log new dataset
+LOGGING_LINE_FORMAT = "%(asctime)s %(levelname)s %(name)s: %(message)s"
+LOGGING_DATETIME_FORMAT = "%Y/%m/%d %H:%M:%S"
+
+
+def _configure_loggers(root_module_name):
```
ibackus, line 11: See comment above.  I think it's worth exposing this as a convenience function.

ibackus (changes requested):

lalmei:
@andyndang @ibackus is it stale or should I should I make the change in the review ?

ibackus:
@lalmei the specific changes here may be stale (not sure), but the general feedback should be applied for python logging, i.e:
* Use the logger
* don't configure the logger at import (we can use a convenience function if we want)
* initialize loggers within function calls, not at the module level
* Use warnings as follows:

> warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning
> 
> logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted

andyndang:
I'll cancel this one! @lalmei

PR #90, opened 2020-11-12
RAPIDS CuDF integration (sans tests)

bernease:
Adds optional import (try-block) of RAPIDS CuDF dataframe object (though not included in setup.py dependencies)

NOTE: There were a number of pending formatting changes from black unrelated to this change. I will add them if it blocks build.

### DIFF
--- a/src/whylogs/app/logger.py
+++ b/src/whylogs/app/logger.py
@@ -37,8 +37,7 @@ class Logger:
         writers=List[Writer],
         verbose: bool = False,
     ):
-        """
-        """
+        """"""
         if session_timestamp is None:
             session_timestamp = datetime.datetime.now(datetime.timezone.utc)
         self.dataset_name = dataset_name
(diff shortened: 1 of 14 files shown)


ibackus:
We should add a very simple test just to verify the CuDF works.

Longer term this brings out the issue that we'll need to start having tests which target optional dependencies.  This can be done pretty easily (though it's a bit hacky) by modifying `sys.modules` in a `pytest` fixture.

bernease:
I completely agree. I was also trying to figure out how to target optional
dependencies in the test in another change but hadn't gone the sys.modules
route (or find success).

andyndang:
> We should add a very simple test just to verify the CuDF works.

Surprisingly hard! We can do mock but actual CuDF requires enormous infra set up! 

There CI pipeline doesn't have GitHub action tests either (since the library requires CUDA, understandably): https://github.com/rapidsai/cudf/tree/branch-0.17/.github

Anyway this CR seems to require some rework as in it does not address the issue here: https://github.com/rapidsai/cudf/issues/6743

bernease:
Replicated in #91. Closing.

PR #91, opened 2020-11-12
WHY-1747 Support tracking a CuDF dataframe

andyndang:
This is caused by https://github.com/rapidsai/cudf/issues/6743

Manually tested since we don't have the infra to test with CuDF automatically

### DIFF
--- a/src/whylogs/core/datasetprofile.py
+++ b/src/whylogs/core/datasetprofile.py
@@ -27,6 +27,14 @@ from whylogs.util.data import getter, remap
 from whylogs.util.dsketch import FrequentNumbersSketch
 from whylogs.util.time import from_utc_ms, to_utc_ms
 
+# Optional import for cudf
+try:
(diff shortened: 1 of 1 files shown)


ibackus:
@andyndang do you just want us to review?  If so, mark us as reviewers

If you are the owner of this PR I think you should assign yourself.  I find that workflow is good on github

andyndang:
> @andyndang do you just want us to review? If so, mark us as reviewers

Yes 👍

ibackus (approved):
Seems reasonable, even if I can't test this.

bernease (approved):
Looks fine. Should this be paired with a change in dependencies on our end?

andyndang:
> Looks fine. Should this be paired with a change in dependencies on our end?

No. We can't have any dependency with CUDF since it doesn't work on normal environments. They are not installable via pypi so I won't put them in any extra dependency group either

PR #100, opened 2020-12-02
Add tox, setup.cfg and pyproject.toml to the package

andyndang:
Disable black code validation in pyproject.toml for now


setup.cfg
```diff
@@ -1,10 +1,146 @@
...
+install_requires =
+    click>=click==7.1.2
+    python-dateutil>=2.8.1
+    protobuf>=3.12.2
+    pyyaml>=5.3.1
+    pandas>1.0
+    marshmallow>=3.7.1
+    numpy>=1.18
+    whylabs-datasketches>=2.0.0b7
+    boto3==1.14.1
```
ibackus: I think these lines should be changed to:

```python
    boto3 >= 1.14.1
    botocore >= 1.17.44
    s3fs==0.4.2
```

As far as I know, we don't need any newer `s3fs` features from v5.x, and it's v 5.x which introduces the conflicting aiobotocore dependencies.  Additionally, this allows us to not peg the boto3/botocore dependencies which offers more flexibility as a library.  

We _really should not_ peg the boto3/botocore dependencies for a library since these are so widely used for AWS, whereas the s3fs library is less broadly used, and new features are probably less important for other libraries that a user may have installed.

ibackus (changes requested):

setup.cfg
andyndang: Can't :). See: https://github.com/whylabs/whylogs-python/issues/99, which is caused by: https://github.com/aio-libs/aiobotocore/issues/840

andyndang: Until we move off S3fs we have to peg them, or pip won't let me develop properly!!

scripts/profiler.py
```diff
@@ -161,28 +161,15 @@ def run(
...
-    reader = csv_reader(
-        input_path,
-        fmt,
-        parse_dates=parse_dates,
-        nrows=nrows,
-        sep=separator,
-        dropna=dropna,
-        infer_dtypes=infer_dtypes,
-    )
+    reader = csv_reader(input_path, fmt, parse_dates=parse_dates, nrows=nrows, sep=separator, dropna=dropna, infer_dtypes=infer_dtypes,)
```
lalmei, line 172: is this from `black` ?

andyndang: Yeah probably :D

setup.cfg
ibackus: @andyndang, pinning to s3fs==0.4.2 removes the aiobotocore dependency.

andyndang: Oh yes! Thanks. Fixed in the next rev

ibackus: Okay, looks good to me now!

ibackus (commented):
Just a question...I'm not 100% sure what changes this actually introduces to the build system.  It appears that using `pip install -e .[all]` works, which provides the installation in editable mode to my current environment that I require.  Is that correct @andyndang ?

andyndang:
> Just a question...I'm not 100% sure what changes this actually introduces to the build system. It appears that using `pip install -e .[all]` works, which provides the installation in editable mode to my current environment that I require. Is that correct?

I've updated so that you still do `pip install -e ".[dev]"` will keep working. However it's better to `pip install -r requirements-dev.txt` and then `make develop`

lalmei:
+1 for makefile

lalmei (approved):
LGTM

ibackus:
> > Just a question...I'm not 100% sure what changes this actually introduces to the build system. It appears that using `pip install -e .[all]` works, which provides the installation in editable mode to my current environment that I require. Is that correct?
> 
> I've updated so that you still do `pip install -e ".[dev]"` will keep working. However it's better to `pip install -r requirements-dev.txt` and then `make develop`

Just for posterity, pip installing the `requirements-dev.txt` is fine for a deterministic whylogs-only development environment.  It does not work for:

* Simultaneous development of another project which depends on whylogs as a library
* Exploring other versions of whylogs dependencies.  Whylogs is a library, not an application, we should work on it and expect it to work across a broad range of dependency versions

andyndang:
> Simultaneous development of another project which depends on whylogs as a library

In this case, you should use `pip install -e .` (minimal dependencies) and then `make develop`

ibackus:
Not necessarily if you are doing development on both whylogs and your own project.  (I do this all the time).  You may want the whylogs development dependencies for running tests locally etc.

andyndang:
> Not necessarily if you are doing development on both whylogs and your own project. (I do this all the time). You may want the whylogs development dependencies for running tests locally etc.

If you want to run tests on whylogs as you modify them you should just run them using `tox`, which will handle dependencies for you :). I don't see the case where you need to test both whylogs and your custom projects at the same time in the same environment using the same `python setup.py test` command?

ibackus:
> If you want to run tests on whylogs as you modify them you should just run them using `tox`, which will handle dependencies for you :). I don't see the case where you need to test both whylogs and your custom projects at the same time in the same environment using the same `python setup.py test` command?

Makes sense!

PR #101, opened 2020-12-03
added coverage computation

lalmei:
It is not added to ci yet.


ibackus (changes requested):
Looks mainly good, but the one issue is the unit tests need to be run separately from the component tests.  I believe you can update existing coverage if pytest gets run twice.

lalmei:
> Looks mainly good, but the one issue is the unit tests need to be run separately from the component tests. I believe you can update existing coverage if pytest gets run twice.

I will try to see if the coverage data aggregate correctly, if not I say we run together for now, the test are currently rather quick.

ibackus:
> I will try to see if the coverage data aggregate correctly, if not I say we run together for now, the test are currently rather quick.

Agreed.

ibackus (approved):

andyndang (approved):

PR #110, opened 2020-12-18
Why 1967 logger bug fix

### LINKED ISSUE #109: log_dataframe in session doest  ot output profile

lalmei:
Close #109

### DIFF
--- a/src/whylogs/app/session.py
+++ b/src/whylogs/app/session.py
@@ -119,6 +119,8 @@ class Session:
 
         return logger
 
+
+
     def log_dataframe(
         self,
         df: pd.DataFrame,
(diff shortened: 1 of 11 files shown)


src/whylogs/app/session.py
```diff
@@ -150,7 +152,7 @@ def log_dataframe(
             dataset_name, dataset_timestamp, session_timestamp, tags, metadata
         )
 
-        ylog.log_dataframe(df)
+        _ = ylog.log_dataframe(df)
```
andyndang: Python noob here. What's the purpose of the `_` here?

andyndang (approved):

src/whylogs/app/session.py
ibackus: @andyndang Just a dummy variable.  A convention to show that there is output but it's not used for anything.

lalmei: throw away variable, I removed it. there was confusion with this api call and previous versions  I found in the notebook that expected a profile from `logger.log_dataframe` .

PR #114, opened 2020-12-23
log rotation with time

lalmei:


ibackus (approved):
This seems reasonable to me at the moment.

ibackus (approved):

PR #116, opened 2021-01-02
segment support

lalmei:
you can check `scrips/segements_logging.py for the use cases


scripts/segments_logging.py
```diff
@@ -0,0 +1,27 @@
...
+        print(session.get_config())
+        profile = logger.log_dataframe(df)
+        time.sleep(2)
+        profile = logger.log_dataframe(df)
+
+    with session.logger("my_rotated_seg_two_keys", segments=["home_ownership", "sub_grade"], with_rotation_time="s", cache=1) as logger:
+        print(session.get_config())
+        profile = logger.log_csv("data/lending-club-accepted-10.csv")
+        time.sleep(2)
+        profile = logger.log_dataframe(df)
```
andyndang: just from the API perspective, what does `logger.log...` returns now that we pass in a list of segment keys?

lalmei: the "log" api is bit all over the place, in session "log_*" returns a profile, while in logger, it simply does that, log the data into the current active datasetprofile. These lines are an artifact, since it doest return anything those are None profiles. It came from a previous version, and i will remove it as to not create a confusion

andyndang: FYI, there's a use case of accessing the underlying dataset profiles when running whylogs without relying on the `writer` to write out the data. That's why the API returns them in the session (just makes it a little bit easier).

There might be an argument to have a `DatasetRun` or `DatasetProfilesBatch` object that contains all the segment data here, though the same is still a bit hazy to me

src/whylogs/core/datasetprofile.py
```diff
@@ -131,6 +131,7 @@ def __init__(
         self._metadata = metadata.copy()
         self.columns = columns
         
+        self._active=True
```
andyndang: Redundant? I don't see the usage anywhere.

Also, a profile shouldn't have active or not - active is tied to logger/session because they are coupled with a writer and log rotation -  but a profile is just a holder for data

src/whylogs/app/logger.py
```diff
@@ -34,15 +37,20 @@ def __init__(self,
...
         session_timestamp: Optional[datetime.datetime] = None,
-        tags: typing.Dict[str, str] = None,
-        metadata: typing.Dict[str, str] = None,
+        tags: Dict[str, str] = {},
+        metadata: Dict[str, str] = None,
         writers = List[Writer],
         verbose: bool = False, 
         with_rotation_time: Optional[str] = None, 
         cache: int =1,
+        segments: Optional[Union[List[List[Dict]], List[str]]] = None,
```
andyndang: this signature is very nested - I'd rather keep type hierarchy lighweight. For the `key, value` case, can we just stick to defining an object instead?

src/whylogs/app/logger.py
```diff
@@ -75,45 +80,82 @@ def __enter__(self):
...
+            self.segments=None
+            self.segment_type=None
+
+    def intialize_profiles(self, 
+        dataset_timestamp: Optional[datetime.datetime] = datetime.datetime.now(datetime.timezone.utc)) -> None:
+        
 
+        full_profile=None
+        if self.full_profile_check():
+            full_profile = DatasetProfile(
```
andyndang, line 139: we call it `overall` - let's keep the terminology consistent

src/whylogs/app/logger.py
```diff
@@ -75,45 +80,82 @@ def __enter__(self):
...
+                                    self.dataset_name,
+                                    dataset_timestamp=dataset_timestamp,
+                                    session_timestamp=self.session_timestamp,
+                                    tags=self.tags,
+                                    metadata=self.metadata,
+                                    session_id=self.session_id
+                                )
+
+        self._profiles.append({"full_profile": full_profile, "segmented_profiles" : {}})
```
andyndang: any reason we're not defining an explicit type but rather using a `dict` to hold data?

src/whylogs/app/logger.py
```diff
@@ -168,15 +212,31 @@ def flush(self,rotation_suffix:str =None):
...
+                else:
+                    writer.write(self._profiles[-1]["full_profile"],rotation_suffix)
+
+            if self.segments is not None:
+
+                for hashseg, each_seg_prof in self._profiles[-1]["segmented_profiles"].items():
+                    seg_suffix = hashseg
+                    full_suffix= "_"+seg_suffix
+                    if rotation_suffix is None:
+                        writer.write(each_seg_prof,full_suffix)
```
andyndang: spacing? our linter doesn't check for this?

src/whylogs/app/logger.py
```diff
@@ -168,15 +212,31 @@ def flush(self,rotation_suffix:str =None):
...
+                    else:
+                        full_suffix+=rotation_suffix
+                        writer.write(each_seg_prof,full_suffix)
+
+    
+
+    def full_profile_check(self,):
+        return ((self.segments is None ) or ((self.segments is not None) and self.profile_full_dataset))
```
andyndang: what does this method do? can you add a bit of explanation?

src/whylogs/app/logger.py
```diff
@@ -189,19 +249,21 @@ def close(self) -> Optional[DatasetProfile]:
...
         self._active = False
-        profile = self._profiles[-1]
+        profile = self._profiles[-1]["full_profile"]
         self._profiles = None
         return profile
 
     def log(
         self,
-        features: typing.Dict[str, any] = None,
+        features: Dict[str, any] = None,
```
andyndang: shouldn't this be `Optional[...]` (not related to the current PR but still wondering)

src/whylogs/app/logger.py
```diff
@@ -216,20 +278,47 @@ def log(
...
-            self._profiles[-1].track(features)
+            # full profile
+            if self.full_profile_check():
+                self._profiles[-1]["full_profile"].track(features)
+            
+            if self.segments:
+                self.log_segments(pd.DataFrame(features))
         else:
-            self._profiles[-1].track_datum(feature_name, value)
+            if (self.segments is None ) or ((self.segments is not None) and self.profile_full_dataset):
```
andyndang: extract this to a boolean with a meaningful name. I think I understand what you're trying to do but long conditions are just hard to humans to grok

src/whylogs/app/logger.py
```diff
@@ -216,20 +278,47 @@ def log(
...
+                self._profiles[-1]["full_profile"].track_datum(feature_name, value)
+            
+            if self.segments:
```
andyndang, line 326: I'm not sure I understand the following code block in terms of how the data is segmented? Given it's a piece of core logic, it's worth extracting to a method with documentation

src/whylogs/app/logger.py
```diff
@@ -75,45 +80,82 @@ def __enter__(self):
...
         self.close()
 
+
+    @property
+    def profile(self,):
+        """
+        :return: the last backing dataset profile
+        :rtype: DatasetProfile
+        """
+        return self._profiles[-1]["full_profile"]
```
andyndang, line 102: what does the last backing DS represent here? A segment or  the full dataset? Do we keep all of the profiles in memory?

src/whylogs/app/logger.py
```diff
@@ -75,45 +80,82 @@ def __enter__(self):
...
     @property
-    def profile(self):
+    def segmented_profiles(self,)->DatasetProfile:
         """
         :return: the last backing dataset profile
         :rtype: DatasetProfile
         """
-        return self._profiles[-1]
+        
+        return self._profiles[-1]["segmented_profiles"]
```
andyndang, line 110: the name reads `segmented_profiles ` but we're getting a single profile back? The logic doesn't make sense to me here

src/whylogs/core/datasetprofile.py
lalmei: we can remove it for now, we do cache the profiles, but the idea of the active profile will be to keep the profile metadata alive when uncache the profile, so that the profile could be reloaded.

src/whylogs/app/logger.py
lalmei: Agree. it is essentially `SegmentTag`.

src/whylogs/app/logger.py
lalmei: yes

src/whylogs/app/logger.py
lalmei: +1

src/whylogs/app/logger.py
lalmei: no, more of an implement first, refactor later. Once we settle on the API we can create the data structures.

src/whylogs/app/logger.py
lalmei: I wil check, I have been finding these everywhere too.

src/whylogs/app/logger.py
lalmei:
```suggestion
        features: Optional[Dict[str, any]] = None,
```

scripts/segments_logging.py
lalmei: You can acess all the current active profiles with `logger.profile` -> `DatasetProfile` or `logger.segmented_profiles`-> `Dict[str,DatasetProfile]`.

src/whylogs/app/logger.py
lalmei: actually I find `overall` is a bit confusing versus full_profile and segmented_profile.

src/whylogs/app/logger.py
lalmei: It is the current active profile that is updated by `logger.log_*`.

src/whylogs/app/logger.py
lalmei: you get a `Dict[str,DatasetProfile]`, so you have acess to each segment. i will update the documentation.

lalmei: Done

src/whylogs/app/logger.py
lalmei: had previously defined, missed one

src/whylogs/app/logger.py
lalmei: Yeah this is a bit confusing, since it is a tracking a segmented datum. I think they way i did is inconsistent with the groupby method, since that method keeps the segmented feature in the data frame. I will fix this.

src/whylogs/app/logger.py
lalmei: done

src/whylogs/core/datasetprofile.py
lalmei:
```suggestion

```

.github/workflows/continuous-integration.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       max-parallel: 6
       matrix:
-        python-version: [ 3.6,3.7,3.8]
+        python-version: [3.6,3.7, 3.8]
```
andyndang, line 17: nit: consistent spacing?

src/whylogs/app/config.py
```diff
@@ -124,15 +136,17 @@ def __init__(
         pipeline: str,
         writers: List[WriterConfig],
         verbose: bool = False,
-        with_rotation_time: str =None,
-        cache :int = None,
+        with_rotation_time: str = None,
```
andyndang, line 139: nit: shouldn't this be `Optional[str]`?

andyndang (approved):
Looking good. Minor comments

Issue #107, opened 2020-12-18
Do not create local path when writer is `s3`


andyndang:
When setting the writer to `s3`, whylog still creates local path and thus you'll get `s3://bucket/etc..` locally.

This can be reproduced easily locally.

lalmei:
Merged in #118

PR #118, opened 2021-01-07
fix s3 writer creating local folders and added s3 mock tests

lalmei:


lalmei:
Closes #107

ibackus (approved):
Looks pretty good.  I see you are using moto.  Just a head's up if you find yourself needing to run external moto servers (which is required for libraries other than boto3 that access AWS) I do have some code for doing that.

It's significantly more overhead, but more flexible, and should be pretty much copy-pastable if you want.

I don't think it will be required right away, but may be an issue if testing anything which uses s3fs

andyndang (approved):
Looking good!

PR #124, opened 2021-01-18
[WIP] Update various configs

andyndang:
* Move coverage configs to setup.cfg
* Update default region for moto's s3
* Add missing submodules to API doc


docs/conf.py
```diff
@@ -165,14 +166,9 @@ def setup(app):
...
 # a list of builtin themes.
-html_theme = "sphinx_rtd_theme"
-
-# Theme options are theme-specific and customize the look and feel of a theme
-# further.  For a list of options available for each theme, see the
-# documentation.
-html_theme_options = {"sidebar_width": "300px", "page_width": "1200px"}
+html_theme = "whylogs"
 
-html_theme_path = ["_themes", ]
```
lalmei: can we use the  catalyst theme ? https://sphinx-themes.org/#theme-catalyst-sphinx-theme rather than custom one. It will be more difficult to maintain I believe.

PR #132, opened 2021-01-22
Enable flake8. Fixed existing errors

andyndang:
Note that our tox configuration at the moment will run `tests` before `linting`. Pretty annoying since it's super slow...

We should add a note about using Flake8 with pycharm somewhere: https://plugins.jetbrains.com/plugin/11563-flake8-support


tox.ini
```diff
@@ -16,12 +16,14 @@ passenv = GITHUB_*
 extras =
     all
     testing
-commands = python setup.py test
+commands = coverage run -m pytest
```
lalmei:
```suggestion
commands = pytest --cov='src/.' --cov-report term-missing --cov-report xml --verbose
```

lalmei (approved):
LGTM

PR #133, opened 2021-01-22
[bugfix] Fix mlfow tracking with predict API

andyndang:
The current code throws exception when the output is not a Pandas dataframe

* Tested locally with a sklearn model (np.array output):
```
mlflow models predict -m data/0/c67968018081464eabf1bece1dde68f3/artifacts/model -i input.json
```


lalmei (approved):
LGTM

PR #134, opened 2021-01-25
✨ bounding boxes and full local dataset

lalmei:


requirements.txt
```diff
@@ -16,3 +16,6 @@ s3transfer==0.3.3
 six==1.15.0
 urllib3==1.26.2
 whylabs-datasketches==2.0.0b7
+xlrd==2.0.1
+openpyxl==3.0.6
+puremagic == 1.10
```
andyndang: I don't see `puremagic` in `setup.cfg`

andyndang (changes requested):
Need to add dep to setup.cfg (I can't seem to be able to make suggestions for unchanged sections of the code)

requirements.txt
andyndang: Same with tqdm

src/whylogs/app/logger.py
```diff
@@ -355,6 +360,30 @@ def log_image(self,
 
         track_image(self._profiles[-1]["full_profile"])
 
+    def log_local_dataset(self, root_dir, folder_feature_name="folder_feature"):
+        from PIL.Image import Image as ImageType
+        dst = LocalDataset(root_dir)
+        for idx in tqdm(range(len(dst))):
```
andyndang: I don't think we should use progress bar here. It will interfere with users logging set up. We should rely on standard loggers only if we need to log out debug information

andyndang (changes requested):

src/whylogs/app/logger.py
lalmei: I dont think it interferences with the logging, it goes to stdout only.

requirements.txt
lalmei: thanks missed those

src/whylogs/app/logger.py
lalmei: Also this is only used when you log a dataset, which can take a while depending on the size.

andyndang: If you are in Docker, stdout is your logging FYI :)

Maybe put this behind a parameter such as `show_progress_bar`? And it should be an optional dependency imported locally here

lalmei: i will wrap it but I dont like to hide thin requirements. Otherwise it catches folks by surprise in the middle of run time

lalmei: @andyndang done

src/whylogs/app/logger.py
```diff
@@ -355,6 +364,82 @@ def log_image(self,
...
+            try:
+                from tqdm import tqdm
+            except ImportError as e:
+                show_progress = False
+                logger.debug(str(e))
+                logger.debug(
+                    "Unable to load tqdm; install tqdm for progress support")
+
+        dst = LocalDataset(root_dir)
+        for idx in tqdm(range(len(dst)), disable=(not show_progress)):
```
andyndang, line 391: This will fail if `tqdm` is not installed FYI

src/whylogs/app/logger.py
```diff
@@ -355,6 +364,82 @@ def log_image(self,
...
+            logger.debug(
+                "Unable to load PIL; install Pillow for image support")
+
+        if show_progress:
+            try:
+                from tqdm import tqdm
+            except ImportError as e:
+                show_progress = False
+                logger.debug(str(e))
+                logger.debug(
```
andyndang:
```suggestion
                logger.warning(
```
Warning is more appropriate here since the flag is explicitly set

andyndang (changes requested):

src/whylogs/app/logger.py
lalmei: yes also if you try to load images it will fail

requirements-test.txt
```diff
@@ -7,7 +7,7 @@ jmespath==0.10.0
 marshmallow==3.9.1
 matplotlib==3.3.3
 numpy==1.19.4
-pandas==1.1.4
+pandas<=1.2.1
```
lalmei:
```suggestion
pandas==1.1.4
```

src/whylogs/app/logger.py
lalmei: put it as a requirement

andyndang (approved):
LGTM

lalmei:
I will merge this after the drafter to test it out

Issue #99, opened 2020-11-30
Stop using s3fs 


andyndang:
S3fs consumes aiobotocore, which pins botocore version: https://github.com/aio-libs/aiobotocore/issues/840

With new Pip resolution logic, we are seeing failures when consuming whylogs in an environment that depends on a different version of boto3.

This means that users cannot install whylogs AND use different version of boto3. This is a major blocker for using the package for most people. 

Solution: need to stop using s3fs

lalmei:
Closed in #140

PR #140, opened 2021-01-31
Remove s3fs refs and use smart open instead

lalmei:
Smart open has a thin wrapper around boto3 only and also supports other cloud storages like azure, gcp etc.


lalmei:
Closes #99

andyndang (approved):
LGTM

PR #159, opened 2021-02-22
[Bugfix] Fix merge logic for various corner cases

andyndang:
* Default `ModelProfile` to `None` (it's a special case and shouldn't appear everywhere)
* Fixed merging two `ModelProfile` after deserializing from protobuf bytes
* Various cosmetics cleanups
* Bumped version


lalmei (approved):
Lgtm

PR #164, opened 2021-02-24
Fall back to number unique summary or string unique summary 

andyndang:
This is to enable the Python side to be compatible with the java output while this issue is being addressed:
https://github.com/whylabs/whylogs-java/issues/17


lalmei (approved):
LGTM

PR #168, opened 2021-03-11
Update README.md

andyndang:
* Update top k - actually 128 rather than 30
* Fixed broken notebook links
* Add binder link


lalmei (approved):
LGTM

PR #171, opened 2021-03-15
regression metrics

lalmei:


src/whylogs/core/metrics/model_metrics.py
```diff
@@ -10,22 +11,35 @@ class ModelMetrics:
 
     Attributes:
         confusion_matrix (ConfusionMatrix): ConfusionMatrix which keeps it track of counts with NumberTracker
+        regression_metrics (RegressionMetrics): Regression Metrics keeps track of a common regression metrics in case the targets are continous.
     """
 
-    def __init__(self, confusion_matrix: ConfusionMatrix = None):
-        if confusion_matrix is None:
-            confusion_matrix = ConfusionMatrix()
+    def __init__(self, confusion_matrix: ConfusionMatrix = ConfusionMatrix(),
```
andyndang: Don't initialize object here as part of function parameters

src/whylogs/core/metrics/regression_metrics.py
```diff
@@ -0,0 +1,118 @@
...
+        Returns:
+              RegressionMetrics: merged regression metrics
+        """
+
+        if self.count == 0:
+            return other_reg_met
+        if other_reg_met.count == 0:
+            return self
+
+        new_reg = RegressionMetrics()
```
andyndang: What about `prediction_field` etc? You also need to validate these fields match between two objects

tests/unit/core/metrics/test_regression_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+    regmet_sum.add(df["predictions"].to_list(), df["targets"].to_list())
+
+    regmet_2 = RegressionMetrics()
+    df_2 = pd.read_parquet(os.path.join(os.path.join(TEST_DATA_PATH, "metrics", "2021-02-13.parquet")))
+    regmet_2.add(df_2["predictions"].to_list(), df_2["targets"].to_list())
+    regmet_sum.add(df_2["predictions"].to_list(), df_2["targets"].to_list())
+
+    merged_reg_metr = regmet.merge(regmet_2)
+
+    assert merged_reg_metr.count == regmet_sum.count
```
andyndang, line 70: need to assert on field names

andyndang (changes requested):

src/whylogs/core/metrics/regression_metrics.py
lalmei: fixed in 8dc266ec0

tests/unit/core/metrics/test_regression_metrics.py
lalmei: fixed in 8dc266e

src/whylogs/core/metrics/model_metrics.py
lalmei: fixed in 8dc266e

src/whylogs/core/metrics/model_metrics.py
```diff
@@ -54,6 +68,14 @@ def compute_confusion_matrix(self, predictions: List[Union[str, int, bool]],
             self.confusion_matrix = self.confusion_matrix.merge(
                 confusion_matrix)
 
+    def compute_regression_metrics(self, predictions: List[float],
```
andyndang: `predictions` can be `List[int, float]` though, right?

src/whylogs/core/metrics/model_metrics.py
```diff
@@ -10,22 +11,35 @@ class ModelMetrics:
...
         confusion_matrix (ConfusionMatrix): ConfusionMatrix which keeps it track of counts with NumberTracker
+        regression_metrics (RegressionMetrics): Regression Metrics keeps track of a common regression metrics in case the targets are continous.
     """
 
-    def __init__(self, confusion_matrix: ConfusionMatrix = None):
+    def __init__(self, confusion_matrix: ConfusionMatrix = None,
+                 regression_metrics: RegressionMetrics = None,
+                 model_type: ModelType = ModelType.UNKNOWN):
         if confusion_matrix is None:
             confusion_matrix = ConfusionMatrix()
```
andyndang: can we have a case when both these objects are non-null and have values? I can't think of a case when both of these will live together.

I'd rather keep them separately and enforce that only one of them can be non-None at a time (and that will determine `model_type` if necessary)

andyndang (changes requested):

src/whylogs/core/metrics/model_metrics.py
lalmei: very true

src/whylogs/core/metrics/model_metrics.py
lalmei: there will be cases when both will live together, but we can keep one or the other for now.

lalmei: for now they are in two separate branches

src/whylogs/core/model_profile.py
```diff
@@ -64,41 +67,50 @@ def compute_metrics(self, targets,
...
+                                                    target_field=target_field,
+                                                    prediction_field=prediction_field)
+            self.metrics.model_type = ModelType.REGRESSION
+
+        elif tgt_type in ("binary", "multiclass") or model_type == ModelType.CLASSIFICATION:
+            self.metrics.model_type = ModelType.CLASSIFICATION
+
+            # if score are not present set them to 1.
+            if scores is None:
+                scores = np.ones(len(targets))
```
cswarth, line 83: np.ones() returns an array, so does it work to wrap that in np.array() a couple lines later?

src/whylogs/core/model_profile.py
```diff
@@ -48,14 +51,14 @@ def compute_metrics(self, targets,
         Parameters
```
cswarth, line 51: pydoc-markdown package we use for extracting doc comments is expecting Sphinx style documentation, but this method is using Numpy style.
I don't care which style we use but we ought to standardize on one, and the documentation for this method is not recognized and gets formatted badly.

src/whylogs/app/logger.py
```diff
@@ -159,45 +158,44 @@ def _intialize_profiles(self,
...
-            elif m.group(2) == 'd':
-                interval *= 60 * 60 * 24  # one day
-                self.suffix = "%Y-%m-%d"
-            else:
-                raise TypeError("Invalid rotation interval, expected integer followed by one of 's', 'm', 'h', or 'd'")
-            # time in seconds
-            current_time = int(datetime.datetime.utcnow().timestamp())
-            self.interval = interval * self.interval_multiplier
-            self.rotate_at = self.rotate_when(current_time)
+        if with_rotation_time is None:
```
cswarth, line 161: We're still going to need to merge my PR to update the documentation and get the unit test for the change to with_rotation_time

cswarth (approved):
couple of nits but lgtm
Would like to see commits squashed

src/whylogs/core/model_profile.py
cswarth: Don't worry about the doc change, I'll go through and standardize our comment style in one PR

andyndang (approved):
LGTM

src/whylogs/core/model_profile.py
lalmei: scores may be just a list from the user side.

src/whylogs/app/logger.py
lalmei: we merged that one already

PR #184, opened 2021-03-26
fix req and dev readme with protobuf step

lalmei:
Still need your input on where to modify and change.


requirements-dev.txt
```diff
@@ -1,120 +1,51 @@
...
 setuptools-black==0.1.5
 six==1.15.0
-smmap==3.0.4
-SQLAlchemy==1.3.20
-sqlparse==0.4.1
-tabulate==0.8.7
-toml==0.10.2
-tornado==6.1
+smart-open==4.1.2
+sphinx-autoapi>=1.4.0
```
andyndang, line 40: we shouldn't have sphinx devs in our dev dependencies. This is for documentation dependencies and they are separate

cswarth (approved):
LGTM

requirements-dev.txt
lalmei: i disagree, my understanding is that this was included as a full dependency for a developer working on this project. which what the team has been asking for. otherwise lets just remove this file.

lalmei: but agree it shouldn't be part of the main requirements

PR #186, opened 2021-03-31
Enable data to be logged to WhyLabs anonymous sessions

naddeoa:
This change adds a feature to whylogs that allows it to log data to
WhyLabs anonymous sessions. Anonymous sessions allow people to jump into
whylogs with a dataset, log it to dataset profiles, and send it to
WhyLabs for visualization and alerts without requiring a login.

This change updates whylogs to have a `Session` concept that is
essentially a wrapper around a bunch of loggers. Sessions manage
loggers and logger life cycles, as well as call WhyLabs to create and
sessions and upload the dataset profiles that loggers create.

This also adds an example for anonymous sessions with dates

This adds a snapshot of a dataset from kaggle
(https://www.kaggle.com/yugagrawal95/sample-media-spends-data) and an
example that shows how properly log that data to whylogs for an
anonymous session on WhyLabs. This dataset actually has a day date
column so its much clearer how the logger is supposed to work in regards
to dates.


src/whylogs/app/session.py
```diff
@@ -66,96 +120,50 @@ def __exit__(self, tpe, value, traceback):
...
-        session_timestamp: Optional[datetime.datetime] = None,
-        tags: Dict[str, str] = None,
-        metadata: Dict[str, str] = None,
-        segments: Optional[Union[List[Dict], List[str]]] = None,
-        profile_full_dataset: bool = False,
-        with_rotation_time: str = None,
-        cache_size: int = 1,
-        constraints: DatasetConstraints = None,
-    ) -> Logger:
+    def logger(self, args: LoggerArguments = defaultLoggerArgs) -> Logger:
```
lalmei, line 129:
```suggestion
    def logger(self, args: LoggerArguments = defaultLoggerArgs, datetime_feature_name: str = None) -> Logger:
```

lalmei:
@naddeoa @phantomgoose  should we close this PR ?

PR #196, opened 2021-04-09
Add whylabs-client to dependencies to allow creation of anonymous sessions

phantomgoose:


setup.cfg
```diff
@@ -39,6 +39,7 @@ install_requires =
     pandas>1.0
     marshmallow>=3.7.1
     numpy>=1.18
+    whylabs-client==0.1
```
andyndang: nit: use `>=` here. If you are worry you can use `whylabs-client >=0.1, <= 0.2` so that if we make breaking changes to the API it's reflected in the client.

andyndang (approved):

setup.cfg
```diff
@@ -39,6 +39,7 @@ install_requires =
     pandas>1.0
     marshmallow>=3.7.1
     numpy>=1.18
+    whylabs-client >=0.1, <= 0.2
```
andyndang, line 42: I think the spaces here are causing flake8 to fail FYI

andyndang (approved):

setup.cfg
phantomgoose: idk i reran the jobs and everything passed. Looked like a random test failure 🤷

PR #203, opened 2021-04-20
Configuration and docs updates

naddeoa:
This further strips configuration from the setup.cfg file working towards deleting it. The DEVELOPMENT.md file has been rewritten to be a bit more to-the-point and minimalist.

The patch version is also bumped in preparation for testing the poetry publishing logic and targets for updating the version number are now in the Makefile, which now has colored output.


DEVELOPMENT.md
```diff
@@ -43,43 +23,70 @@ Poetry manages virtualenvs as well. Typically, on a project that uses virtualenv
 poetry shell
 ```
 
-You shouldn't have to actually do that unless you want to run some of the things that are pip installed directly. One use case would be manually running `pytest` on a single test file rather than running them all with `make test`.
-
-## Managing Dependencies
+You shouldn't have to actually do that unless you want to run some of the executables things that are installed from pypi directly. One use case would be manually running `pytest` on a single test file rather than running them all with `make test`.
```
lalmei, line 26:
```suggestion
You shouldn't have to actually do that unless you want to run some of the executables that are installed from pypi directly. One use case would be manually running `pytest` on a single test file rather than running them all with `make test`.
```

DEVELOPMENT.md
```diff
@@ -43,43 +23,70 @@ Poetry manages virtualenvs as well. Typically, on a project that uses virtualenv
...
 
-Poetry manages all of the dependencies. If you need to add something then you should run the following.
+## Manging Dependencies
+This is done through poetry. There needs be a good reason to add dependencies that should be included in the commit message.
```
lalmei, line 29:
```suggestion
This is done through poetry. There needs be a good reason(s) to add dependencies, and they should be included in the commit message.
```

lalmei (approved):
LGTM, minor suggestions

.bumpversion.cfg
```diff
@@ -1,6 +1,5 @@
 [bumpversion]
-current_version = 0.4.5-dev1
-commit = True
+current_version = 0.4.6-dev0
```
andyndang: Hey any reason we're bumping the minor version? We haven't released `0.4.5`.

You can bump the `dev-1` by doing `bump2version build` rather `patch`

andyndang (dismissed):
LGTM except the bumping of the `patch` version

lalmei (approved):
LGTM

PR #204, opened 2021-04-21
Move profile IO into background threads

andyndang:
Publish whylogs using background threads

This allows us to avoid blocking the main Python process when writing whylogs to disk.

 This is very important for latency sensitive applications such as
    real time model serving

 In addition, this change makes the Writer objects track these threads and wait for them
    when the session ends


lalmei (approved):
LGTM

PR #205, opened 2021-04-21
Add a guest session notebook and progress indicators

naddeoa:
This adds a notebook for the guest session feature as well as tqdm to the `Session` class to report progress when closing sessions.


pyproject.toml
```diff
@@ -60,6 +60,7 @@ autoflake = "^1.4"
 flake8 = "^3.9.1"
 aiohttp = "^3.7.4"
 aiohttp_cors = "^0.7.0"
+jupyter = "^1.0.0"
```
andyndang: No we shouldn't depend on Jupyter for dev. We already depend on `ipykernel`. You only need 1 jupyter for the whole machine and not for individual packages

src/whylogs/app/session.py
```diff
@@ -8,6 +8,7 @@
 from uuid import uuid4
 
 import pandas as pd
+from tqdm.auto import tqdm
```
andyndang, line 11: we shouldn't depend on `tqdm` as a hard requirement. It's a cosmetic thing, and in this case, Jupyter notebook so it can be an extra dependency with option import (`try except...` pattern for import)

I know we introduced tqdm in `logger.py` but this is not a good pattern and we should stop doing it

src/whylogs/app/session.py
```diff
@@ -327,10 +328,12 @@ def close(self):
 
         self._active = False
         loggers = list(self._loggers.items())
-        for name, logger in loggers:
-            if logger.is_active():
-                logger.close()
-            self.remove_logger(name)
+        with tqdm(loggers) as t:
```
andyndang: This will polutes user's stdout!! Imagine running a web application with expected logging format in stdout, only to be polluted by tqdm in non-interactive mode!

A better solution is to detect if you are in a Jupyter notebook and attempt to import tqdm at that point, and if not available, should just not fail the program but could just print out some warning about no progress bar due to `tqdm` not available.

See: https://stackoverflow.com/questions/15411967/how-can-i-check-if-code-is-executed-in-the-ipython-notebook

andyndang (changes requested):
Added various comment about tqdm + Jupyter logic

src/whylogs/app/session.py
naddeoa: You caught the review while I was still working on it. The current version has tqdm disabled unless a flag is passed to the session to enable it. We can add the conditional tqdm import as well if you want to min/max memory but the library boasts a very slim runtime.

src/whylogs/app/session.py
naddeoa: Talked offline, we will depend on it but it will default to off.

pyproject.toml
naddeoa: This is removed. I had to add a dev dependency on the widgets though or we wouldn't be able to use tqdm's html progress bar and it can't pull that package from the system environment it seems.

andyndang (approved):

PR #206, opened 2021-04-22
Fix bug that omits generated protobuf sources from builds

naddeoa:
For some reason, poetry wasn't including the generated protobuf sources
in the wheel or the source tarballs. I couldn't find anything that was
excluding them so I ended up explicitly including them in the
pyproject.toml and that fixes the issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,27 +33,27 @@ help:
 
 bump-patch: ## Bump the patch version (_._.X) everywhere it appears in the project
 	@$(call i, Bumping the patch number)
-	bumpversion patch --allow-dirty
+	poetry run bumpversion patch --allow-dirty
 
 bump-minor: ## Bump the minor version (_.X._) everywhere it appears in the project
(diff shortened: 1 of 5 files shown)


andyndang (approved):
LGTM

Makefile
```diff
@@ -33,27 +33,27 @@ help:
 
 bump-patch: ## Bump the patch version (_._.X) everywhere it appears in the project
 	@$(call i, Bumping the patch number)
-	bumpversion patch --allow-dirty
+	poetry run bumpversion patch --allow-dirty
```
andyndang, line 36: https://github.com/peritus/bumpversion is not being maintained. You want `bump2version`

lalmei:
btw notice that the coverage is no longer being computed .

lalmei:
it is not a blocker but we should fix this soon.

andyndang (approved):

PR #227, opened 2021-05-07
Static HTML profile page

valer-whylabs:
Create a static HTML page where users can load their generated profile (JSON) file.
The page should give them an insight, how the observatory profile page works - with limited functionalities.


viewer/index.html
```diff
@@ -0,0 +1,224 @@
...
+
+  <body>
+    <nav class="navbar navbar-inverse fixed-top">
+      <div class="container-fluid">
+        <div class="navbar-header">
+          <h5>Profile viewer for whylogs</h5>
+          <ul class="navbar-nav d-inline-block">
+            <li class="nav-item">
+              <p>
+                Maintianed by
```
phantomgoose: typo: Maintained :)

viewer/index.html
```diff
@@ -0,0 +1,224 @@
...
+                <a class="navbar-link" href="https://try.whylabsapp.com/" target="_blank" rel="noreferrer noopener"
```
phantomgoose: we may want to add a `?utm_source` here so we know who's coming into the sandbox from this page

viewer/index.html
```diff
@@ -0,0 +1,224 @@
...
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+
+    <!-- Bootstrap core JavaScript
+    ================================================== -->
+    <!-- Placed at the end of the document so the pages load faster -->
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
```
phantomgoose: do we need jquery for this?

viewer/index.html
mrmcduff: what would the utm_source be? Should we make a special tag for that?

viewer/index.html
mrmcduff: It's used in the JS -- sort of our replacement interactivity (since there is no react)

viewer/index.html
```diff
@@ -0,0 +1,224 @@
...
+    <script src="https://d3js.org/d3.v6.min.js"></script>
+    <script
```
mrmcduff: @valer-whylabs  -- Just realized; we can't call out to external libraries like this, so we should download the minified versions of them from
https://jquery.com/download/
https://d3js.org/
and
https://getbootstrap.com/docs/4.3/getting-started/download/

Yet another reason why it's good that we kept the dependencies to a minimum

viewer/index.html
mrmcduff:
```suggestion
                Maintained by
```

viewer/whylogs-script.js
```diff
@@ -0,0 +1,547 @@
...
+
+(function () {
+  var MESSAGES = {
+    error: {
+      noInputElementFound: "It seems we could not find the file input element.",
+      noBrowserSupport: "This browser does not seem to support the `files` property of file inputs.",
+      noFileSelected: "Please select a file.",
+      fileAPINotSupported: "The file API is not supported on this browser yet.",
+      invalidJSONFile:
+        "It seems the JSON file you are trying to load is not valid due because it does not match whylogs format. Please check the file validity and try again.",
```
mrmcduff:
```suggestion
        "The JSON file you are trying to load is does not match the expected whylogs format. Please check the file and try again.",
```

viewer/whylogs-script.js
```diff
@@ -0,0 +1,547 @@
...
+  }
+
+  function scrollToFeatureName(event) {
+    var TABLE_HEADER_OFFSET = 90;
+    var $tableWrap = $(".wl-table-wrap");
+    var featureNameId = event.currentTarget.dataset.featureNameId;
+    var currentOffsetTop = $tableWrap.scrollTop();
+    var offsetTop = $('[data-feature-name="' + featureNameId + '"]').offset().top;
+
+    // $tableWrap.scrollTop(currentOffsetTop + offsetTop - TABLE_HEADER_OFFSET);
```
mrmcduff: Remove commented line!

mrmcduff (changes requested):
Great work! We do need to address a few of the suggestions (typo / text changes) and download the minified versions of the external libraries.

viewer/index.html
phantomgoose: i think it doesn't matter, as long as it doesn't change (very often). Sam might have better ideas, but we could do something like `whylogs_static_html_page` even. Im honestly not sure what the best practices around this are, I just know we track this in Heap :)

viewer/whylogs-script.js
```diff
@@ -0,0 +1,547 @@
...
+      noBrowserSupport: "This browser does not seem to support the `files` property of file inputs.",
+      noFileSelected: "Please select a file.",
+      fileAPINotSupported: "The file API is not supported on this browser yet.",
+      invalidJSONFile:
+        "It seems the JSON file you are trying to load is not valid due because it does not match whylogs format. Please check the file validity and try again.",
+    },
+  };
+
+  // HTML Elements
+  var $selectedProfile = $(".wl__selected-profile");
```
phantomgoose: is there a reason to use `var` keyword in this file?

viewer/whylogs-script.js
```diff
@@ -0,0 +1,547 @@
...
+  // Constants and variables
+  var batchArray = [];
+  var featureSearchValue = "";
+  var isActiveInferredType = {};
+  var propertyPanelData = [];
+  var profiles = [];
+  var jsonData = {};
+
+  // Util functions
+  function debounce(func, wait, immediate) {
```
phantomgoose: cant we use lodash? or trying to avoid dependencies?

viewer/whylogs-script.js
phantomgoose: There's an extra `is` here :)

`The JSON file you are trying to load does not match the expected whylogs format. Please check the file and try again`

viewer/index.html
phantomgoose: You guys probably saw this already, but it looks like there's a library for generating static pages with react: https://github.com/react-static/react-static

But yea, jquery is probably a bit easier to work with than vanilla js in this case (especially if we're trying to keep dependencies to a minimum).

viewer/index.html
samgracie: I was thinking something like that, Alex. Let's set the tag to `whylogs_html_viewer`, it will be an important metric to break out.

viewer/index.html
samgracie: Should we include the full libraries or extract the specific parts we need for the html viewer?

viewer/index.html
valer-whylabs: We were discussing react-static, but because we already had something in vanilla and in order to keep dependencies at the minimum, we decided to continue without vanilla.

viewer/whylogs-script.js
valer-whylabs: There is no specific reason for that. I started writing code with ES5 because we are not using any compiler/transpiler and I wanted to have a greater browser support. However, every required browser now supports ES6 by default, so it was my silly mistake.

I would probably recommend a quick refactoring to ES6, because we are mixing arrow functions and regular functions as well. Just to be more consistent.

viewer/whylogs-script.js
valer-whylabs: I can import only `debounce` from Lodash as it will be 2-3 functions only, nothing heavy and complex. However, I was trying to avoid external libraries if they are not necessary.

phantomgoose: i think we'll end up having to write a lot of helper functions, especially if we're just sticking with vanilla/jquery. Adding lodash prolly makes sense, and gives us peace of mind that these utils are implemented correctly and dont need tests, so we can just focus on the page itself.

viewer/whylogs-script.js
phantomgoose: +1 for refactoring to ES6 for readability and consistency!

viewer/index.html
mrmcduff: It's all local; I'm not sure what the size is nor how difficult it is to extract

valer-whylabs: The whole `viewer` directory takes up 4.5MB on the disk. We can reduce it by removing some of the font types, as they only take up 2.2MB.

viewer/whylogs-script.js
valer-whylabs: I would leave it as is for now, because we are using only the debounce util function. Once we need one more function I will import Lodash. What do you think?

mrmcduff: I'm inclined to be fine for this release with just the one instance -- it's important to keep in mind that we will update this as more functionality becomes available.

mrmcduff (approved):

lalmei (approved):
LGTM

PR #230, opened 2021-05-24
Update feature list based on checked inferred types

valer-whylabs:
The bug that is fixed with this PR:
- Feature name list is being filtered as well, when we are changing the inferred type checkboxes.


lalmei (approved):
LGTM

PR #233, opened 2021-05-29
check if it pd.null first

lalmei:
## Description

Checked if the value being tracked is a pd.null first. 
else check if it is None 

### General Checklist

* [x] Documentation updated


andyndang:
This doesn't address the second comment:

```
In addition, when value is None, it's not reflected in SchemaCounters for Null value
```

I'm working on the PR for this already FYI

lalmei:
As discussed This was a WIP.

PR #234, opened 2021-05-29
Fix 232 by storing null_count in schema tracker only

### LINKED ISSUE #232: [bug] Incorrect null count logic
We should use `pd.isnull` for checking null count

In addition, when value is `None`, it's not reflected in SchemaCounters for Null value
https://github.com/whylabs/whylogs/blob/97d12699ac3482f229698872cbe2bff372e63c40/src/whylogs/core/columnprofile.py#L101

andyndang:
## Description

Backwards compatibility: summary still has the null count under "counters"

For existing profiles, we can workaround by adding the two values together.
For new profiles after this PR, the `null_count` will always come from SchemaTracker and not Counters

Addressing https://github.com/whylabs/whylogs/issues/232

### General Checklist

* [ ] Documentation updated


lalmei (approved):
LGTM

PR #236, opened 2021-06-10
String features such as lenght, word count, character pos tracking

lalmei:
## Description

adding String features such as lenght, word count, character pos tracking

### General Checklist

* [x] Documentation updated


proto/src/messages.proto
```diff
@@ -71,13 +71,22 @@ message NumbersMessage {
...
+}
+
 message StringsMessage {
   int64 count = 1;
 
   // sketches
   bytes theta = 2;
   bytes items = 3;
   bytes compact_theta = 4;
+  NumbersMessage length =5;
```
andyndang:
```suggestion
  NumbersMessage length = 5;
```

proto/src/messages.proto
```diff
@@ -71,13 +71,22 @@ message NumbersMessage {
...
+  NumbersMessage token_length= 6;
```
andyndang:
```suggestion
  NumbersMessage token_length = 6;
```

proto/src/messages.proto
```diff
@@ -71,13 +71,22 @@ message NumbersMessage {
...
+  CharPosMessage char_pos_tracker =7;
```
andyndang:
```suggestion
  CharPosMessage char_pos_tracker = 7;
```

proto/src/summaries.proto
```diff
@@ -43,9 +43,18 @@ message FrequentItemsSummary {
   repeated FrequentItem items = 1;
 }
 
+
+message CharPosSummary {
+  string character_list=1;
```
andyndang, line 48:
```suggestion
  string character_list= 1;
```

proto/src/summaries.proto
```diff
@@ -43,9 +43,18 @@ message FrequentItemsSummary {
...
+  repeated NumberSummary char_pos_map =2;
```
andyndang:
```suggestion
  repeated NumberSummary char_pos_map = 2;
```

proto/src/summaries.proto
```diff
@@ -43,9 +43,18 @@ message FrequentItemsSummary {
...
+}
+
 message StringsSummary {
   UniqueCountSummary unique_count = 1;
   FrequentStringsSummary frequent = 2;
+  NumberSummary length =3;
```
andyndang, line 55:
```suggestion
  NumberSummary length = 3;
```

proto/src/summaries.proto
```diff
@@ -43,9 +43,18 @@ message FrequentItemsSummary {
...
+  NumberSummary token_length=4;
```
andyndang, line 56:
```suggestion
  NumberSummary token_length = 4;
```

src/whylogs/core/statistics/stringtracker.py
```diff
@@ -0,0 +1,248 @@
...
+from .numbertracker import NumberTracker
+
+MAX_ITEMS_SIZE = 128
+MAX_SUMMARY_ITEMS = 100
+
+logger = logging.getLogger(__name__)
+
+
+class CharPosTracker:
+    def __init__(self, character_list: str = "abcdefghijklmnopqrstuvwzyz0123456789-"):
```
andyndang: coding: extract this to a constant.

Looking at this list it looks very limited. We should support ASCII characters at least

src/whylogs/core/statistics/stringtracker.py
```diff
@@ -0,0 +1,245 @@
...
+
+        self.character_list = character_list
+        self._character_list_map = {x: idx for idx, x in enumerate(character_list)}
```
andyndang: We're early create an entry for each character here. This is not memory efficient since many of them would be empty in case of say, zip code. It's better to create them on access

src/whylogs/core/statistics/stringtracker.py
```diff
@@ -0,0 +1,245 @@
...
+
+        self.char_pos_map = [NumberTracker() for _ in self.character_list]
```
andyndang: Same thing

src/whylogs/core/statistics/stringtracker.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+
+    def __init__(
+        self,
+        count: int = None,
+        items: frequent_strings_sketch = None,
+        theta_sketch: ThetaSketch = None,
+        length: NumberTracker = None,
+        token_length: NumberTracker = None,
+        char_pos_tracker=None,
```
andyndang: spacing (styling)

src/whylogs/core/statistics/stringtracker.py
```diff
@@ -0,0 +1,245 @@
...
+
+        self.character_list = character_list
+        self._character_list_map = {x: idx for idx, x in enumerate(character_list)}
+
+        self.char_pos_map = [NumberTracker() for _ in self.character_list]
+
+    def update(self, value):
+
+        for indx, char in enumerate(value):
+            if char in self.character_list:
```
andyndang: FYI `self.character_list` is a list, so this look up is not very efficient. a `set` is more appropriate.

andyndang: Also, we should track the unsupported characters as well

andyndang (commented):
Overall looks good. Feedback is mostly around the char tracker and the behavior (eager creation of trackers) and the supported set of characters

src/whylogs/core/statistics/stringtracker.py
lalmei: Thats a good point, i will add that.

lalmei: Agreed. The list is not large now, but can become quite large later on.

src/whylogs/core/statistics/stringtracker.py
lalmei: Will double check, seems our format/fix missed this and all the protobuf, will look into it.

andyndang: I don't think the formatter looks at protobuf yet btw!

src/whylogs/core/statistics/stringtracker.py
lalmei: If we want to work out of the box we need need a default, since this only instantiated with the column is created, we can have a method there to define the character list which can be pasted with the track API. is that reasonable ?

andyndang: I think we're talking about different things here.
* Out of box character support: yes, we should have a predefined list, and ideally users can customize them
* Memory efficiency: just because a character is listed in the list, we should **eagerly** create a map entry for them. These entries can be lazily created (missing = empty values, otherwise) -> this is what I was trying to address in my comment.

lalmei: done

src/whylogs/core/statistics/stringtracker.py
lalmei: done

lalmei: done

src/whylogs/core/statistics/stringtracker.py
lalmei: changed data structure

src/whylogs/core/datasetprofile.py
```diff
@@ -235,7 +247,7 @@ def track_metrics(
             score_field=score_field,
         )
 
-    def track(self, columns, data=None):
+    def track(self, columns, data=None,character_list=None, token_method=None):
```
andyndang: spacing
```suggestion
    def track(self, columns, data=None, character_list=None, token_method=None):
```

src/whylogs/core/datasetprofile.py
```diff
@@ -291,7 +303,7 @@ def track_array(self, x: np.ndarray, columns=None):
         columns = [str(c) for c in columns]
         return self.track_dataframe(pd.DataFrame(x, columns=columns))
 
-    def track_dataframe(self, df: pd.DataFrame):
+    def track_dataframe(self, df: pd.DataFrame,character_list=None, token_method=None):
```
andyndang:
```suggestion
    def track_dataframe(self, df: pd.DataFrame, character_list=None, token_method=None):
```

src/whylogs/core/datasetprofile.py
```diff
@@ -308,7 +320,7 @@ def track_dataframe(self, df: pd.DataFrame):
 
             x = df[col].values
             for xi in x:
-                self.track(col_str, xi)
+                self.track(col_str, xi,character_list=None, token_method=None)
```
andyndang:
```suggestion
                self.track(col_str, xi, character_list=None, token_method=None)
```

src/whylogs/core/datasetprofile.py
```diff
@@ -864,6 +895,7 @@ def get_dataset_frame(dataset_summary: DatasetSummary, mapping: dict = None):
     quantile = flatten_dataset_quantiles(dataset_summary)
     col_out = {}
     for _k, col in dataset_summary.columns.items():
+        # print(f"col::{col}")
```
andyndang:
```suggestion
```

src/whylogs/core/statistics/stringtracker.py
```diff
@@ -0,0 +1,258 @@
...
+
+class CharPosTracker:
+    def __init__(self, character_list: str = "abcdefghijklmnopqrstuvwzyz0123456789-@!#$%^&*()[]{\}"):
+
+        self.character_list = set(character_list)
+        self.char_pos_map = {}
+
+    def update(self, value):
+        for indx, char in enumerate(value.lower()):
+
```
andyndang, line 49:
```suggestion
```

src/whylogs/core/statistics/stringtracker.py
```diff
@@ -0,0 +1,258 @@
...
+        Args:
+
+            other (CharPosTracker): to be merged
+
+        """
+        new_character_list = self.character_list.union(other.character_list)
+
+        # initialize merged
+        new_char_pos_tracker = CharPosTracker(
```
andyndang: what about the map? can we add a unit test for this?

src/whylogs/core/statistics/stringtracker.py
```diff
@@ -0,0 +1,258 @@
...
+    def merge(self, other):
+        """
+        Merges two Char Pos Frequency Maps
+
+        Args:
+
+            other (CharPosTracker): to be merged
+
+        """
+        new_character_list = self.character_list.union(other.character_list)
```
andyndang, line 52: this look wrong. Should you attempt to de-dupe the characters so that "abc" union with "abc" doesn't result in "abcabc"?

andyndang (changes requested):

andyndang (changes requested):
Adding minor changes

andyndang (changes requested):
GitHub is confusing

andyndang (approved):
LGTM

PR #244, opened 2021-06-25
Dev/bernease/autosegment

bernease:
## Description

Autosegmentation feature. Calculates ideal segments for dataset using information gain.

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/src/whylogs/app/config.py
+++ b/src/whylogs/app/config.py
@@ -131,10 +131,10 @@ class MetadataConfig:
     """
 
     def __init__(
-            self,
-            type: str,
-            output_path: str,
-            path_template: Optional[str] = None,
+        self,
+        type: str,
+        output_path: str,
+        path_template: Optional[str] = None,
     ):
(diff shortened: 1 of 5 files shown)


src/whylogs/app/config.py
```diff
@@ -214,9 +304,11 @@ class SessionConfigSchema(Schema):
         validate=validate.Length(min=1),
         required=True,
     )
+    metadata = fields.Nested(MetadataConfigSchema, required=False)
 
     @post_load
     def make_session(self, data, **kwargs):
+        print(f"config.py, line 311, data: {data}")
```
lalmei:
```suggestion

```

src/whylogs/app/metadata_writer.py
```diff
@@ -0,0 +1,72 @@
+import json
+import os
```
lalmei, line 2:
```suggestion
import os
from smart_open import open
```

PR #248, opened 2021-06-30
NLP metrics

lalmei:
### General Checklist

* [x] Documentation updated


lalmei:
Still need to add a notebook for the overall metrics, and some more tests

src/whylogs/core/metrics/nlp_metrics.py
```diff
@@ -0,0 +1,80 @@
+from typing import List, Union
+
+import jiwer
```
bernease: We decided to try / import for this in our meeting.

bernease (approved):
LGTM

src/whylogs/core/metrics/nlp_metrics.py
lalmei: it was done

PR #250, opened 2021-07-01
Refactor datasetprofile's flatten profile

lalmei:
## Description

Refactor datasetprofile's flatten profile, also added some missing unit tests for viz and remove string typing conversion.

### General Checklist

* [x] Documentation updated

### DIFF
--- a/src/whylogs/core/columnprofile.py
+++ b/src/whylogs/core/columnprofile.py
@@ -111,9 +111,7 @@ class ColumnProfile:
 
             self.string_tracker.update(value, character_list=character_list, token_method=token_method)
 
-        # TODO: Implement real typed data conversion
-        typed_data = TypedDataConverter.convert(value)
-
+        typed_data = value
         if not pd.isnull(typed_data):
             self.cardinality_tracker.update(typed_data)
             self.frequent_items.update(typed_data)
@@ -124,6 +122,7 @@ class ColumnProfile:
             # Note: bools are sub-classes of ints in python, so we should check
             # for bool type first
             self.counters.increment_bool()
+
         self.number_tracker.track(typed_data)
 
         self.constraints.update(typed_data)
--- a/src/whylogs/core/columnprofile.py
+++ b/src/whylogs/core/columnprofile.py
@@ -110,8 +110,9 @@ class ColumnProfile:
         if isinstance(value, str):
 
             self.string_tracker.update(value, character_list=character_list, token_method=token_method)
+        # TODO: Implement real typed data conversion
+        typed_data = TypedDataConverter.convert(value)
 
-        typed_data = value
         if not pd.isnull(typed_data):
             self.cardinality_tracker.update(typed_data)
             self.frequent_items.update(typed_data)
--- a/src/whylogs/core/metrics/nlp_metrics.py
+++ b/src/whylogs/core/metrics/nlp_metrics.py
@@ -57,6 +57,8 @@ class NLPMetrics:
 
         return merged_nlp_metrics
 
+        return merged_nlp_metrics
+
     def to_protobuf(
         self,
     ) -> NLPMetricsMessage:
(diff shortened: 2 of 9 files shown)


src/whylogs/core/metrics/nlp_metrics.py
```diff
@@ -0,0 +1,80 @@
...
+            predictions (Union[str,List[str]]):
+            targets (Union[List[str],str]):
+
+        """
+        if transform:
+            mes = jiwer.compute_measures(truth=targets, hypothesis=predictions, truth_transform=transform, hypothesis_transform=transform)
+        else:
+            mes = jiwer.compute_measures(truth=targets, hypothesis=predictions)
+
+        self.mer.update(mes["mer"])
```
andyndang: Can we store these strings as constants?

src/whylogs/core/metrics/nlp_metrics.py
```diff
@@ -0,0 +1,80 @@
+from typing import List, Union
+
+import jiwer
```
andyndang: Can we do `try... import... except` here? I don't want whylogs to have a hard dependency o `jiwer`

src/whylogs/core/metrics/nlp_metrics.py
```diff
@@ -0,0 +1,80 @@
...
+
+from whylogs.core.statistics import NumberTracker
+from whylogs.proto import NLPMetricsMessage
+
+
+class NLPMetrics:
```
andyndang: Can you add some basic tests for this, especially around merge and serialization/deserialization?

andyndang (changes requested):

src/whylogs/core/metrics/nlp_metrics.py
lalmei: sure

src/whylogs/core/metrics/nlp_metrics.py
lalmei: it is a light dependency, we can seperate when we work with that specific issue.

src/whylogs/core/metrics/nlp_metrics.py
lalmei: merge was added, will add one for serialization and deseriallization

src/whylogs/core/columnprofile.py
```diff
@@ -111,9 +111,7 @@ def track(self, value, character_list=None, token_method=None):
 
             self.string_tracker.update(value, character_list=character_list, token_method=token_method)
 
-        # TODO: Implement real typed data conversion
-        typed_data = TypedDataConverter.convert(value)
-
+        typed_data = value
```
bernease: The TypedDataConverter seems to take a string object and use the yaml parser to convert it to a more appropriate type. I know Isaac had discussed getting rid of this as a default, but was that your goal here?

My understanding is that this change affects the underlying profiling, not just the summary.

bernease (approved):
Approved when the one comment is addressed.

src/whylogs/core/columnprofile.py
andyndang: Good catch. This seems to be a big change to the behavior

lalmei: @andyndang isnt this the behaviour in the java side ?

lalmei: the idea is that string are strings and int are ints. parse ints from string is not the behavours that most users want, for example " 00123" is usually expected as a string for so that it can keep track of the leading zeros, think upcs. Once it is converted as a feature then it is consumed into own format, as a string , int or encoding as the model user defines it.

src/whylogs/core/metrics/nlp_metrics.py
lalmei: done

src/whylogs/core/columnprofile.py
andyndang: Yes currently it is. We are trying to do type-detection. This is a significant behavior change and I would not include it here

lalmei: moving to a seperate PR

lalmei: done

andyndang (approved):
LGTM

PR #251, opened 2021-07-01
WIP: simple log api

lalmei:
## Description

simple api wrapper. 

### General Checklist

* [ ] Documentation updated


src/whylogs/app/log.py
```diff
@@ -0,0 +1,21 @@
+from whylogs.app.logger import Logger
+from whylogs.app.session import _LoggerKey, get_or_create_session
+
+
+def log(data: dict, dataset_name: str = "", **kwargs) -> Logger:
+
+    session = get_or_create_session()
+
+    logger_key = str(_LoggerKey(dataset_name=dataset_name, dataset_timestamp=session._session_time, session_timestamp=session._session_time, **kwargs))
```
andyndang, line 9: This feels wrong. Why is this method managing the internal states of `session`? We shouldn't touch private variables unless we're mucking around with external libraries. In this case `session` is within the repo so why are we touching its private variables?

andyndang (commented):
I'm not a fan of this change overall:
* We're modifying internal state (breaking encapsulation)
* Too many side effects from the method itself
* I'm not fully sure if this solves a specific problem that users have requested?

src/whylogs/app/log.py
lalmei: session time and dataset time lose their meaning if folks are logging straight from whylogs.log(), this is simply tying both variables together. I agree we need to be a bit more careful so it doesnt propagate into other use cases.

PR #253, opened 2021-07-06
Add example showing custom tokenizer for string tracking

lalmei:
## Description

Demo of custom tokenizer calls, and push along minor bug fixes

### General Checklist

* [x] Documentation updated


src/whylogs/app/logger.py
```diff
@@ -299,6 +299,8 @@ def log(
         features: Optional[Dict[str, any]] = None,
         feature_name: str = None,
         value: any = None,
+        character_list: str = None,
```
andyndang, line 302: should the character list be tied to the profile/logger? It seems wrong to have this change across different `.log` call on the same logger?

lalmei: yeah it needs to stay the same otherwise there is an warning in string tracker. I can update to raise an error.

lalmei: This logic is already part of the original string tracker PR.

bernease (approved):
As discussed in person, this API stays but we'll think more carefully about it in the API redesign in the future.

lalmei:
There was no change in the api here, it was already approved, api changes are in the roadmap already

PR #255, opened 2021-07-07
Autosegmentation

lalmei:
## Description

Autosegmentation

### General Checklist

* [x] Documentation updated


src/whylogs/features/autosegmentation.py
```diff
@@ -0,0 +1,105 @@
+from typing import Dict, List, Optional, Union
+
+import numpy as np
+import pandas as pd
+
+
+def _entropy(series: pd.Series, normalized: bool = True):
```
andyndang: Can we have some unit test for these methods?

lalmei: will do

tests/unit/features/test_autosegmentation.py
```diff
@@ -0,0 +1,51 @@
+import pandas as pd
+import pytest
+
+from whylogs.features.autosegmentation import _entropy, _weighted_entropy,_estimate_segments
+
+# import numpy as np
+
+
+def test_entropy_ints():
+    series = pd.Series([0, 1, 2, 4])
```
andyndang, line 14: can you add tests for empty series as well? Just make sure we handle edge cases

andyndang (approved):
LGTM

PR #257, opened 2021-07-12
Fix merging for non-tracked character positions

lalmei:
## Description

Fix merging of non-tracked characters. 

### General Checklist

* [x] Documentation updated


cswarth (changes requested):
Can also you check for this at the end of  test_stringtracker.py::test_merge please.

cswarth (approved):
LGTM

PR #260, opened 2021-07-19
Check additional attribute in Logger

lalmei:
## Description

Add additional  check to `Logger` instantiation attributes 
plus speed up multiple loggers unit test.

### General Checklist

* [x] Documentation updated


andyndang (approved):
LGTM

PR #269, opened 2021-07-29
Remove default metadata writer

lalmei:
## Description

While metadata writer config was not required, a default one was being generated. We removed default local metadata writer. 
### General Checklist

* [x] Documentation updated

### DIFF
--- a/src/whylogs/app/config.py
+++ b/src/whylogs/app/config.py
@@ -211,7 +211,7 @@ class SessionConfig:
         project: str,
         pipeline: str,
         writers: List[WriterConfig],
-        metadata: Optional[MetadataConfig] = False,
+        metadata: Optional[MetadataConfig] = None,
         verbose: bool = False,
         with_rotation_time: str = None,
         cache_size: int = 1,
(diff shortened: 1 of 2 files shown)


src/whylogs/app/config.py
```diff
@@ -221,8 +221,6 @@ def __init__(
         self.pipeline = pipeline
         self.verbose = verbose
         self.writers = writers
-        if not metadata:
```
andyndang, line 224: Can't comment on unchanged line. But can you update line 214 as well:

```        
metadata: Optional[MetadataConfig] = False,
```

should be
```
        metadata: Optional[MetadataConfig] = None,
```

src/whylogs/app/session.py
```diff
@@ -414,7 +414,9 @@ def session_from_config(config: SessionConfig) -> Session:
     Construct a whylogs session from a `SessionConfig`
     """
     writers = list(map(lambda x: writer_from_config(x), config.writers))
-    metadata_writer = metadata_from_config(config.metadata)
+    metadata_writer = None
```
andyndang, line 417: Can't comment on line 99, but can you make this optional?

```
        metadata_writer: MetadataWriter,
```

```
        metadata_writer: Optional[MetadataWriter]=None,
```

andyndang (commented):
A few minor comments

andyndang (approved):
LGTM

PR #275, opened 2021-08-07
bump version in tag and mainline and publish distribution

lalmei:
## Description

automated the version bump and pypi publish during release and pre-release

### General Checklist

* [x] Documentation updated

### DIFF
--- a/.github/workflows/push_release.yml
+++ b/.github/workflows/push_release.yml
@@ -34,11 +34,36 @@ jobs:
           ${{ runner.os }}-pip-
     - name: Install dependencies
       run: make install
-    
+
+    - name: Autobump version
+      run: |
+        # from refs/tags/v1.2.3 get 1.2.3
+        VERSION=$(echo $GITHUB_REF | sed 's#.*/v##')
+        echo $VERSION
+        git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+        git config --local user.name "github-actions[bot]"
+        poetry run bump2version release --verbose --commit --no-tag --new-version $VERSION
+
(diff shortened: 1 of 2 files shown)


.github/workflows/push_release.yml
```diff
@@ -34,11 +34,36 @@ jobs:
...
-    
+
+    - name: Autobump version
+      run: |
+        # from refs/tags/v1.2.3 get 1.2.3
+        VERSION=$(echo $GITHUB_REF | sed 's#.*/v##')
+        echo $VERSION
+        git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+        git config --local user.name "github-actions[bot]"
+        poetry run bump2version release --verbose --commit --no-tag --new-version $VERSION
```
naddeoa, line 45: IIRC, I added targets to the makefile for bumping. Are any of those sufficient? I was trying to get it to the point where we could do anything we want through make so we wouldn't have to copy/paste commands out of the CI files to reproduce them locally.

lalmei: yeah this is to manually bump the version based on the tagging used for the release. second I added the auto-commit option and to make sure it didn't tag since the branch is already tagged (irrespective of the bumpversion config file)

lalmei: we can add it to the makefile to clean it up but it needs to take a argument

lalmei: In other words, this matches the tag we use for the release (in the github release notes) those are set automatically to bump it by patch by the release drafter.

cswarth (approved):
LGTM

PR #295, opened 2021-08-27
Stop JSONifying data. Bump schema minor version to 2

andyndang:
## Description

 Instead, we just take a crude `str` approach. This ensures that we have consistent data in various environments.

 The previous behavior: the logic adds double quotes to string values

### General Checklist

* [ ] Documentation updated


cswarth (approved):
LGTM

PR #300, opened 2021-09-02
Add custom metric plugin support src/whylogs/v2 folder for experimental changes

jamie256:
Experimental V2 changes to support metric plugins.

See https://github.com/whylabs/whylogs/blob/dev/jamie/v2_experimental/examples/custom_metrics.ipynb for an initial example.

## Description

* Added an abstract base class, dataclass: MetricPlugin which helps define some of the serialization and shape required to integrate with whylogs DatasetProfile and logging.
  * Note that the default serialization packs the dataclass fields into a protobuf Struct, but this behavior can be extended by authors of plugins if they desire more control over how their metrics are serialized.
  *  Packing a metric into a proto message is how we integrate with our logging and serialization logic so there is some design tension between the shape of proto messages and supporting the flexibility of custom metrics.
  * Choosing the type to instantiate from a serialized proto PluginMessage is based on the environments in scope subclasses of MetricPlugin and then also must match at least the class name specified.
  *  Current metrics plugins are scoped to columnar as a first step.
 * New proto message definitions are works in progress, but sketching out a few ideas here that might work and give us the flexibility we need.
 * Logger has a new method to add metric plugins and read them.

### General Checklist

* [ ] Documentation updated


src/whylogs/v2/core/metrics/metric_plugin.py
```diff
@@ -0,0 +1,137 @@
...
+    def to_summary(self):
+        return self.serialize()
+
+    @staticmethod
+    def deserialize(data: bytes, enforce_strict_name_matching: bool = False) -> "MetricPlugin":
+        message = PluginMessage.FromString(data)
+        return MetricPlugin.from_protobuf(message, enforce_strict_name_matching)
+
+    def to_json(self) -> dict:
+        return asdict(self)
```
ramannanda9: You might want to distinguish top-level fields like name and have subclass params as 'params' to avoid clashes and do the same for deserialization.

```python
return {'cls_name': name() , 'params': asdict(self)}
```

jamie256: Good call, I'll fix that up to handle the top level fields better. Thanks for taking a look!

ramannanda9: Also just as an option try exploring https://lidatong.github.io/dataclasses-json/  for custom serializer and deserializers

jamie256: dataclasses_json handles nested classes and private fields better, thanks. I created a dev release from this to try out:
https://pypi.org/project/whylogs/0.6.4.dev0/

You can use try the example notebook: https://github.com/whylabs/whylogs/blob/dev/jamie/v2_experimental/examples/custom_metrics.ipynb

jamie256:
Dev release of the above is here: https://pypi.org/project/whylogs/0.6.5.dev1/ 

new examples in the notebook here:
https://github.com/whylabs/whylogs/blob/dev/jamie/v2_experimental/examples/custom_column_profiles.ipynb

PR #302, opened 2021-09-03
Fix model metric model_type switching and update tests

### LINKED ISSUE #301: Metric model type switches to classification when whole numbers are logged
### Summary
The model profile's metric model type is inferred based on the type of the data logged, and if a regression model happens to log data that are all consistent with integers, e.g. [118.0, 1.0, 2.0] then the metric model infers that to be a classification model and switches.

Expected behavior would be:
1) not to switch the model_type after the model type was inferred.
2) Rely more explicitly on the API rather than trying to infer the model type from the data values, because the types can overlap, and the model types can also overlap. For example the data: [1.0, 600.0] could be continuous, binary, or multiclass

### Steps to Reproduce it

``` python
# 1) given a regression model log some data such as: targets [1.0, 118.3]
logger.log_metrics(targets, prediction)
```

that results in regression model type being inferred, but now lets say we log the values:

``` python
# 2) using same regression model log some more data: targets [1.0, 112.0, 203.0]
logger.log_metrics(targets, prediction)
```
Because the values are all consistent with integers the model type is inferred to be classification and switched, even though model type was already set as regression.

### Example

``` python
# 2) using a regression model log some float data that is consistent with integer values: targets [1.0, 112.0, 203.0]
logger.log_metrics(targets, prediction)
```

The profile's metric.model_type is overwritten to classification

The profile's metric.model_type should not be overriden if initialized to a specific type, and we should let the API shape infer type (e.g. scores --> classification, not infer from whole number floats).

jamie256:
## Description

Fixes #301 by avoiding mutation of model metric's model_type via when log_metrics() is called post model type initialization and removing some of the automatic model type from data inference code in favor of relying more on the method called. E.g. callers should either pass in scores parameter and/or set the model_type to log classification metrics.

Removed some of the NLP model metrics code as that is not fully implemented yet to simplify the behavior.

Added some new tests and updated the model profile tests.

### General Checklist

* [x ] Tests added for this feature/bug
* [ x] Conform by the style guides, by using formatter
* [ ] Documentation updated
* [ x] (optional) Please add a label to your PR

### DIFF
--- a/examples/Logging.ipynb
+++ b/examples/Logging.ipynb
@@ -3,16 +3,15 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
-   "outputs": [],
    "source": [
     "import os.path\n",
     "import pandas as pd"
-   ]
+   ],
+   "outputs": [],
+   "metadata": {}
   },
   {
    "cell_type": "markdown",
-   "metadata": {},
    "source": [
     "In this notebook, we will explore how to generate logs using the whylogs Python library. \n",
     "\n",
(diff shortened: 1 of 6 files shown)


andyndang (approved):
LGTM

examples/Logging.ipynb
```diff
@@ -23,147 +22,149 @@
...
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
-   "outputs": [],
    "source": [
     "frequencies = summary['frequent_strings']\n",
-    "frequencies['num_sats']"
-   ]
+    "frequent_strings = [key for key in frequencies.keys()]\n",
+    "print(frequent_strings)\n",
```
jamie256, line 160: This should avoid errors reading a specific frequent string from the lending club data.

jamie256:
Added a fix for #303 to get a clean notebook test.

PR #307, opened 2021-09-10
Fixes #306 log rotation serialization concurrency bug

### LINKED ISSUE #306: Race condition with aggressive log rotation
### Summary

The number of features seen in the dataset summaries can be truncated if log rotation occurs while the first dataframe is being logged and before it has updated the metrics for all columns.

### Steps to Reproduce it

Configure log rotation with aggressive timing (e.g. 1s intervals and log a large dataframe both in the number of columns and rows).

Less than all the features can be serialized in the dataset summary if log rotation occurs before the first track_dataframe call has completed.

We should not serialize the dataset profiles while they are being updated, so that we get all or none with respect to a datasets columns.

jamie256:
## Description
Fixes #306 with a reentrant lock in Datasetprofiles to prevent serialization of the profile while still computing metrics or logging a dataframe. The lock prevents a race condition between the main thread logging data and serialization by io threads if using aggressive log rotation intervals.

### General Checklist

* [ x] Tests added for this feature/bug
* [ x] Conform by the style guides, by using formatter
* [ ] Documentation updated
* [ x] (optional) Please add a label to your PR


andyndang:
While this is an okay change, a bigger question I have is it seems we are not creating new DatasetProfile object, but instead keep re-using the old ones when we serialize the log. That seems to be a problematic behavior as well IMO.

lalmei:
any idea why it is failing in python 3.6 ?

PR #315, opened 2021-09-24
[IMPROVEMENT] Switch image profiling to using ImageStats

jamie256:
For better perf and more meaningful metrics over sets of images rather than over pixels in a single image

* Turn off the previous Hue Staturation Brightness per pixel transforms and metrics.
* Switch to using ImageStat.Stats on image converted to HSV to log mean and stddev of Hue, Saturation, Brightness over the image
* Updated tests and notebook so they still work
* Added in a few more files of flowers so we can log more than one image to get some distributions in example notebook

## Description

We had feedback that image logging was too slow and resulted in confusingly high metric counts (because we were calculating these per pixel). We can still calculate statistics over all the pixels but using vectorized methods on PIL already to compute these, then add them alongside metadata as part of default logging.

### General Checklist

* [ ] Documentation updated


andyndang (approved):
LGTM

Issue #160, opened 2021-02-22
Add regex check on contraints


lalmei:

cswarth:
Can we get some more details of what you would like to see in the enhancement?

jamie256:
This is in!

PR #317, opened 2021-09-25
Add regex pattern matching to ValueConstraint

FelipeAdachi:
## Description

Added MATCH and NOMATCH for regex pattern matching to ValueConstraints.

Intended to be used with string values. Will log as failure whenever a non-string value is encountered.

Relates to issue #160 

### General Checklist

* [x ] Tests added for this feature/bug
* [x ] Conform by the style guides, by using formatter
* [ ] Documentation updated


proto/src/constraints.proto
```diff
@@ -34,11 +36,16 @@ message SummaryConstraintMsg {
...
 
-/* ValueConstraints express a binary boolean relationship between an implied numeric value and a literal.
+/* ValueConstraints express a binary boolean relationship between an implied numeric value and a literal, or between a string value and a regex pattern.
    These are applied to every incoming value that is processed by whylogs.  */
+
+
 message ValueConstraintMsg {
   string name = 1;
-  double value = 2;
+  oneof second_field {
```
cswarth, line 45: nice!

cswarth (approved):
LGTM, very nice

src/whylogs/core/statistics/constraints.py
```diff
@@ -31,6 +32,8 @@
     Op.NE: lambda x: lambda v: v != x,
     Op.GE: lambda x: lambda v: v >= x,
     Op.GT: lambda x: lambda v: v > x,  # assert incoming value 'v' is greater than some fixed value 'x'
+    Op.MATCH: lambda x: lambda v: re.match(x, v) is not None,
```
jamie256, line 35: Does re.compile also work here?

tests/unit/core/statistics/test_constraints.py
```diff
@@ -67,6 +67,45 @@ def test_value_constraints(df_lending_club, local_config_path):
     assert report[1][1][0][2] == 50
 
 
+def test_value_constraints_pattern_match(df_lending_club, local_config_path):
+
+    regex_state_abbreviation = r"^[a-zA-Z]{2}$"
+    contains_state = ValueConstraint(Op.MATCH, regex_pattern=regex_state_abbreviation)
+
+    regex_date = r"^[a-zA-Z]{3}-[0-9]{4}$"
+    contains_date = ValueConstraint(Op.NOMATCH, regex_pattern=regex_date)
```
jamie256, line 76: Should this constraint variable should be name something like "not_contains_date" or "no_dates" since it is using Op.NOMATCH?

jamie256 (approved):
Looks good, minor comments.

src/whylogs/core/statistics/constraints.py
cswarth: yes, missed this. Would be good to compile the regex once when constructing the in-memory constraint.

PR #321, opened 2021-09-30
Fix whylogs init

### LINKED ISSUE #311: whylogs CLI is broken
### Summary

cli is missing after pip install

### Steps to Reproduce it

```
whylogs init
```
doesn't work anymore

jamie256:
## Description

Fix #311 by adding poetry scripts section to our pyproject.toml file and referencing the whylogs cli scripts.

Previously the whylogs cli and whylogs-demo were referenced in the setup.cfg as entry points, looks like it might have been missed when migrating to poetry.

You can verify this works locally by running

`poetry install && poetry run whylogs init`

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,10 @@ classifiers = [
     "Topic :: Software Development :: Libraries :: Python Modules"
 ]
 
+[tool.poetry.scripts]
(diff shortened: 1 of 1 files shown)


andyndang (approved):
LGTM

Issue #322, opened 2021-10-01
use re.compile over re.match in regex Constraints


jamie256:
### Summary

Small follow up from code review to use re.compile since the pattern will be known and used repeatedly it makes sense to use the compiled version of the regex.

### Relevant logs and/or screenshots

 see https://github.com/whylabs/whylogs/pull/317

jamie256:
fixed

PR #325, opened 2021-10-03
Changed MATCH and NOMATCH to use re.compile instead of re.match. Chan…

FelipeAdachi:
…ged name of constraint in test_constraints.py to not_contains_date.

## Description

Changed constraints regex pattern matching to use re.compile instead of re.match.
Relates to #322 

Changed constraint name in test_constraints.py to `not_contains_date`, since it's a NOMATCH assertion.

Both changes are from follow-ups from PR #317 

### General Checklist

* [ ] Documentation updated


jamie256 (approved):
LGTM!

PR #326, opened 2021-10-04
Add warning on larger dataframe logging, add perf test and fixture for conditional tests w/ --load

jamie256:
* Also added a test to profile log_dataframe on a dataframe 4x size of lending club 1000 as a baseline for perf.
* Test does not pass but will output some statistics on most expensive calls.

Next would be to integrate the output stats better so we can track this perf over time in an automated way. By default this new test will be skipped when running tests.

## Description

When we add new trackers or features, these can increase the time it takes to log data frame but we didn't have tests to report that or get insight into where time is spent during a log_dataframe() call. Walking through onboarding scenario it can be too long for log_dataframe to return or report progress depending on the size and types of data chosen in an initial data frame. So the first step to addressing this scenario better is to measure what the time is and where time is spent when logging some benchmark datasets. Next we will add some warnings if the amount of data being logged in one call is large and is known to take significant time.

To run the test, include the --load parameter, e.g.

`
poetry run pytest tests/component -o log_cli=true -o log_level=INFO --load
`

```
63319106 function calls (63319047 primitive calls) in 38.848 seconds

   Ordered by: cumulative time
   List reduced from 386 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        2    0.002    0.001   34.841   17.421 /home/jamie/projects/whylogs/src/whylogs/app/logger.py:544(log_dataframe)
        2    0.319    0.160   34.839   17.420 /home/jamie/projects/whylogs/src/whylogs/core/datasetprofile.py:259(track_dataframe)
   604302    0.281    0.000   34.494    0.000 /home/jamie/projects/whylogs/src/whylogs/core/datasetprofile.py:203(track)
   604302    0.326    0.000   34.213    0.000 /home/jamie/projects/whylogs/src/whylogs/core/datasetprofile.py:230(track_datum)
   604302    1.864    0.000   33.860    0.000 /home/jamie/projects/whylogs/src/whylogs/core/columnprofile.py:100(track)
    97768    0.342    0.000   14.637    0.000 /home/jamie/projects/whylogs/src/whylogs/core/statistics/stringtracker.py:211(update)
    97768    1.717    0.000   13.014    0.000 /home/jamie/projects/whylogs/src/whylogs/core/statistics/stringtracker.py:40(update)
   435987    0.145    0.000   10.174    0.000 /home/jamie/projects/whylogs/src/whylogs/core/types/typeddataconverter.py:30(convert)
    97768    0.089    0.000    9.999    0.000 /home/jamie/.cache/pypoetry/virtualenvs/whylogs-dFpe3D9Y-py3.8/lib/python3.8/site-packages/yaml/__init__.py:154(safe_load)
    97768    0.163    0.000    9.910    0.000 /home/jamie/.cache/pypoetry/virtualenvs/whylogs-dFpe3D9Y-py3.8/lib/python3.8/site-packages/yaml/__init__.py:103(load)
```

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/src/whylogs/core/datasetprofile.py
+++ b/src/whylogs/core/datasetprofile.py
@@ -268,11 +268,19 @@ class DatasetProfile:
         # workaround for CUDF due to https://github.com/rapidsai/cudf/issues/6743
         if cudfDataFrame is not None and isinstance(df, cudfDataFrame):
             df = df.to_pandas()
+        element_count = df.size
+        large_df = element_count > 200000
+        if large_df:
+            logger.warning(f"About to log a dataframe with {element_count} elements, logging might take some time to complete.")
+        count = 0
         for col in df.columns:
             col_str = str(col)
 
             x = df[col].values
             for xi in x:
+                count = count + 1
+                if large_df and (count % 200000 == 0):
+                    logger.warning(f"Logged {count} elements out of {element_count}")
                 self.track(col_str, xi, character_list=None, token_method=None)
 
     def to_properties(self):
--- a/tests/component/conftest.py
+++ b/tests/component/conftest.py
@@ -3,14 +3,30 @@ import sys
 
 import pytest
 
-import whylogs
-
 _MY_DIR = os.path.realpath(os.path.dirname(__file__))
 # Allow import of the test utilities packages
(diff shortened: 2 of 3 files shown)


andyndang (approved):
LGTM

PR #327, opened 2021-10-04
Update API token when updated in the env variable and recreate client

### LINKED ISSUE #324: Cannot change API token after setting it
It's set once here when the client is created:

https://github.com/whylabs/whylogs/blob/992369d34e593f403035e8aa9db53e1659675070/src/whylogs/whylabs_client/wrapper.py#L33

jamie256:
## Description
Fixes #324 by updating the _api_key when updated in the environment and adding more verbose output when upload completes.

### General Checklist

* [ ] Documentation updated


andyndang (approved):
LGTM

PR #331, opened 2021-10-05
Fix missing ImagePixelWidth and reduce default set of image tracking

### LINKED ISSUE #330: ImagePixelWidth and IamgePixelLength missing from image logging
### Summary

When logging images we track many exif tags that are often empty in image datasets, and we weren't including the actual image size.

### Steps to Reproduce it

1. log images (such as the tesdata/images)
2. inspect profiles
3. Note that ImageWidth is missing from most images, but ImagePixelWidth is not included in the default set of metadata included in the profile.

Include a smaller set of columns for images such that there is less noise in the default profiles and include the image size.

jamie256:
## Description

Fixes #330 by adding ImagePixelWidth to the default set of image metadata tags we track.
- removed many of the default tags to reduce the noise when image tracking uses default settings.
- updated tests

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/src/whylogs/core/image_profiling.py
+++ b/src/whylogs/core/image_profiling.py
@@ -17,22 +17,9 @@ DEFAULT_IMAGE_FEATURES = []
 
 
 _DEFAULT_TAGS_ATTRIBUTES = [
-    "ImageWidth",
-    "ImageLength",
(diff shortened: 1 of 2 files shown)


andyndang (approved):
LGTM

PR #333, opened 2021-10-12
[API] Simplify WhyLabs writer

andyndang:
## Description

* Simplify WhyLabs writer
* Remove the anonymous session upload logic
* Depcrate start_whylabs_session()
* Relax WriterConfig validation (in case users still have the old config)

### General Checklist

* [ ] Documentation updated


src/whylogs/app/session.py
```diff
@@ -459,12 +460,9 @@ def reset_default_session():
 
 def start_whylabs_session(
     path_to_config: Optional[str] = None,
-    data_collection_consent: Optional[bool] = None,
     report_progress: Optional[bool] = False,
 ):
-    if not data_collection_consent:
-        raise PermissionError("When creating a session that will send data to WhyLabs, data_collection_consent must be set to True")
-
+    warnings.warn("start_whylabs_session() is deprecated. Please sign up for WhyLabs", DeprecationWarning)
```
jamie256, line 465: This looks good for deprecating anonymous sessions, we should call this out in the release.

src/whylogs/app/config.py
```diff
@@ -82,19 +93,17 @@ class WriterConfig:
...
         filename_template: Optional[str] = None,
-        data_collection_consent: Optional[bool] = False,
         transport_parameters: Optional[TransportParameterConfig] = None,
     ):
         self.type = type
         self.formats = formats
         self.output_path = output_path
         self.path_template = path_template
         self.filename_template = filename_template
-        self.data_collection_consent = data_collection_consent
```
jamie256: Much better with optional parameters on formats and output_path.

We might do better to leave the data_collection_consent parameter for a release (with deprecation warning) so we're not breaking callers without giving them some time with the deprecation warning.

jamie256 (commented):
The simplifications we get by using Optional are nice. The S3 example notebook is failing because it is passing the data_collection_consent parameter:

```
E                 2 from whylogs.app.config import WriterConfig, SessionConfig
E                 3 
E           ----> 4 s3_writer_config= WriterConfig(type="s3",formats=["json","flat","protobuf"],
E                 5              output_path="s3://super_awesome_bucket/",
E                 6              path_template="$name/dataset_summary",
E           
E           TypeError: __init__() got an unexpected keyword argument 'data_collection_consent'
```

src/whylogs/app/session.py
andyndang: Will add it to the commit header

jamie256 (approved):
Looks Good!

PR #342, opened 2021-10-19
Make project name and pipeline name optional parameters for session

jamie256:
## Description

The Session constructor requires 'project' and 'pipeline' parameters but we don't have special handling around these parameters other than as concepts to help practitioners organize their logging. Based on feedback in the onboarding scenario these parameters should be optional. Also make writers optional to support this scenario.

### General Checklist

* [ ] Documentation updated


naddeoa:
How are pipelines actually used to help organize logging? Does the name of the pipeline appear in the profile or the profile name on disk? I still don't understand that part.

jamie256:
> How are pipelines actually used to help organize logging? Does the name of the pipeline appear in the profile or the profile name on disk? I still don't understand that part.

Functionally, 'project' and 'pipeline' don't pass through to the profiles, so they don't 'do' anything, they merely live on the session.

naddeoa (approved):

PR #344, opened 2021-10-25
Enable pickling via protobuf

naddeoa:
## Description

We were never able to pickle our dataset profile objects. Our
integration work with Ray made this a pressing issue since Ray depends
heavily on calling pickle on objects. This change will enable pickle by
serializing/deserializing to/from protobuf.

In doing this, I stumbled upon other actual issues, like we apparently
don't deserialize from protobuf correctly in certain cases, like when we
have no set dataset_timestamp. It ends up as a 1970 date. I have two
tests that are omitted that I'm going to work on next to fix that
issue.

### General Checklist

### DIFF
--- a/src/whylogs/core/datasetprofile.py
+++ b/src/whylogs/core/datasetprofile.py
@@ -110,6 +110,13 @@ class DatasetProfile:
         # Store Name attribute
         self._tags["name"] = name
 
+    def __getstate__(self):
+        return self.serialize_delimited()
+
+    def __setstate__(self, serialized_profile):
+        profile = DatasetProfile.parse_delimited_single(serialized_profile)[1]
+        self.__dict__.update(profile.__dict__)
+
     @property
     def name(self):
         return self._tags["name"]
--- a/src/whylogs/core/datasetprofile.py
+++ b/src/whylogs/core/datasetprofile.py
@@ -96,10 +96,12 @@ class DatasetProfile:
             metadata = {}
         if session_id is None:
             session_id = uuid4().hex
-
+        if dataset_timestamp is None:
+            self.dataset_timestamp = datetime.datetime.now(datetime.timezone.utc)
+        else:
+            self.dataset_timestamp = dataset_timestamp
         self.session_id = session_id
         self.session_timestamp = session_timestamp
-        self.dataset_timestamp = dataset_timestamp
         self._tags = dict(tags)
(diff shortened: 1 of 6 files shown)


src/whylogs/core/datasetprofile.py
```diff
@@ -110,6 +110,15 @@ def __init__(
         # Store Name attribute
         self._tags["name"] = name
 
+    def __getstate__(self):
+        return self.serialize_delimited()
+
+    def __setstate__(self, serizlied_profile):
```
jamie256: minor: looks like typo in param name? 'serizlied_profile'

src/whylogs/core/datasetprofile.py
```diff
@@ -110,6 +110,15 @@ def __init__(
...
+        profile = DatasetProfile.parse_delimited_single(serizlied_profile)[1]
+        print("inside")
```
jamie256: remove print statements, you can put some diagnostic message in a logger as debug level if you think that would help.

andyndang: print debugging

jamie256 (commented):
some minor comments

jamie256:
Nice find on the time deserialization issues. I'll review again when those fixes are in and update here.

src/whylogs/util/time.py
```diff
@@ -33,6 +33,6 @@ def from_utc_ms(utc: Optional[int]) -> Optional[datetime.datetime]:
     dt : datetime.datetime
         Datetime object
     """
-    if utc is None:
+    if utc is None or utc == 0:
```
jamie256, line 36: Cool, thanks for moving that here.

tests/core/test_datasetprofile.py
```diff
@@ -0,0 +1,78 @@
...
+
+    profile = logger.profile
+    pickled_profile = pickle.dumps(profile)
+    unpickled_profile: DatasetProfile = pickle.loads(pickled_profile)
+
+    profiles_eq(profile, unpickled_profile)
+
+
+def test_serde_with_dataset_timezone():
+    session = Session("project", "pipeline", writers=[])
```
jamie256, line 42: project, pipeline and writers should all be optional if you wanted to simplify that.

jamie256 (approved):
Looks good!

PR #355, opened 2021-11-12
Put in some guardrails against overly large confusion matrices

jamie256:
## Description

After investigating some larger profile sizes we saw the confusion matrix is currently not limited in size and is likely a mistake when created with more than 256 labels. We need further work to efficiently handle larger numbers of labels, so this change is a simple mitigation until we have that support.

### General Checklist

* [ ] Documentation updated


tests/unit/core/metrics/test_confusion_matrix.py
```diff
@@ -10,6 +12,36 @@
 from whylogs.proto import ScoreMatrixMessage
 
 
+def _generateTestTargetsAndPredictions(label_size: int, output_length, rand_ratio: float):
+    labels = [i for i in range(label_size)]
```
richard-rogers: why not just `labels = range(label_size)` ?

tests/unit/core/metrics/test_confusion_matrix.py
```diff
@@ -10,6 +12,36 @@
...
+    return targets, predictions
+
+
+def _generateTestTargetsAndPredictionsForEachLabel(label_size: int, output_length, rand_ratio: float):
+    labels = [i for i in range(label_size)]
+    targets = []
+    predictions = []
+
+    for i in range(output_length):
+        label = labels[i % label_size]
```
richard-rogers, line 21: Pass label selector in as a callable so these two can be unified?

tests/unit/core/metrics/test_confusion_matrix.py
jamie256: yeah better to simplify, done.

tests/unit/core/metrics/test_confusion_matrix.py
jamie256: I removed one of the methods for now, but good suggestion.

tests/unit/core/metrics/test_confusion_matrix.py
```diff
@@ -157,6 +175,29 @@ def test_confusion_matrix_to_protobuf():
...
 
 
+def test_over_threshold_confusion_matrix():
+    with pytest.raises(ValueError):
+        ConfusionMatrix(range(257))
+
+
+def test_merged_labels_over_threshold_confusion_matrix():
+    labels = range(200)
+    more_labels = [j for j in range(201, 300)]
```
richard-rogers: Why not just `range(201, 300)` ?

tests/unit/core/metrics/test_confusion_matrix.py
```diff
@@ -10,6 +12,22 @@
 from whylogs.proto import ScoreMatrixMessage
 
 
+def _generateTestTargetsAndPredictionsForEachLabel(label_size: int, output_length, rand_ratio: float):
+    # labels = [i for i in range(label_size)]
```
richard-rogers: remove comment?

jamie256: nice catch! thanks.

tests/unit/core/metrics/test_confusion_matrix.py
jamie256: true, done.

richard-rogers (approved):

PR #358, opened 2021-11-18
Fix Black failure

richard-rogers:
Black seems to be pulling in versions of typed-ast that break
unless constrained to typed-ast >= 1.5

Also added my development environment Dockerfile & fixed another doctoring typo.

### General Checklist

* [ ] Documentation updated


jamie256 (approved):
Looks great, thanks!

PR #359, opened 2021-11-18
Use log_async in WhylabsWriter

richard-rogers:
### General Checklist

* [ ] Documentation updated


pyproject.toml
```diff
@@ -35,7 +35,8 @@ pyyaml = ">=5.3.1"
 pandas = ">=1.0.0"
 marshmallow = ">=3.7.1"
 numpy = ">=1.18.0"
-whylabs-client = "@>=0.1"
+whylabs-client = { path = "../whylabs-client-0.1.1.dev0.tar.gz" }
```
jamie256: You can reference the dev release, its on pypi rather than local path.

src/whylogs/whylabs_client/wrapper.py
```diff
@@ -89,7 +89,11 @@ def _upload_whylabs(dataset_profile, dataset_timestamp, profile_path):
         raise EnvironmentError('Missing dataset ID. Specify it via "datasetId" tag or WHYLABS_DEFAULT_DATASET_ID environment variable')
     try:
         with open(profile_path, "rb") as f:
-            log_api.log(org_id=org_id, model_id=dataset_id, dataset_timestamp=dataset_timestamp, file=f)
+            result = log_api.log_async(org_id=org_id, dataset_id=dataset_id) // dataset_timestamp=dataset_timestamp) looks like timestamp is built in whylabs client ?
+            http_response = requests.put(result["upload_url"], data=f.read())
+            assert (
```
jamie256: Rather than assert when the request returns something other than 'ok' (e.g. 200), let's log as error, we don't need to raise in this case we just need to communicate the status, its also inside a try block already and non 200 returns are valid failure cases modeled by the result already.

jamie256 (changes requested):

jamie256:
Another minor thing, your commit message has a typo:
"whylabs writier" should be Whylabs writer or use the class name WhylabsWriter

pyproject.toml
```diff
@@ -35,7 +35,7 @@ pyyaml = ">=5.3.1"
 pandas = ">=1.0.0"
 marshmallow = ">=3.7.1"
 numpy = ">=1.18.0"
-whylabs-client = "@>=0.1"
+whylabs-client = "^0.1.1.dev0"# { path = "../whylabs-client-0.1.1.dev0.tar.gz" }
```
jamie256: remove comment

jamie256 (changes requested):
Minor comments

jamie256 (approved):
Looks Good!

PR #365, opened 2021-11-30
Fix exception related to logging array like column values

### LINKED ISSUE #363: Column profiler crashes when column contains lists
```
  File "/Users/arnov/Documents/code/project/src/utils/monitoring.py", line 20, in log_dataframe
    ylog.log_dataframe(df)
  File "/Users/arnov/.pyenv/versions/imputations/lib/python3.8/site-packages/whylogs/app/logger.py", line 569, in log_dataframe
    self._profiles[-1]["full_profile"].track_dataframe(df)
  File "/Users/arnov/.pyenv/versions/imputations/lib/python3.8/site-packages/whylogs/core/datasetprofile.py", line 293, in track_dataframe
    self.track(col_str, xi, character_list=None, token_method=None)
  File "/Users/arnov/.pyenv/versions/imputations/lib/python3.8/site-packages/whylogs/core/datasetprofile.py", line 229, in track
    self.track_datum(columns, data)
  File "/Users/arnov/.pyenv/versions/imputations/lib/python3.8/site-packages/whylogs/core/datasetprofile.py", line 247, in track_datum
    prof.track(data, character_list=None, token_method=None)
  File "/Users/arnov/.pyenv/versions/imputations/lib/python3.8/site-packages/whylogs/core/columnprofile.py", line 109, in track
    if pd.isnull(value):
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

The column looks like this:

```
df['codes']
0       [6420, 68204, 68203, 6832, 6810]
1                                 [6630]
2                          [4726, 47789]
3             [6420, 70102, 66292, 6612]
4             [6420, 70102, 64923, 6612]
                     ...
696                               [6420]
697    [6420, 70102, 64923, 70101, 6612]
698                              [70221]
699                        [49391, 4932]
700                [47713, 47712, 47711]
Name: codes, Length: 701, dtype: object
```

The problem is that pd.isnull() evaluates for each element of the list, and `if` tries to convert that to one boolean, which doesn't work

jamie256:
## Description
Fixes #363 - whylogs should handle array like column values and log these as unknown types until we support these as first class types.

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/src/whylogs/core/types/typeddataconverter.py
+++ b/src/whylogs/core/types/typeddataconverter.py
@@ -64,7 +64,7 @@ class TypedDataConverter:
         dtype : TYPES
         """
         dtype = TYPES.UNKNOWN
-        if pd.isnull(typed_data):
+        if typed_data is None:
             dtype = TYPES.NULL
         elif isinstance(typed_data, bool):
             dtype = TYPES.BOOLEAN
@@ -74,4 +74,10 @@ class TypedDataConverter:
             dtype = TYPES.INTEGRAL
         elif isinstance(typed_data, str):
             dtype = TYPES.STRING
+        if isinstance(typed_data, list) or isinstance(typed_data, np.ndarray) or isinstance(typed_data, pd.Series):
+            if all(pd.isnull(typed_data)):
+                dtype = TYPES.NULL
+        elif pd.isnull(typed_data):
+            dtype = TYPES.NULL
+
         return dtype
--- a/src/whylogs/core/types/typeddataconverter.py
+++ b/src/whylogs/core/types/typeddataconverter.py
@@ -49,6 +49,17 @@ class TypedDataConverter:
         return data
 
     @staticmethod
+    def _is_array_like(value):
+        return isinstance(value, list) or isinstance(value, np.ndarray) or isinstance(value, pd.Series)
+
+    @staticmethod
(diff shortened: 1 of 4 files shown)


src/whylogs/core/types/typeddataconverter.py
```diff
@@ -74,4 +74,10 @@ def get_type(typed_data):
             dtype = TYPES.INTEGRAL
         elif isinstance(typed_data, str):
             dtype = TYPES.STRING
+        if isinstance(typed_data, list) or isinstance(typed_data, np.ndarray) or isinstance(typed_data, pd.Series):
```
richard-rogers: Seems like duplicate of `_is_array_like()` ?

jamie256: It is the same condition, but we are going to soon replace/remove the typeddataconverter so I didn't home the method there and was trying for a targeted scope fix. We could pull the _is_array_like out of ColumnProfile but I think it is better to focus on the next version of this API over growing the type inspection logic.

jamie256: But I think you are right, its still better to move this to typeddataconverter than to duplicate the condition even, udpated.

richard-rogers (approved):

PR #366, opened 2021-11-30
Created ComplexSummaryConstraint that extends SummaryConstraints. Def…

MilenaTrajanoska:
…ined stddev_between_constraint as a function. Implemented simple unit tests for checking the functionality of the ComplexSummaryConstraint.

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/src/whylogs/core/statistics/constraints.py
+++ b/src/whylogs/core/statistics/constraints.py
@@ -356,7 +356,11 @@ class SummaryConstraints:
         v = [c.to_protobuf() for c in self.constraints.values()]
         if len(v) > 0:
             scmsg = SummaryConstraintMsgs()
-            scmsg.constraints.extend(v)
+            for msg in v:
+                if isinstance(msg, SummaryConstraintMsgs):
+                    scmsg.constraints.extend(msg.constraints)
+                else:
+                    scmsg.constraints.extend([msg])
             return scmsg
         return None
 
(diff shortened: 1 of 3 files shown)


src/whylogs/core/statistics/constraints.py
```diff
@@ -438,3 +444,56 @@ def report(self):
         l1 = [(k, v.report()) for k, v in self.value_constraint_map.items()]
         l2 = [(k, s.report()) for k, s in self.summary_constraint_map.items()]
         return l1 + l2
+
+
+class ComplexSummaryConstraint(SummaryConstraints):
```
jamie256, line 449: Overall this is nice pattern. I think we need merge defined here if we were to push this (and some checks to make sure we are merging two identical ComplexSummaryConstraints in terms of their components), but let's hold off on the complex constraint approach for the in-between use case, and bring this back if we need it in later constraints.

jamie256 (commented):
I like the approach, but I also want to get as far as we can without introducing the concept of a new type to our users, so in this case the ComplexSummaryConstraint is something I'd like to hide from users and have as an implementation detail.

We likely will need this pattern for more complex scenarios, but the in between constraint seems simple enough we could support as first class operation (and then maybe also have some optimizations around that in terms of evaluating?).

src/whylogs/core/statistics/constraints.py
MilenaTrajanoska: Sure, sounds good.

PR #369, opened 2021-12-02
Added BTWN operation in Op enum, new Op.BTWN lambda summary function,…

pecop2:
… edit SummaryConstraint class and SummaryConstraintMsg accordingly, unit tests for BTWN

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/proto/src/constraints.proto
+++ b/proto/src/constraints.proto
@@ -32,10 +32,10 @@ message SummaryConstraintMsg {
   oneof second {
      string second_field = 3;
      double value = 4;
-     SummaryBetweenConstraintMsg between = 5;
+     SummaryBetweenConstraintMsg between = 7;
   }
-  Op op = 6;
-  bool verbose = 7;
+  Op op = 5;
+  bool verbose = 6;
 }
 
 message SummaryBetweenConstraintMsg {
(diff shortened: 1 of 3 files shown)


proto/src/constraints.proto
```diff
@@ -31,9 +32,22 @@ message SummaryConstraintMsg {
   oneof second {
      string second_field = 3;
      double value = 4;
+     SummaryBetweenConstraintMsg between = 5;
   }
-  Op op = 5;
-  bool verbose = 6;
+  Op op = 6;
+  bool verbose = 7;
```
jamie256: This will be a breaking change if we change the codes for protobuf files because previous versions serialized will have the old codes. In this case I think its ok because we shouldn't have existing profiles with constraints since they are new, but we need to be careful of changes to protobuf files: adding things is ok, but need to preserve the existing codes. So lets use '7' for the between field and then leave 'op' and 'verbose' unchanged?

jamie256 (commented):
Overall looks good. Can we change the codes so that old codes remain unchanged in proto files?

proto/src/constraints.proto
pecop2: Yes, I think it wouldn't be a problem.

jamie256 (approved):
LGTM!

PR #375, opened 2021-12-14
ValueConstraint: Assert column values in set

pecop2:
## Description

IN operator, new field for the set in ValueConstraintMsg, checks for value errors, adapting to_protobuf and from_protobuf.
Function columnValuesInSetConstraint() and tests for the changes

### General Checklist

* [ ] Documentation updated


src/whylogs/core/statistics/constraints.py
```diff
@@ -86,6 +88,10 @@ def __init__(self, op: Op, value=None, regex_pattern: str = None, name: str = No
         self.op = op
         self.total = 0
         self.failures = 0
+
+        if (isinstance(value, set) and op != Op.IN) or (not isinstance(value, set) and op == Op.IN):
```
richard-rogers: `isinstance(value, set) != (op == Op.IN)`

tests/unit/core/statistics/test_constraints.py
```diff
@@ -401,3 +408,49 @@ def test_max_between_constraint_invalid():
...
+    cvisc2 = columnValuesInSetConstraint(value_set={3, 4, 5})
+    with pytest.raises(AssertionError):
+        cvisc1.merge(cvisc2)
+
+
+def test_merge_values_in_set_constraint_same_value_set():
+    cvisc1 = columnValuesInSetConstraint(value_set={"abc", "b", "c"})
+    cvisc2 = columnValuesInSetConstraint(value_set={"abc", "b", "c"})
+    merged = cvisc1.merge(cvisc2)
+    print(merged.report())
```
richard-rogers: Should be an assert?

src/whylogs/core/statistics/constraints.py
```diff
@@ -568,3 +589,12 @@ def maxBetweenConstraint(lower_value=None, upper_value=None, lower_field=None, u
 
 def maxLessThanEqualConstraint(value=None, field=None, verbose=False):
     return SummaryConstraint("max", Op.LE, value=value, second_field=field, verbose=verbose)
+
+
+def columnValuesInSetConstraint(value_set: "set", verbose=False):
```
richard-rogers: Should this be `value_set: Set[Any]` ?

tests/unit/core/statistics/test_constraints.py
```diff
@@ -401,3 +408,49 @@ def test_max_between_constraint_invalid():
...
+
+
+def test_serialization_deserialization_values_in_set_constraint():
+    val_set = {"abc", 1, 2}
+    cvisc = columnValuesInSetConstraint(value_set=val_set)
+
+    cvisc.from_protobuf(cvisc.to_protobuf())
+    json_value = json.loads(message_to_json(cvisc.to_protobuf()))
+
+    print(json_value)
```
richard-rogers: remove `print`?

proto/src/constraints.proto
```diff
@@ -32,10 +34,10 @@ message SummaryConstraintMsg {
   oneof second {
      string second_field = 3;
      double value = 4;
-     SummaryBetweenConstraintMsg between = 5;
+     SummaryBetweenConstraintMsg between = 7;
   }
-  Op op = 6;
-  bool verbose = 7;
+  Op op = 5;
+  bool verbose = 6;
```
richard-rogers, line 40: We want to make sure these are backwards compatible.

pecop2: This in fact is a fix for: https://github.com/whylabs/whylogs/commit/f90b45ae53d50ffde66beec8f4e6184061beef32#diff-3ead279fdd02046f469418a748586aa737b3a3cf99be981b99db1bee165a2a22L35
So now editing them back like they previously were.

jamie256: ahh makes sense. While constraints are a new feature, we want to make sure we don't break deserializing an older protobuf serialized filed, thanks for fixing it back to how they previously were.

tests/unit/core/statistics/test_constraints.py
jamie256: Having debug/diagnostic output as part of a test is really useful, maybe replace the print statement with something like this:
https://github.com/whylabs/whylogs/blob/09e0289175ddcffb05a0e37953eed5e8347a1ca4/tests/component/test_performance_log_dataframe.py#L103

tests/unit/core/statistics/test_constraints.py
```diff
@@ -401,3 +408,49 @@ def test_max_between_constraint_invalid():
...
+    cvisc = columnValuesInSetConstraint(value_set={2, 5, 8})
+    ltc = ValueConstraint(Op.LT, 1)
+    dc = DatasetConstraints(None, value_constraints={"annual_inc": [cvisc, ltc]})
+    config = load_config(local_config_path)
+    session = session_from_config(config)
+    profile = session.log_dataframe(df_lending_club, "test.data", constraints=dc)
+    session.close()
+    report = dc.report()
+
+    assert report[0][1][0][1] == len(df_lending_club)
```
jamie256, line 427: Not your fault, but our API for the report object is not easy to read, can you add a comment on why the length of the lending club data is what you are checking here?

proto/src/constraints.proto
```diff
@@ -59,6 +61,7 @@ message ValueConstraintMsg {
   oneof second_field {
     double value = 2;
     string regex_pattern = 5;
+    google.protobuf.ListValue value_set = 6;
```
jamie256, line 64: Can you explain a little why you chose ListValue as the type here? What were the other options considered? I see that you can support mixed sets of strings and numeric types in your tests. What does the protobuf serialization look like?

jamie256 (commented):
Looks good, (pending Richard's comments are addressed), and the tests are nice! I left a few questions and comments.

proto/src/constraints.proto
MilenaTrajanoska: I chose ListValue as the type to be able to support mixed value sets, because it's a wrapper over a repeated field of arbitraty values. Since this value is stored as a set in the code, there won't be any duplicate values in the protobuf serialization. Other options were to use repeated string / int / float, etc., but these would only support sets of values with only one type. I didn't find an implementation for serializing a set instead of a list.

The protobuf serialization looks like this:

name: "value IN {1, \'abc\', 2}"
op: IN
value_set {
  values {
    list_value {
      values {
        number_value: 1.0
      }
      values {
        string_value: "abc"
      }
      values {
        number_value: 2.0
      }
    }
  }
}

tests/unit/core/statistics/test_constraints.py
MilenaTrajanoska: Sure I'll add it in the tests. I check for the length of the dataframe to validate if all the rows were reported.

jamie256 (approved):

PR #378, opened 2021-12-15
Value Constraint: Contains specific pattern constraint

MilenaTrajanoska:
## Description

containsEmailConstraint
containsCreditCardConstraint
containsSSNConstraint
containsURLConstraint
stringLengthEqualConstraint
stringLengthBetweenConstraint

Fix numeric value constraint name() failing when value equals zero
Fix failing merge of two regex_pattern constraints

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/proto/src/constraints.proto
+++ b/proto/src/constraints.proto
@@ -32,10 +32,10 @@ message SummaryConstraintMsg {
   oneof second {
      string second_field = 3;
      double value = 4;
-     SummaryBetweenConstraintMsg between = 5;
+     SummaryBetweenConstraintMsg between = 7;
   }
-  Op op = 6;
-  bool verbose = 7;
+  Op op = 5;
+  bool verbose = 6;
 }
 
 message SummaryBetweenConstraintMsg {
--- a/src/whylogs/core/columnprofile.py
+++ b/src/whylogs/core/columnprofile.py
@@ -113,7 +113,7 @@ class ColumnProfile:
             self.string_tracker.update(value, character_list=character_list, token_method=token_method)
         # TODO: Implement real typed data conversion
 
-        self.constraints.update_typed(value)
+        self.constraints.update(value)
 
         typed_data = TypedDataConverter.convert(value)
 
@@ -133,7 +133,7 @@ class ColumnProfile:
 
         self.number_tracker.track(typed_data)
 
-        self.constraints.update(typed_data)
+        self.constraints.update_typed(typed_data)
 
     def _unique_count_summary(self) -> UniqueCountSummary:
         cardinality_summary = self.cardinality_tracker.to_summary(_UNIQUE_COUNT_BOUNDS_STD)
--- a/src/whylogs/core/columnprofile.py
+++ b/src/whylogs/core/columnprofile.py
@@ -112,6 +112,9 @@ class ColumnProfile:
         if isinstance(value, str):
             self.string_tracker.update(value, character_list=character_list, token_method=token_method)
         # TODO: Implement real typed data conversion
+
+        self.constraints.update(value)
+
         typed_data = TypedDataConverter.convert(value)
 
         if not TypedDataConverter._are_nulls(typed_data):
@@ -129,8 +132,6 @@ class ColumnProfile:
             return
 
         self.number_tracker.track(typed_data)
-
-        self.constraints.update(typed_data)
 
     def _unique_count_summary(self) -> UniqueCountSummary:
         cardinality_summary = self.cardinality_tracker.to_summary(_UNIQUE_COUNT_BOUNDS_STD)
--- a/src/whylogs/core/columnprofile.py
+++ b/src/whylogs/core/columnprofile.py
@@ -113,7 +113,7 @@ class ColumnProfile:
             self.string_tracker.update(value, character_list=character_list, token_method=token_method)
         # TODO: Implement real typed data conversion
 
-        self.constraints.update(value)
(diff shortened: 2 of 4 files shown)


src/whylogs/core/statistics/constraints.py
```diff
@@ -120,17 +120,32 @@ def update(self, v) -> bool:
     def merge(self, other) -> "ValueConstraint":
         if not other:
             return self
+        val = None
+        pattern = None
         assert self.name == other.name, f"Cannot merge constraints with different names: ({self.name}) and ({other.name})"
         assert self.op == other.op, f"Cannot merge constraints with different ops: {self.op} and {other.op}"
-        assert self.value == other.value, f"Cannot merge value constraints with different values: {self.value} and {other.value}"
-        merged_value_constraint = ValueConstraint(op=self.op, value=self.value, name=self.name, verbose=self._verbose)
+        if getattr(self, "value", None) is not None and getattr(other, "value", None) is not None:
```
richard-rogers: `if all([getattr(...]):`

src/whylogs/core/statistics/constraints.py
```diff
@@ -120,17 +120,32 @@ def update(self, v) -> bool:
...
+            val = self.value
+            assert self.value == other.value, f"Cannot merge value constraints with different values: {self.value} and {other.value}"
+        elif getattr(self, "regex_pattern", None) and getattr(other, "regex_pattern", None):
```
richard-rogers: `elif all([getattr...]):`

richard-rogers (changes requested):

proto/src/constraints.proto
```diff
@@ -32,10 +32,10 @@ message SummaryConstraintMsg {
   oneof second {
      string second_field = 3;
      double value = 4;
-     SummaryBetweenConstraintMsg between = 5;
```
jamie256: Can we avoid changing the codes on the proto files, or explain why we need this?

src/whylogs/core/columnprofile.py
```diff
@@ -112,6 +112,9 @@ def track(self, value, character_list=None, token_method=None):
         if isinstance(value, str):
             self.string_tracker.update(value, character_list=character_list, token_method=token_method)
         # TODO: Implement real typed data conversion
+
+        self.constraints.update(value)
```
jamie256, line 116: Does this change the behavior of the other constraints?

jamie256 (commented):
Looks good, couple questions.

proto/src/constraints.proto
pecop2: Same as the previous question on the other PR, we already merged a "wrong" PR (mainline contains already changed proto files) so this would be the way it should have been done in the first place, after this we only append things and don't change the previous ones if that makes sense.

jamie256: Thanks for explanation, if you add a comment to any of those PRs where you are reverting that we can skip us commenting on that :-)

src/whylogs/core/columnprofile.py
jamie256: For example, we have schema tracker count the types, so wanted to see if we change the "inferred data type" for example.

pecop2: This seems like it does, I think @MilenaTrajanoska changed this for the problem with the yaml.safe_load function. She should confirm though.

MilenaTrajanoska: It would not change the inferred data type metric, but there might be an inconsistency. For example if we have '1.2' as a string it would check it in a string matching pattern, but later on it will be converted to float 1.2 which would update the number trackers instead of the string trackers. We could instead check the type of the value and if it's a string, see if any of the constraints are MATCH or NON_MATCH, then skip the conversion and pass it as a string to the trackers, and if the column does not have string based constraints, we would do type conversion  first and pass it to the constraints and trackers as the inferred type.

jamie256: Ok, thanks for clarifying. I'm trying to make sure we preserve the existing behavior in the trackers. We can change things for the constraints, but changing behavior in the trackers would be a disruptive change. If it is easier we can have two methods rather then the one "update" on `self.constraints` and introduce an update_typed() so that we can update some constraints with the pre-conversion raw string (and the string matching/regex constraints would go there), and then the other updates can occur in the normal location post type conversion.

jamie256:
This looks good, update the branch and resolve conflicts and we could merge, remaining comments are optional and can be addressed post merge if you like.

src/whylogs/core/statistics/constraints.py
```diff
@@ -437,7 +449,13 @@ def to_protobuf(self) -> ValueConstraintMsgs:
 
     def update(self, v):
         for c in self.constraints.values():
-            c.update(v)
+            if c.op not in (Op.MATCH, Op.NOMATCH):
```
jamie256: We will probably want to keep two separate lists of the contraints rather than check for the regex constraint as the switch for these methods, so there are raw value constraints and coerced typed constraints that can be added and we avoid the special logic for regex constraint in this code path and its bound at construction time instead?

MilenaTrajanoska: Yes, I wasn't sure if this was what you meant by having two types of updates. I'm going to update the PR with this change.

jamie256 (approved):

jamie256 (approved):
LGTM!

richard-rogers (approved):

PR #379, opened 2021-12-15
SummaryConstraint: Assert distinct values in set constraint

pecop2:
## Description

Assert distinct values in set constraint. New Ops in proto for SET: IN, CONTAIN, EQ. New lambda functions for the operations, using string_tracker and number_tracker theta sketches for distinct column values set operations

distinctValuesInSetConstraint
distinctValuesEqualSetConstraint
distinctValuesContainSetConstraint

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/proto/src/constraints.proto
+++ b/proto/src/constraints.proto
@@ -20,7 +20,7 @@ enum Op {
     NOMATCH = 8;
     BTWN = 9;
     IN_SET = 10;
-    CONTAINS_SET = 11;
+    CONTAIN_SET = 11;
     EQ_SET = 12;
   }
 
--- a/proto/src/constraints.proto
+++ b/proto/src/constraints.proto
@@ -1,6 +1,7 @@
 syntax = "proto3";
 
 import "messages.proto";
+import "google/protobuf/struct.proto";
 
 option java_package = "com.whylogs.core.constraint";
 option java_outer_classname = "Constraints";
@@ -18,6 +19,9 @@ enum Op {
     MATCH = 7;
     NOMATCH = 8;
     BTWN = 9;
+    IN_SET = 10;
+    CONTAINS_SET = 11;
+    EQ_SET = 12;
   }
 
 /* Summary constraints specify a relationship between a summary field and a literal value,
@@ -32,10 +36,11 @@ message SummaryConstraintMsg {
   oneof second {
      string second_field = 3;
      double value = 4;
-     SummaryBetweenConstraintMsg between = 5;
+     SummaryBetweenConstraintMsg between = 7;
+     google.protobuf.ListValue reference_set = 8;
   }
-  Op op = 6;
-  bool verbose = 7;
+  Op op = 5;
+  bool verbose = 6;
 }
 
 message SummaryBetweenConstraintMsg {
--- a/src/whylogs/core/datasetprofile.py
+++ b/src/whylogs/core/datasetprofile.py
@@ -687,7 +687,12 @@ class DatasetProfile:
             if feature_name in self.columns:
                 colprof = self.columns[feature_name]
                 summ = colprof.to_summary()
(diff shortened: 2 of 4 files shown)


src/whylogs/core/statistics/constraints.py
```diff
@@ -213,7 +232,27 @@ def __init__(
         self.value = value
         self.upper_value = upper_value
 
-        if self.op == Op.BTWN:
+        if self.op in (Op.IN_SET, Op.CONTAIN_SET, Op.EQ_SET):
+            if value is not None or upper_value is not None or second_field is not None or third_field is not None or reference_set is None:
```
richard-rogers: `if any([value, upper_value, second_field, third_field, reference_set]):`

src/whylogs/core/statistics/constraints.py
```diff
@@ -248,19 +287,58 @@ def __init__(
...
-        if not self.func(summ):
-            self.failures += 1
-            if self._verbose:
-                logger.info(f"summary constraint {self.name} failed")
+        summ = update_dict["number_summary"]
+        column_string_theta = update_dict["string_theta"]
+        column_number_theta = update_dict["number_theta"]
+
+        if self.op in (Op.IN_SET, Op.CONTAIN_SET, Op.EQ_SET):
+            if not _summary_funcs1[self.op](self.string_theta_sketch)(column_string_theta) or not _summary_funcs1[self.op](self.numbers_theta_sketch)(
```
richard-rogers: `if not all([_summary_funcs1.... ]):`

richard-rogers (changes requested):

src/whylogs/core/statistics/constraints.py
pecop2: Thanks, so much better, but shouldn't it be: 
`if any([value, upper_value, second_field, third_field, not reference_set])` ?

proto/src/constraints.proto
```diff
@@ -32,10 +36,11 @@ message SummaryConstraintMsg {
   oneof second {
      string second_field = 3;
      double value = 4;
-     SummaryBetweenConstraintMsg between = 5;
```
jamie256: Can you explain the code changes here?

jamie256 (commented):
This looks great, just a proto change question

proto/src/constraints.proto
pecop2: Of course: the ones with the op, verbose and between are clear by now, as the answer to the question at the PR concerning the same should be sufficient.
google.protobug.ListValue reference_set = 8; --> this one is used for the set provided by the user to be used with the IN, CONTAIN, and EQ set operations. As it can contain more than one data type, this was the possible solution.
We coordinated with @MilenaTrajanoska to use the same field too: https://github.com/whylabs/whylogs/pull/375/files/49c9e48adbe91126aba107bd595a4c47b0d1a315#diff-3ead279fdd02046f469418a748586aa737b3a3cf99be981b99db1bee165a2a22R64

src/whylogs/core/datasetprofile.py
```diff
@@ -687,7 +687,14 @@ def apply_summary_constraints(self, summary_constraints: Optional[Mapping[str, S
...
                 colprof = self.columns[feature_name]
                 summ = colprof.to_summary()
-                constraints.update(summ.number_summary)
+                update_dict = {
+                    "number_summary": summ.number_summary,
+                    "string_theta": colprof.string_tracker.theta_sketch.theta_sketch,
+                    "number_theta": colprof.number_tracker.theta_sketch.theta_sketch,
+                }
+
+                # constraints.update(summ.number_summary)
```
jamie256: remove commented out code

richard-rogers (approved):

PR #380, opened 2021-12-16
Autobump Java package version

richard-rogers:
### General Checklist

* [ ] Documentation updated

### DIFF
--- a/.github/workflows/push_release.yml
+++ b/.github/workflows/push_release.yml
@@ -35,7 +35,7 @@ jobs:
     - name: Install dependencies
       run: make install
 
-    - name: Autobump version
+    - name: Autobump Python version
       run: |
         # from refs/tags/v1.2.3 get 1.2.3
         VERSION=$(echo $GITHUB_REF | sed 's#.*/v##')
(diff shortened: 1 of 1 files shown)


.github/workflows/push_release.yml
```diff
@@ -44,6 +44,15 @@ jobs:
         git config --local user.name "github-actions[bot]"
         poetry run bump2version patch --verbose --commit --no-tag --new-version $VERSION
 
+    - name: Autobump Java version
+      run: |
+        VERSION=$(echo $GITHUB_REF | sed 's#.*/v##')
+        echo $VERSION
+        git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+        git config --local user.name "github-actions[bot]"
+        poetry run bump2version patch --verbose --commit --no-tag --new-version $VERSION
```
jamie256, line 53: Can you describe what the previous version was and how the bump2version updates that for the java side?

jamie256 (commented):
Left a question about the format of the version numbers we get from this.

.github/workflows/push_release.yml
richard-rogers: It's currently at 0.1.5-b0. I think this will make it match the mainline release tag (same as Python package). Alternatively, I could grep the current Java version out of `java/.bumpversion.cfg` and bump that...

jamie256: ok, so going from 0.1.5-b0 -> 0.6.17? We can do that, and sync the versions.

jamie256 (approved):
Synchronizing the package versions!

jamie256:
After some discussion on this, lets keep the java version separate so we don't bump that version when there are no changes there since we are updating the python side more frequently at present.

.github/workflows/push_release.yml
jamie256: Better to have the ability to manage them separately if that is reasonable

richard-rogers:
Decided to leave Java version unbumped

PR #381, opened 2021-12-16
Summary Constraint: quantileBetweenConstraint

MilenaTrajanoska:
## Description

Implemented quantileBetweenConstraint
Added quantile_value to protobuf

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/src/whylogs/core/datasetprofile.py
+++ b/src/whylogs/core/datasetprofile.py
@@ -25,6 +25,7 @@ from whylogs.proto import (
     DatasetSummary,
     MessageSegment,
     ModelType,
+    NumberSummary,
 )
 from whylogs.util import time
 from whylogs.util.time import from_utc_ms, to_utc_ms
@@ -687,10 +688,11 @@ class DatasetProfile:
             if feature_name in self.columns:
                 colprof = self.columns[feature_name]
                 summ = colprof.to_summary()
-                update_dict = {
-                    "number_summary": summ.number_summary,
-                    "number_kll_sketch": colprof.number_tracker.histogram,
-                }
+
+                update_dict = _create_update_summary_dictionary(
+                    number_summary=summ.number_summary,
+                    quantile=colprof.number_tracker.histogram,
+                )
 
                 constraints.update(update_dict)
             else:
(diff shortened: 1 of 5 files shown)


src/whylogs/core/statistics/constraints.py
```diff
@@ -251,13 +259,26 @@ def name(self):
...
-    def update(self, summ: NumberSummary) -> bool:
+    def update(self, update_dict: dict) -> bool:
         self.total += 1
-        if not self.func(summ):
+        summ = update_dict["number_summary"]
+        column_number_kll_sketch = update_dict["number_kll_sketch"]
+
+        if self.first_field == "quantile":
+            quantile_summary = quantiles_from_sketch(column_number_kll_sketch, [self.quantile_value])
+            quantile_val = type("Object", (), {self.first_field: quantile_summary.quantile_values[0]})
```
jamie256: Can you explain this some? Do we need a wrapper or utility function for the quantile values?

MilenaTrajanoska: The user specifies the quantile value for which they want to apply the constraint. This value is passed in the constructor of the SummaryConstraint in the quantile_value argument. The quantiles_from_sketch method calculates the quantiles and their value based on the number summary kll sketch, and the passed quantiles (list) as a second parameter. The first_first field here is only used to check if this is a quantile constraint and then calculate the quantile value (one quantile), which is then passed as an anonymous object with an attribute quantile = the_quantile_value. This is because the implementation of the summary operators requires an object from which it gets the corresponding field (field_name) using the operator. 

We could use a utility function, it would make the interface clearer.

jamie256: yes a utility function here would be better to encapsulate that interface expectation/wrapping. And/or add a comment hinting at the reason we need to do this in the code so that reading the code is easier to understand.

jamie256:
Can you resolve the conflicts and update?

src/whylogs/core/datasetprofile.py
```diff
@@ -796,3 +799,34 @@ def array_profile(
...
+    update_dict.update(
+        {
+            field_name: getattr(number_summary, field_name)
+            for field_name in dir(number_summary)
+            if str.islower(field_name) and not str.startswith(field_name, "_") and not callable(getattr(number_summary, field_name))
+        }
+    )
+    update_dict.update(kwargs)
+
+    return type("Object", (), update_dict)
```
jamie256, line 832: This is better factored into a utility function. I think we should revisit this and consider replacing with a facade class, we can also update the consuming code to not require this shape to make it easier to compose the column profile and summary fields we need access to.

jamie256 (approved):
Looks good.

PR #387, opened 2021-12-20
ValueConstraint: Datetime/json constraints

pecop2:
## Description

Assert strftime format
Assert dateutil parseable
Assert JSON parseable
Match JSON schema

New op: APPLY_FUNC, evaluation using a function passed when creating the constraint
dateUtilParseableConstraint()
jsonParseableConstraint()
matchesJsonSchemaConstraint()
strftimeFormatConstraint()

Some of the concerns in one of the previous PRs would still be valid for this one, e.g. if statements, but after merging the previous PRs I would update this one from mainline accordingly.
Previous unmerged constraints included here, so the review can wait after merging those.

### General Checklist

* [ ] Documentation updated


src/whylogs/core/statistics/constraints.py
```diff
@@ -20,6 +27,77 @@
 
 logger = logging.getLogger(__name__)
 
+
+def _is_strftime_format(strftime_val, format):
```
jamie256: for these functions, your implementation is doing the expensive thing of actually parsing the object so let's make these utility functions a little more reusable (even if internal):
1) rename the prefix from "_is_" to "_try_parse_"
2) return the parsed object
3) add type hints for the input param and output (minor issue but helps readability)

Consider using optional return or tuple of (bool, concrete instance) e.g. Optional[datetime] or (datetime, bool)

src/whylogs/core/statistics/constraints.py
```diff
@@ -82,17 +181,33 @@ class ValueConstraint:
...
+            if value is not None:
+                if not isinstance(value, str):
+                    if apply_function == globals()["_matches_json_schema"]:
+                        try:
+                            value = json.dumps(value)
+                        except (ValueError, TypeError):
+                            raise ValueError("Json schema ivalid. When matching json schema, the schema provided must be valid")
+                    else:
+                        value = str(value)
+                self.value = value
```
jamie256: Can you factor this into a function? It will make it easier to read and test.

jamie256 (commented):
I realize this PR is on top of other un-merged changes, but I left a few initial comments. Maybe you can address those and rebase after the earlier changes are in?

jamie256 (approved):
Looks good, just need to resolve conflicts

pecop2:
I resolved conflicts and edited the update process of the BTWN set constraints, using the update object introduced by @MilenaTrajanoska to follow the same pattern.

PR #396, opened 2021-12-22
Remove git submodules in Java workflows

richard-rogers:
### General Checklist

* [ ] Documentation updated

### DIFF
--- a/.github/workflows/java-continuous-integration.yml
+++ b/.github/workflows/java-continuous-integration.yml
@@ -17,9 +17,7 @@ jobs:
         working-directory: java
     steps:
       - uses: actions/checkout@v2
-        with:
-          submodules: true
-       - name: Change to Java dir
+      - name: Change to Java dir
         run: |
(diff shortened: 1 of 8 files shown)


jamie256:
Let's update this to use the local proto definitions: https://github.com/whylabs/whylogs/tree/mainline/proto/src

Python already builds the proto definitions so no need to update that side of our build step or workflow.

The submodule approach is part of the previous factorization when the java package was separate from the python whylogs. Let's restructure this workflow to not depend on submodules. We can move the proto folder, but lets centralize the local definitions as the authority for message definitions in whylogs.

richard-rogers:
> Let's update this to use the local proto definitions

Looking at the sub-level `build.gradle.kts`, I think it's already setup to use the local proto stuff. I think it just depends on the Python build happening first to build proto.

I also bumped log4j in this PR to deal with the latest vulnerabilities.

jamie256 (approved):
LGTM

PR #397, opened 2021-12-27
SummaryConstraints: Distributional Measures Constraints

MilenaTrajanoska:
## Description

Implemented  parametrizedKSTestPValueGreaterThanConstraint
Implemented columnKLDivergenceLessThanConstraint

KllFloatsSketchMessage in messages.proto

ReferenceDistributionContinuousMessage and ReferenceDistributionDiscreteMessage in constraints.proto

Summary converters for KS test and KL Divergence calculation

### General Checklist

* [ ] Documentation updated


jamie256 (changes requested):
The CI checks are failing because we breached complexity threshold in the __init__ method, this has been building so its not this change alone that created the complexity but we could factor some of the the 'field' and 'op' handling into functions which would also make this easier to read and test. 

After getting the CI to pass also resolve conflicts with mainline.

jamie256 (approved):
LGTM

PR #398, opened 2021-12-27
SummaryConstraint: Table shape constraints

pecop2:
## Description

columnExistsConstraint
numberOfRowsConstraint
columnsMatchSetConstraint

Evaluated using apply_table_shape_constraints
Attribute table_shape_constraints in DatasetConstraints
Attribute total_row_number in DatasetProfile
Tests for DatasetConstraints serialization/deserialization as changes in proto DatasetConstraintMsg were required
Tests for table shape constraints

### General Checklist

* [ ] Documentation updated


src/whylogs/core/statistics/constraints.py
```diff
@@ -54,6 +160,7 @@
     Op.NE: lambda f, f2: lambda s: getattr(s, f) != getattr(s, f2),
     Op.GE: lambda f, f2: lambda s: getattr(s, f) >= getattr(s, f2),
     Op.GT: lambda f, f2: lambda s: getattr(s, f) > getattr(s, f2),
+    Op.BTWN: lambda f, f2, f3: lambda s: getattr(s, f2) <= getattr(s, f) <= getattr(s, f3),
```
Taxuspt, line 183: We should double check this condition. If the values are all the same or if the value is equal to one of the edge values, the condition will still be true.

src/whylogs/core/statistics/constraints.py
```diff
@@ -77,13 +191,33 @@ class ValueConstraint:
...
+            raise ValueError("Value constraint must provide a set of values for using the IN operator")
+
+        if self.op == Op.APPLY_FUNC:
+            if value is not None:
+                if not isinstance(value, str):
+                    if apply_function == globals()["_matches_json_schema"]:
+                        try:
+                            value = json.dumps(value)
+                        except (ValueError, TypeError):
+                            raise ValueError("Json schema ivalid. When matching json schema, the schema provided must be valid")
```
Taxuspt: typo in "ivalid"

src/whylogs/core/statistics/constraints.py
```diff
@@ -193,77 +396,286 @@ def __init__(
...
 
         merged_constraint.total = self.total + other.total
         merged_constraint.failures = self.failures + other.failures
         return merged_constraint
 
     @staticmethod
     def from_protobuf(msg: SummaryConstraintMsg) -> "SummaryConstraint":
-        if msg.HasField("value") and not msg.HasField("second_field"):
+
+        if msg.HasField("reference_set") and not msg.HasField("value") and not msg.HasField("second_field") and not msg.HasField("between"):
```
Taxuspt: This check could be turned into a function, since we do it several times in the next lines. This way we would avoid the confusing `and not`,..., `and not`,...

src/whylogs/core/statistics/constraints.py
```diff
@@ -193,77 +396,286 @@ def __init__(
...
+            return SummaryConstraint(
+                msg.first_field,
+                msg.op,
+                reference_set=set(msg.reference_set),
+                name=msg.name,
+                verbose=msg.verbose,
+            )
+        elif (
+            (msg.HasField("value") or msg.HasField("value_str"))
```
Taxuspt: Same comment has above. Turning the if condition into a function

src/whylogs/core/statistics/constraints.py
```diff
@@ -437,4 +863,110 @@ def to_json(self) -> str:
...
+def columnExistsConstraint(column: str, verbose=False):
+    return SummaryConstraint("columns", Op.CONTAIN, value=column, verbose=verbose)
+
+
+def numberOfRowsConstraint(n_rows: int, verbose=False):
+    return SummaryConstraint("total_row_number", Op.EQ, value=n_rows, verbose=verbose)
+
+
+def columnsMatchSetConstraint(reference_set: Set[str], verbose=False):
+    return SummaryConstraint("columns", Op.EQ, reference_set=reference_set, verbose=verbose)
```
Taxuspt, line 1197: Not sure we are adopting PEP8, but we should have an empty line and the end of the file.

Taxuspt (changes requested):
Looks good.
I left a few comments, mostly about styles.
I don't know if this was discussed with Jamie, but should the code be PEP8 compliant?

src/whylogs/core/statistics/constraints.py
pecop2: We are using <= in the between operations everywhere. If that's wrong we can change it.

Taxuspt: We can merge it like this and then confirm with Jaimie that this is the expected behaviour.

pecop2: This part with the BTWN is already merged in mainline with the previous PRs, I would say Jamie approved it.

src/whylogs/core/statistics/constraints.py
pecop2: I am certain that in my editor there is always a new line at eof, as I run "make pre-commit" that throws a warning about it.

src/whylogs/core/statistics/constraints.py
```diff
@@ -152,6 +146,11 @@ def _matches_json_schema(json_data: Union[str, dict], json_schema: Union[str, di
     == round(theta_a_not_b().compute(ref_str_sketch, getattr(update_obj, f)["string_theta"]).get_estimate(), 1)
     == round(theta_a_not_b().compute(ref_num_sketch, getattr(update_obj, f)["number_theta"]).get_estimate(), 1)
     == 0.0,
+    Op.IN: lambda f, v: lambda s: getattr(s, f) in v,
+    Op.CONTAIN: lambda f, v: lambda s: v in getattr(s, f),
+    # Op.QUANTILE_BETWEEN: lambda first_field, lower, upper, quantile_value: lambda update_object: lower <=
```
jamie256: remove commented out code?

jamie256 (commented):
When ready, point to mainline and I can take another look.

jamie256 (changes requested):
Please resolve conflicts (hopefully that also triggers CI to run and report test status, looks like the last test run never reported status).

jamie256 (approved):

src/whylogs/core/datasetprofile.py
```diff
@@ -282,6 +282,7 @@ def track_dataframe(self, df: pd.DataFrame, character_list=None, token_method=No
         # workaround for CUDF due to https://github.com/rapidsai/cudf/issues/6743
         if cudfDataFrame is not None and isinstance(df, cudfDataFrame):
             df = df.to_pandas()
+        self.total_row_number = len(df.index)
```
jamie256: If I log two different pandas dataframes for the same dataset (e.g. loading csv files sequentially into pandas dataframes) will this overwrite the row number rather than update it? dataset profiles should be updateable and mergeable. I think you want to add to the total_row_number instead of assign it and handle the merge by adding the values?

Also there is the track() method which doesn't always go through the track_dataframe method, so potentially we're missing some rows / columns potentially if a user logs data outside of the log_dataframe method, e.g. will this cover our readme example? https://github.com/whylabs/whylogs#quickly-logging-data

jamie256 (changes requested):
General implementation looks good. I think there are some edge cases not handled I left a comment on.

tests/unit/core/statistics/test_constraints.py
```diff
@@ -1264,6 +1267,314 @@ def test_unique_proportion_between_constraint_wrong_datatype():
...
+    if len(set2) > 20:
+        tmp_set = set(list(set2)[:20])
+        reference_set_str = f"{str(tmp_set)[:-1]}, ...}}"
+    else:
+        reference_set_str = str(set2)
+    assert report[5][0] == f"table columns {Op.Name(Op.EQ)} {reference_set_str}"
+    assert report[5][1] == 1
+    assert report[5][2] == 0
+
+
```
richard-rogers, line 1397: parametrize?

src/whylogs/core/datasetprofile.py
pecop2: The way it was implemented, when logging sequentially it would have been a problem since the total_row_number gets overwritten.
And after taking a closer look, there are functions that use track() and/or track_datum() methods directly which would be a problem as well. 
My bad I didn't realize those things earlier.
What I changed, as every method used for tracking (track_dataframe(), track(), track_array() etc.) in the DatasetProfile, ends up using the track_datum() method, that's where I update the variables so everything that gets logged is being accounted for.

jamie256 (approved):

tests/unit/core/statistics/test_constraints.py
jamie256: Yeah it is a good suggestion. Ok to do this in a follow up change.

PR #410, opened 2022-01-08
Java release CI

richard-rogers:
## Description

Adds Java release publishing

### General Checklist

* [ ] Documentation updated


java/spark-bundle/build.gradle.kts
```diff
@@ -18,7 +19,7 @@ val sparkVersion = project.properties.getOrDefault("sparkVersion", "3.1.1") as S
 val artifactBaseName = "whylogs-spark_${sparkVersion}-scala_$scalaVersion"
```
jamie256: I think we also want to update the artifact name here to be consistent with previously published bundled jar

`val artifactBaseName = "whylogs-spark-bundle_${sparkVersion}-scala_$scalaVersion"`

jamie256: We can do in a follow up since its not in your change, but up to you.

jamie256 (approved):
LGTM!

PR #417, opened 2022-01-17
Constraints example notebook

pecop2:
## Description

Example jupyter notebook for the new constraints.

### General Checklist

* [ ] Documentation updated


jamie256 (changes requested):
Cool, lots of examples here!

When ready can you update this PR on mainline?

* Can we rename the example file to something more descriptive than suffix of "2". I'm open to suggestions, but you can use something like "More_Constraints" if its hard to think of a good name.
* Can we include a few examples of constraints generation if there are updates there?

jamie256 (changes requested):
Looks like there is a method import error in the constraints notebook test, can you take a look?

 ```
ImportError: cannot import name 'numberOfRowsConstraint' from 'whylogs.core.statistics.constraints' 
```

pecop2:
> Looks like there is a method import error in the constraints notebook test, can you take a look?
> 
> ```
> ImportError: cannot import name 'numberOfRowsConstraint' from 'whylogs.core.statistics.constraints' 
> ```

With 'numberOfRowsConstraint' being a part of the Table shape constraints PR that's not merged yet, that could be the problem.

jamie256 (approved):

PR #419, opened 2022-01-19
Dev/richard/unicode

richard-rogers:
### General Checklist

* [ ] Documentation updated


src/whylogs/core/statistics/stringtracker.py
```diff
@@ -158,6 +168,128 @@ def to_summary(self):
...
+    def __init__(self, block_list: Dict[str, List[int]] = None):
+
+        if block_list is None:
+            # might want to make this a multimap or make each value a list of ranges
+            # maybe make the defualt {} so nothing's tracked to match old StringTracker behaviour
+            block_list = {
+                "emoji": [0x1F600, 0x1F64F],
+                "basic-latin": [0x0000, 0x007F],
+                "extended-latin": [0x0080, 0x02AF],
+            }
```
richard-rogers: Include a bucket for characters that aren't in any listed block.

Makefile
```diff
@@ -102,7 +102,7 @@ requirements.txt:
 
 $(build.proto): $(src.proto)
 	@$(call i, Generating python source for protobuf)
-	protoc -I $(src.proto.dir) --python_out=$(build.proto.dir) $(src.proto)
+	protoc -I $(src.proto.dir) --python_out=$(build.proto.dir) --experimental_allow_proto3_optional $(src.proto)
```
richard-rogers: Add this to the Java build, or drop it if unneeded.

src/whylogs/core/statistics/stringtracker.py
```diff
@@ -251,8 +386,9 @@ def merge(self, other):
         new_length = self.length.merge(other.length)
         new_token_length = self.token_length.merge(other.token_length)
         new_char_pos_tracker = self.char_pos_tracker.merge(other.char_pos_tracker)
+        new_char_block_tracker = self.char_block_tracker.merge(other.char_block_tracker)  # block map?
 
-        return StringTracker(count, items_copy, new_theta, new_length, new_token_length, new_char_pos_tracker)
+        return StringTracker(count, items_copy, new_theta, new_length, new_token_length, new_char_pos_tracker, self.token_method, new_char_block_tracker)
```
richard-rogers: Validate that both `StringTracker`s have the same `token_method`?

Makefile
jamie256: I don't think we need the optional tag on the protobuf fields, they can be empty in general. The 'optional' may be useful when consuming code requires a default value, but in general fields can be empty and we handle that. https://developers.google.com/protocol-buffers/docs/pythontutorial?hl=en#defining-your-protocol-format

Or can you explain why you wanted these marked optional.

PR #427, opened 2022-01-27
Change total row number computation

pecop2:
## Description

Changed the way of computing the total_row_number used for the Table shape constraints, using each column profile counters tracker instead of creating and updating a separate dictionary. This as well addresses the concern of merging two dataset profiles and possibility of inconsistencies when using a separate dictionary for tracking the total row number.

### General Checklist

* [ ] Documentation updated


jamie256 (approved):
Thanks for the follow up here, looks good!

PR #431, opened 2022-02-03
Display human readable name for constraints

MilenaTrajanoska:
## Description

Changed default constraints naming to a more human readable format

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/src/whylogs/core/statistics/constraints.py
+++ b/src/whylogs/core/statistics/constraints.py
@@ -1203,6 +1203,15 @@ class MultiColumnValueConstraints(ValueConstraints):
             return MultiColumnValueConstraints({v.name: v for v in value_constraints})
         return None
 
+    def to_protobuf(self) -> ValueConstraintMsgs:
+        v = [c.to_protobuf() for c in self.raw_value_constraints.values()]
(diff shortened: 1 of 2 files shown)


src/whylogs/core/statistics/constraints.py
```diff
@@ -258,13 +260,16 @@ def __init__(self, op: Op, value=None, regex_pattern: str = None, apply_function
 
     @property
     def name(self):
```
MilenaTrajanoska, line 262: The name is serialized to protobuf even if the user does not supply a name parameter. This is needed because if we have a value constraint for the values to be greater than 10, then the default name for ValueConstraint(op.GT, 10) when no name is supplied would be "value GT 10". After we call to_protobuf on the same constraint, the internal implementation casts 10 to 10.0. If the name attribute is not serialized, then when we create the constraint using from_protobuf again, the new name would be "value GT 10.0", which would make these constraints unmergable since their names would now be different.

jamie256: Thanks for the context, makes sense.

jamie256 (approved):
LGTM!

PR #433, opened 2022-02-08
Dev/loka/jupyter notebook visualization

Jirayr-Solvee:
## Description

This PR gives ability to get list of visualizations for profiled data inside jupyter notebook

### General Checklist

* [ ] Documentation updated


src/whylogs/viewer/index-hbs-cdn-all-in-jupyter-full-summary-statistics.html
```diff
@@ -0,0 +1,328 @@
...
+           <div class="statistic-number">${newNumber}
+             <div class="statistic-measurement">${suffixe}</div>
+           </div>
+           <span class="tooltiptext">${number}</span>
+         </div>`
+
+      const randomNumbers = (range) => Math.floor(Math.random() * range)
+
+      Handlebars.registerHelper("observations", function () {
+        const {value, newValue, suffixe} = abbreviateNumber(randomNumbers(900000000))
```
jamie256: These are still random numbers as placeholders? Let me know when the reports are using the profile data.

jamie256 (changes requested):
Can we make sure we have axis labels for the double histogram chart (same for any of the new charts).

For the drift reports, can we include thresholds and method used to detect drift (maybe its something in a legend or hover).

src/whylogs/viewer/index-hbs-cdn-all-in-jupyter-full-summary-statistics.html
jamie256: fixed

jamie256 (changes requested):
Thanks for making those fixes, looking better! I like the difference charts :-)

1) The ranges labeled are helpful for the summary_drift_report categories! Can we also add a hover info or line in the legend that references the algorithm used for this test?

2) For the visualization.summary_drift_report(preferred_cell_height="1000px") cell there is a severe drift detected with a value as "undefined" for "uniform_integers" is that correct? If this is a divide by zero case I'd expect it to be same treatment as a max value and no drift detected, or is something else going on?

3) for the visualization.double_histogram(feature_names="uniform_integers") the lable values are useful, can we mention the units? for each axis or some hint on what this is?

jamie256 (commented):
Did we solve the case when comparing distributions that don't align?

FelipeAdachi:
- **double_histogram**: For all of the mixture features, it seems there's a long tail on the X-axis, all the way to 3.0, when max values go up to 1.2~1.3. We can see an example in the attached image. This sould be reproducible directly from the code in `Profile_Viewer_In_Notebook`.

- **double_histogram**: Another issue on double_histogram is the shape of the bins: in some cases, for the mixture_distributions, the bins have a large width and seem to overlap. The attached image is also an example of that. This sould be reproducible directly from the code in `Profile_Viewer_In_Notebook`

image-20220301-203413

jamie256 (approved):

PR #437, opened 2022-02-09
DatasetConstraints serialization and chi square calculation fixes

pecop2:
## Description

Fixing a serializing bug for DatasetConstraints that contained MultiColumnConstraints and updating the tests.
Fix for Chi-Squared statistic calculation.

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/tests/unit/core/statistics/test_constraints.py
+++ b/tests/unit/core/statistics/test_constraints.py
@@ -2875,6 +2875,38 @@ def test_column_values_unique_within_row_constraint_invalid_params():
         columnValuesUniqueWithinRow(column_A=["A"])
 
 
+def test_multicolumn_value_constraints_report(local_config_path):
+    data = pd.DataFrame(
+        {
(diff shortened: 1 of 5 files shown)


tests/unit/core/statistics/test_constraints.py
```diff
@@ -1563,11 +1563,21 @@ def test_dataset_constraints_serialization():
     set1 = set(["col1", "col2"])
     columns_match_constraint = columnsMatchSetConstraint(set1)
 
+    val_set = {(1, 2), (3, 5)}
```
richard-rogers, line 1566: This function is far too long. Not necessarily a problem to fix in this PR, but we should at least file a ticket to refactor this.

pecop2: I've changed it now, you can take a look when you have the time.

jamie256 (approved):
Looks good, thanks for making the fixes and refactor.

PR #438, opened 2022-02-10
Add Java package smoketest

richard-rogers:
## Description

Adds smoke test to verify the built jar works.

Also updates lombok and graddle

### General Checklist

* [ ] Documentation updated


java/gradlew.bat
```diff
@@ -1,103 +1,103 @@
-@rem
```
jamie256, line 1: Is this just line endings changes?

java/smoketest/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+    dirs '../build/libs', '../cli/build/libs', '../core/build/libs'
+  }
+  mavenCentral()
+}
+
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
+
+dependencies {
+  implementation "ai.whylabs:core:0.1.7-b2"
```
jamie256, line 18: How do we handle whylogs java version changes?

java/smoketest/src/main/java/smoketest/SmokeTest.java
```diff
@@ -0,0 +1,8 @@
+import com.whylogs.core.DatasetProfile;
+import java.time.Instant;
+
+public class SmokeTest {
+    public static void main(String[] args) {
+       DatasetProfile dsp = new DatasetProfile("test", Instant.now());
```
jamie256, line 8: Will this catch the missing protobuf message definitions?

jamie256 (commented):
Left a few questions.

java/gradlew.bat
richard-rogers: I'm not sure if there are other differences between the 6.3 and 7.3 wrapper scripts. Since it's the Windows script, it makes sense to have the Windows line endings.

java/smoketest/build.gradle
richard-rogers: The workflow needs something like the Python's bumpversion. I'm still inclined to keep the Python and Java version numbers the same so the packages don't collide when updating the other language.

java/smoketest/src/main/java/smoketest/SmokeTest.java
richard-rogers: I can add code to explicitly exercise the protobuf code.

jamie256: This is a good start, thanks for clarifying.

java/smoketest/build.gradle
jamie256: ok, let's follow up with the version bump outside this PR.

jamie256 (approved):
LGTM!

PR #444, opened 2022-02-16
Fix KS test p-value computation

MilenaTrajanoska:
## Description

Fix KS test p-value computation
Add condition in entropy calculation

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/src/whylogs/core/summaryconverters.py
+++ b/src/whylogs/core/summaryconverters.py
@@ -289,7 +289,8 @@ def ks_test_compute_p_value(target_distribution: kll_floats_sketch, reference_di
             D_max = D
         j += 1
 
-    p_value = special.kolmogorov(np.sqrt(num_quantiles) * D_max)
+    n_samples = min(target_distribution.get_n(), reference_distribution.get_n())
+    p_value = special.kolmogorov(np.sqrt(n_samples) * D_max)
     return type("Object", (), {"ks_test": p_value})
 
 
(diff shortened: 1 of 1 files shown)


src/whylogs/core/summaryconverters.py
```diff
@@ -251,13 +253,42 @@ def ks_test_compute_p_value(target_distribution: kll_floats_sketch, reference_di
...
+
+    while j < num_quantiles:
+        cdf_target = target_distribution.get_cdf([ref_quantile_values[j]])[0]
+        cdf_ref = reference_distribution.get_cdf([ref_quantile_values[j]])[0]
+        D = abs(cdf_target - cdf_ref)
+        if D > D_max:
+            D_max = D
+        j += 1
+
     n_samples = min(target_distribution.get_n(), reference_distribution.get_n())
```
MilenaTrajanoska, line 292: Is this the right number of samples?

src/whylogs/core/summaryconverters.py
```diff
@@ -251,13 +253,42 @@ def ks_test_compute_p_value(target_distribution: kll_floats_sketch, reference_di
...
     p_value = special.kolmogorov(np.sqrt(n_samples) * D_max)
```
MilenaTrajanoska, line 293: Is this the right distribution for the p-value?

bernease:
Looking into this later today.

jamie256 (approved):
LGTM!

src/whylogs/core/summaryconverters.py
jamie256: Will follow up with Bernease on these questions

PR #447, opened 2022-02-18
Matrix build for Spark and Scala versions

richard-rogers:
### General Checklist

* [ ] Documentation updated


java/spark-bundle/build.gradle.kts
```diff
@@ -14,8 +14,8 @@ plugins {
     id("com.github.johnrengelman.shadow") version ("5.2.0")
 }
 
-val scalaVersion = project.properties.getOrDefault("scalaVersion", "2.12")
-val sparkVersion = project.properties.getOrDefault("sparkVersion", "3.1.1") as String
```
jamie256, line 18: We should at least keep publishing the sparkVersion we were: 3.1.1, looks like your change switches us to 2.4.7 only? Can you update that?

andyndang:
FYI at some point Spark 3.x deprecates Scala 2.11, so you should only build it for Spark 2.x

And from Spark 3.2.0, they support Scala 2.13

java/spark/build.gradle.kts
```diff
@@ -19,8 +19,8 @@ spotless {
     }
 }
 
-val scalaVersion = project.properties.getOrDefault("scalaVersion", "2.12")
-val sparkVersion = project.properties.getOrDefault("sparkVersion", "3.1.1") as String
+val scalaVersion = System.getenv("SCALA_VERSION")
+val sparkVersion = System.getenv("SPARK_VERSION")
```
andyndang: We need the default version FYI

.github/workflows/java-continuous-integration.yml
```diff
@@ -10,6 +10,31 @@ jobs:
...
+          #  spark-version: 3.2.0
+          #- os: ubuntu-latest
+          #  scala-version: 2.12
+          #  spark-version: 3.2.1
+          #- os: ubuntu-latest
+          #  scala-version: 2.13
+          #  spark-version: 3.2.0
+          #- os: ubuntu-latest
+          #  scala-version: 2.13
+          #  spark-version: 3.2.1
```
jamie256: Let's just remove these lines. You can keep some of the changes in a separate branch as we are expanding the versions we build.

jamie256 (approved):
LGTM

PR #460, opened 2022-03-02
Remove internal protobuf class reference

### LINKED ISSUE #459: Remove reference to internal protobuf class MessageMapContainer
### Summary

we import an internal class from protobuf:
```
from google.protobuf.pyext._message import MessageMapContainer
```

which breaks on ARM builds blocking support of M1. This method isn't core to profiling with whylogs but used for debug output and can safely be removed.

jamie256:
## Description

Addresses #459 by removing the internal class reference from protobuf.

### General Checklist

* [ ] Documentation updated


andyndang (approved):
Lgtm!!

PR #468, opened 2022-03-11
Histogram Bin Width and Range

FelipeAdachi:
## Description

We had this 2 bugs:
- The histogram's X axis was too far from the histograms
-  The bin widths were too long and overlapping

hist_bug

To fix it, two changes were made:

- Changed xScale's upper domain to match the max of (max bin value + bin width) between both distributions
- Applied xScale individually to dist[0] and [1] before subtracting to get the bin width (instead of scaling the difference)

### General Checklist

* [ ] Documentation updated


jamie256 (approved):
LGTM! Thanks for making the changes.

PR #469, opened 2022-03-14
Performance Metrics examples

FelipeAdachi:
## Description

- Added two notebooks with examples for logging performance-metrics: Classification and Regression.
- Added images to be displayed within both notebooks
- Added both notebooks to be skipped in test_notebooks, as it would fail, because it needs input from the user

### General Checklist

* [ ] Documentation updated


jamie256 (approved):
LGTM!

PR #485, opened 2022-03-23
fix yscale on double histogram notebook chart

### LINKED ISSUE #479: distribution_chart crops the plot for big differences between distributions
### Summary

The plot gets cropped, giving a wrong idea of the counts in the distribution_chart

### Steps to Reproduce it

When running the code:

```python
session = get_or_create_session()
def profile_generator_1():
    with session.logger("mytestytest", dataset_timestamp=datetime.datetime(2021, 6, 2)) as logger:
        for _ in range(800):
            logger.log({"quality": "bad"})
        for _ in range(250):
            logger.log({"quality": "good"})
        return logger.profile

def profile_generator_2():
    with session.logger("mytestytest", dataset_timestamp=datetime.datetime(2021, 6, 2)) as logger:
        for _ in range(150):
            logger.log({"quality": "bad"})
        for _ in range(80):
            logger.log({"quality": "good"})
        return logger.profile
    
reference_profile = profile_generator_1()
target_profile = profile_generator_2()


visualization = NotebookProfileViewer()
visualization.set_profiles(target_profile=target_profile, reference_profile=reference_profile)

visualization.distribution_chart(feature_names="quality")
```

We have the plot as shown in the image:

dist2

Even though the reference profile has 800 values for "bad", it seems we have only slightly above 150. The same is true for "good" - reference has 250 instances, but that is not reflected in the plot.

jamie256:
Fixes #479 - issue is that the yscale only considered one of the profiles max values when scaling the y axis, which could truncate the bars in one of the profiles.

Also some incidental minor fixes:
Remove placeholder tooltip text, consistently use "target" instead of "current" terminology when comparing profiles.

### General Checklist

* [ ] Documentation updated


FelipeAdachi (approved):
LGTM!

PR #487, opened 2022-03-23
Replace undefined with zeros for missing cell count

jamie256:
## Description

The summary drift report would display undefined when there were no missing values, should be 0.

### General Checklist

* [ ] Documentation updated


FelipeAdachi (approved):
LGTM!

PR #496, opened 2022-03-29
Don't slice using negative indexes to filter  bar charts

jamie256:
## Description

We search for undefined values to filter them, but then also use the -1 to slice the arrays, resulting in wrapping around to the first element and double rendering some bins on our NotebookProfilerViewer charts. The double rendering appears darker with the use of opacity.

The fix is to not try to change the data if there are no undefined values in the findAndDeleteUndefined  function.

### General Checklist

* [ ] Documentation updated


FelipeAdachi (approved):
LGTM!

PR #497, opened 2022-03-29
remove % symbol for total count of missing values

FelipeAdachi:
## Description

Missing value count was displayed between parenthesis and with a % sign at the end. Removed both.

### General Checklist

* [ ] Documentation updated


jamie256 (approved):
LGTM!

PR #498, opened 2022-03-29
pvalue=0 needs to be evaluated as True

FelipeAdachi:
## Description

When pvalue=0, drift was returned as `None`; As 0 is a legit value, drift needs to be 0, and not None.

### General Checklist

* [ ] Documentation updated


jamie256 (approved):
LGTM

PR #499, opened 2022-03-30
dont crash on timedelta64 values

### LINKED ISSUE #500: whylogs throws when processing np.timedelta64[ns] in number tracker
### Summary

passing a timedelta64 value to log_dataframe gets treated as a numeric type and then throws when processed by the variance tracker.

jamie256:
## Description
Fixes #500 

We pass timedelta64 values to number tracker where it throws an error:
```
E       numpy.core._exceptions._UFuncBinaryResolutionError: ufunc 'subtract' cannot use operands with types dtype('<m8[ns]') and dtype('float64')

src/whylogs/core/statistics/datatypes/variancetracker.py:39: UFuncTypeError
```

We skip other time values and unknown types from tracking, so adding a check for this type and also skipping processing with number tracker.

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/src/whylogs/core/columnprofile.py
+++ b/src/whylogs/core/columnprofile.py
@@ -2,8 +2,6 @@
 Defines the ColumnProfile class for tracking per-column statistics
 """
 
-from pandas.api.types import is_timedelta64_dtype
-
 from whylogs.core.statistics import (
     CountersTracker,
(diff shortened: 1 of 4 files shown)


FelipeAdachi (approved):
LGTM!

PR #504, opened 2022-03-31
Fix single-key segmentation bug

FelipeAdachi:
## Description

When segmenting on a single column, segmentation was made, and tagged, based on the first character of said column.
Adding a check on the length of segments, in order to get the proper df subgroup and tags

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/src/whylogs/app/logger.py
+++ b/src/whylogs/app/logger.py
@@ -591,8 +591,12 @@ class Logger:
 
         for each_segment in segments:
             try:
-                segment_df = grouped_data.get_group(each_segment)
-                segment_tags = [{"key": self.segments[i], "value": each_segment[i]} for i in range(len(self.segments))]
+                if len(self.segments) == 1:
+                    segment_df = grouped_data.get_group(each_segment)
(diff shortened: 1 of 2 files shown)


src/whylogs/app/logger.py
```diff
@@ -591,8 +591,12 @@ def log_segments_keys(self, data):
 
         for each_segment in segments:
             try:
-                segment_df = grouped_data.get_group(each_segment)
-                segment_tags = [{"key": self.segments[i], "value": each_segment[i]} for i in range(len(self.segments))]
+                if len(self.segments) == 1:
```
jamie256, line 594: Looks good Felipe! We should consider adding a test for the single dictionary entry case, since we had a bug in this pattern, but I understand your changes work for both single key and single dictionary entry, so let's do that in a follow up change.

e.g. a test that passes something like this:
```
test_segments = [
        [{"key": "home_ownership", "value": "RENT"}],
    ]
```

jamie256 (approved):
LGTM!

Issue #489, opened 2022-03-23
Bump version workflow is broken


jamie256:
### Summary

after release the automoted PR triggers tests that don't report status:
related to https://github.com/peter-evans/create-pull-request/issues/48

for example see:
https://github.com/whylabs/whylogs/pull/488/checks

The PR doesn't reference the version string in the commit or title.

Also we should run this before publishing the wheel so that the version string matches the release number published.

jamie256:
the tests don't trigger because of this I think: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#example-using-more-than-one-event

jamie256:
Fixed by removed required checks on CI that didn't actually run.

PR #510, opened 2022-04-04
Fixes bumpversion to search and replace

TheMellyBee:
### Problem: 
Bump version workflow is broken #489 caused both the version number and a dependency version to be implement. This was due to the search and replace doing a simple search for just the numbers. 

This caused one of the dependencies also on 0.7.0 to be matched and replaced. 

### Fix 
bumpversion has a search and replace keywords that default to:
```
search={current_version}
replace={new_version}
```

Since we have several different file formats each of those needed to be specified 

### Future Maintenance needed
None

### Testing
Manually tested it with the cause that failed (version 0.7.0). Please comment what tests you believe would be helpful on this. 

### General Checklist

* [na] Documentation updated
* [X ] (optional) Please add a label to your PR

## Helpful References
https://pypi.org/project/bumpversion/

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -14,9 +14,17 @@ values =
 	prod
 
 [bumpversion:file:src/whylogs/_version.py]
+search = __version__ = "{current_version}"
+replace = __version__ = "{new_version}"
 
 [bumpversion:file:docs/conf.py]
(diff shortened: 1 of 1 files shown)


richard-rogers (approved):
This hit me once too

PR #511, opened 2022-04-04
Update install protoc action to specify version of protobuf

jamie256:
## Description
Update the install protoc action to specify version of protobuf so we don't install incompatible versions (3.20+drops support for python 3.6 for example and this was being installed). Update both the python and notebook tests workflows.

Also update dependencies since this was a dependency related break and the poetry.lock file was stale.

```
poetry update
```

Update dependencies poetry.lock file only

### General Checklist

* [ ] Documentation updated


TheMellyBee:
LGTM!

TheMellyBee (approved):
LGTM!

Issue #518, opened 2022-04-06
bin collisions in histogram_from_sketch possible


jamie256:
### Summary

If a dataset profile has logged values in a relatively tight range but are large in absolute value, and there are enough bins, the bin edge calculations can hit collisions because of numeric precision on the float values used.

This affects the conversion of the profile to a summary.

jamie256:
mitigation in place to prevent errors due to bin width collisions, will follow up in next version to support calling a double version of the kll_sketch so we have closer to the expected precision coming from python floats.

PR #519, opened 2022-04-06
restrict bin count based on numeric precision on kll floats sketch

jamie256:
## Description

Addresses #518 by reducing the bin count if the width of the bins is lower than rounding of the values.

This is a mitigation for the conversion of profile to summary which uses a default bin count that doesn't work in some cases. We still could use some follow up to better track larger integers not well captured by the floats sketch.

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/src/whylogs/core/summaryconverters.py
+++ b/src/whylogs/core/summaryconverters.py
@@ -124,6 +124,9 @@ def single_quantile_from_sketch(sketch: kll_floats_sketch, quantile: float):
 def _calculate_bins(end: float, start: float, n: int, avg_per_bucket: float, max_buckets: int):
     # Include the max value in the right-most bin
     end += abs(end) * 1e-7
+    abs_end = abs(end)
+    abs_start = abs(start)
+    max_magnitude = max(abs_end, abs_start)
 
     # the kll_floats_sketch use 32bit floats, so we check precision against np.float32
     float_mantissa_bits = np.finfo(np.float32).nmant
@@ -133,7 +136,7 @@ def _calculate_bins(end: float, start: float, n: int, avg_per_bucket: float, max
     width = (end - start) / n_buckets
 
     # check for precision of width with respect to float_mantissa_bits and bin width:
-    bits_in_max = math.floor(math.log2(end))
+    bits_in_max = math.floor(math.log2(max_magnitude))
     width_bits = math.floor(math.log2((end - start) / n_buckets))
     logger.debug(f"bits_in_max is: {bits_in_max}")
     logger.debug(f"width_bits is: {width_bits}")
@@ -143,7 +146,7 @@ def _calculate_bins(end: float, start: float, n: int, avg_per_bucket: float, max
         logger.info(f"Width must be larger than {bits_in_width} bits.")
         new_buckets = math.floor((end - start) / math.pow(2, bits_in_width))
         logger.warn(f"Avoiding bin edge collisions by resizing to {new_buckets} buckets")
-        n_buckets = new_buckets
+        n_buckets = max(new_buckets, 1)
         width = (end - start) / n_buckets
 
     # Calculate histograms from the Probability Mass Function
--- a/src/whylogs/core/summaryconverters.py
+++ b/src/whylogs/core/summaryconverters.py
@@ -145,7 +145,7 @@ def _calculate_bins(end: float, start: float, n: int, avg_per_bucket: float, max
         new_buckets = math.floor((end - start) / min_interval)
         logger.warning(
             f"A bin width of {width} won't work with values in range of [{start}, {end}] "
-            f"because numbers closer to eachother than {int(min_interval)} might not be distinct "
+            f"because numbers closer to each other than {int(min_interval)} might not be distinct "
             "when passed as float32: avoiding bin edge collisions by resizing from: "
             f"{n_buckets} to: {new_buckets} histogram buckets in summary."
         )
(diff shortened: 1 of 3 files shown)


andyndang (approved):
LGTM

src/whylogs/core/summaryconverters.py
```diff
@@ -118,6 +121,41 @@ def single_quantile_from_sketch(sketch: kll_floats_sketch, quantile: float):
     return type("Object", (), {"quantile": qval[0]})
 
 
+def _calculate_bins(end: float, start: float, n: int, avg_per_bucket: float, max_buckets: int):
+    # Include the max value in the right-most bin
+    end += abs(end) * 1e-7
+    abs_end = abs(end)
+    abs_start = abs(start)
+    max_magnitude = max(abs_end, abs_start)
```
richard-rogers, line 129: I would just name this `magnitude`

src/whylogs/core/summaryconverters.py
```diff
@@ -118,6 +121,41 @@ def single_quantile_from_sketch(sketch: kll_floats_sketch, quantile: float):
...
+
+    # the kll_floats_sketch use 32bit floats, so we check precision against np.float32
+    float_mantissa_bits = np.finfo(np.float32).nmant
+
+    # Include the right edge in the bin edges
+    n_buckets = min(math.ceil(n / avg_per_bucket), max_buckets)
+    width = (end - start) / n_buckets
+
+    # check for precision of width with respect to float_mantissa_bits and bin width:
+    bits_in_max = math.floor(math.log2(max_magnitude))
```
richard-rogers:
```
# Figure out the floating point precision at the scale of the bin boundaries
# min_interval is the smallest difference between floats at this scale
log_min_interval = math.floor(math.log2(magnitude)) - float_mantissa_bits
min_interval = 2^^log_min_interval
```

src/whylogs/core/summaryconverters.py
```diff
@@ -118,6 +121,41 @@ def single_quantile_from_sketch(sketch: kll_floats_sketch, quantile: float):
...
+    width_bits = math.floor(math.log2((end - start) / n_buckets))
+    logger.debug(f"bits_in_max is: {bits_in_max}")
+    logger.debug(f"width_bits is: {width_bits}")
+
+    if bits_in_max - float_mantissa_bits > width_bits:
```
richard-rogers:
```
# If the bin width is smaller than min_interval, we need bigger bins
if width < min_interval:
    new_buckets = math.floor((end - start) / min_width
    n_buckets = max(new_buckets, 1)
    width = (end - start) / n_buckets

**Review by richard-rogers — CHANGES_REQUESTED** [review]

File: src/whylogs/core/summaryconverters.py [hunk 4]
```diff
@@ -118,6 +121,41 @@ def single_quantile_from_sketch(sketch: kll_floats_sketch, quantile: float):
     return type("Object", (), {"quantile": qval[0]})
 
 
+def _calculate_bins(end: float, start: float, n: int, avg_per_bucket: float, max_buckets: int):
+    # Include the max value in the right-most bin
+    end += abs(end) * 1e-7
```
richard-rogers, line 126: This will still be 0 if end is passed in as 0... don't want to take log of 0. Probably OK if start != end, but we should ensure we don't take log of 0.

richard-rogers (changes requested):

src/whylogs/core/summaryconverters.py
jamie256: we could but the value is computed considering both the start and end values, so I wanted to help make it more explicit that this was the largest magnitude we have around the bin boundaries.

src/whylogs/core/summaryconverters.py
jamie256: Sure, this is a bit easy to read, done.

src/whylogs/core/summaryconverters.py
jamie256: This considers the bin widths rather than the bits in resolution, maybe this is easier to read, done.

src/whylogs/core/summaryconverters.py
jamie256: I don't think you can get this to call log 0, because we don't log2(end), we log2((end-start)/buckets), and we check for start == end already. I added some new tests to cover when start and end are close to each other and close to 0, also having end == 0.

src/whylogs/core/summaryconverters.py
```diff
@@ -118,6 +121,45 @@ def single_quantile_from_sketch(sketch: kll_floats_sketch, quantile: float):
...
+    # min_interval is the smallest difference between floats at this scale
+    log_min_interval = math.floor(math.log2(max_magnitude)) - float_mantissa_bits
+    min_interval = math.pow(2, log_min_interval)
+
+    # If the bin width is smaller than min_interval, we need bigger bins
+    if width < min_interval:
+        new_buckets = math.floor((end - start) / min_interval)
+        logger.warning(
+            f"A bin width of {width} won't work with values in range of [{start}, {end}] "
+            f"because numbers closer to eachother than {int(min_interval)} might not be distinct "
```
richard-rogers: s/eachother/each other/

richard-rogers (changes requested):

src/whylogs/core/summaryconverters.py
jamie256: done.

richard-rogers (approved):

PR #536, opened 2022-04-11
fix HTML display for mismatching number of frequent items between profiles

FelipeAdachi:
## Description

When `reference_profile` had a higher number of `frequentItems` than `target_profile`, `profile_viewer` generated broken HTML file, because the logic iterated on `target_profile` and used index to try to access `reference_profile`.

Changed to:
- Iterate based on min. number of frequentItems
- frequentItem Strings loaded w/ empty string when trying to access not-existing index

### General Checklist

* [ ] Documentation updated


jamie256 (approved):
Nice find on the root cause Felipe, thanks for getting this fixed!

Issue #527, opened 2022-04-08
NotebookProfileViewer: Error messages in distribution_chart and double_histograms are not helpful


FelipeAdachi:
### Summary

When using `distribution_chart` or `double_histogram` with a feature of unsupported type, the error message just shows `Something went wrong.`, instead of actually saying what is wrong about it.

### What is the expected *correct* behavior?

When something goes wrong, the message should be clear on what's wrong. For example, if categorical variables are not supported for `double_histograms`, or similarly for `distribution_chart`

### Relevant logs and/or screenshots

![image](https://user-images.githubusercontent.com/29715545/162487925-5f579ff9-7407-4204-8cde-4020c36e2752.png)

jamie256:
Also remove the prompt to try again, because that incorrectly implies this is an intermittent issue (which in practice it is something with the way we call the method or the data).

TheMellyBee:
This error is html-generated here:

https://github.com/whylabs/whylogs/blob/04f3c31baf0abd727f360805cfc6a2a1e5f53835/src/whylogs/viewer/templates/index-hbs-cdn-all-in-jupyter-distribution-chart.html#L271

The one above has an error handling that we can used to get a more specific error, but other ones don't. They may be able to be redone to have similar error handling.

PR #543, opened 2022-04-18
Enhancing the Error Message on Notebook Profile Vieweer

TheMellyBee:
### Problem
When using distribution_chart or double_histogram with a feature of unsupported type, the error message just shows Something went wrong., instead of actually saying what is wrong about it.

### Changes Made
1. Added descriptive errors for the categorical and numerical displays
2. Where error was thrown displaying that error
3. Making the UI for the error a little more pleasant (small change, reducing font by 2px due to size on notebook, and desaturated color very slighting)

### Things to Think about
1. Do we want to for the type of error and go for that? If so what ones do we want to catch? 
I did notice undefined is really the only one that seams to make it all the way to that error. Everything else gets caught before then. 
 
## Testing
Manual tested in the workbook

## Linked to
#527

### General Checklist

* [X ] Conform by the style guides, by using formatter
* [ ] Documentation updated


TheMellyBee:
Wooh that took a while! Not due to difficulty in code, but because my testing environment was misbehaving. Finally got it working and manual testing done.

src/whylogs/viewer/templates/index-hbs-cdn-all-in-jupyter-bar-chart.html
```diff
@@ -279,7 +279,7 @@ <h1 class="no-responsive__title">Hold on! :)</h1>
             $(document).ready(() => 
               $(".desktop-content").html(`
                 <p style="height: ${$(window).height()}px" class="error-message"> 
-                  Something went wrong. Please try again.
+                    Error: column must be a stringSummary
```
FelipeAdachi: I think we shouldn't assume that the user knows what stringSummary is. Also, it's somewhat confusing, because even though the code relies on stringSummary, the clearest distinction between both plots should be between categorical/numerical values.

Maybe something like "Error: feature must be either categorical or textual" ?

src/whylogs/viewer/templates/index-hbs-cdn-all-in-jupyter-distribution-chart.html
```diff
@@ -268,7 +268,7 @@ <h1 class="no-responsive__title">Hold on! :)</h1>
             $(document).ready(() => 
               $(".desktop-content").html(`
                 <p style="height: ${$(window).height()}px" class="error-message"> 
-                  Something went wrong. Please try again.
+                     Histogram failed. Check variables are supported.
```
FelipeAdachi: I think "variables are supported" still seems a bit vague.
Can we ask for numerical, like:
"Histogram Failed. Features must be of numerical type."
?

FelipeAdachi (approved):
LGTM!

bernease:
Also LGTM despite the failing notebook test.

PR #548, opened 2022-04-25
Bug/np float issue

### LINKED ISSUE #546: After whylogs upgrade float32 values are counted as  type_unknown_count  in flat_table.
### Steps to Reproduce it

1. Create a np.float32 data that works in whylogs 0.6.17
1. Upgrade from whylogs from 0.6.17 to 0.72
2. np.float32 will be counted as. `type_unknown_count`

### Example
```
>>> typed_data
0          0.000
1          0.000
2          0.000
3          0.015
4          0.451
           ...  
5054752    0.000
5054753    0.041
5054754    0.000
5054755    0.041
5054756    0.000
Name: column1, Length: 5054757, dtype: float32
>>> TypedDataConverter.get_type(typed_data)
0
```

It shows the type as 0 or unknown_type for floats

`TYPES.FRACTIONAL` with `FLOAT_TYPES = (float, np.float) `

which does not include the np.float32 and np.float5=64

This should work for np.float32 and np.float64

Update FLOAT_TYPES to have these two as well

### Relevant logs and/or screenshots
See example above

### Did it work on a previous version ?

Yes 0.6.17

TheMellyBee:
## Description

This fixes #546 allowing for use of np.float32 by changing the np.float we were accepting to np.floating. Note that the np.floating is the np API for handing floating points of all scalars and the previous were deprecated. 

https://numpy.org/devdocs/reference/arrays.scalars.html#numpy.floating

Adds testing for this use case and to make sure merge is still working, but tests are not in place around downcasting. Further work needs to be done on ensuring we can work with float64 depending on the datasketches fork https://github.com/whylabs/whylogs-sketching. 

### General Checklist

* [X] Tests added for this feature/bug
* [ ] Documentation updated
* [X] (optional) Please add a label to your PR

### DIFF
--- a/tests/unit/core/statistics/datatypes/test_floattracker.py
+++ b/tests/unit/core/statistics/datatypes/test_floattracker.py
@@ -68,6 +68,7 @@ def test_merge_floattrackers_should_addup():
     merge_second = second.merge(first)
     assert merge_second.__dict__ == merge_first.__dict__
 
+
 def test_merge_floattrackers_should_addup():
     float32_tracker = FloatTracker()
     float32_array = np.array([1.0, 2.0, 3.0], dtype=np.float32)
@@ -83,12 +84,6 @@ def test_merge_floattrackers_should_addup():
     vals2 = [4.0, 5.0, 6.0]
     for val in vals2:
         simple_tracker.update(val)
-
-    assert simple_tracker.count == len(vals2)
-    assert simple_tracker.max == max(vals2)
-    assert simple_tracker.min == min(vals2)
-    assert simple_tracker.sum == sum(vals2)
-
 
     merge_32_and_simple = float32_array.tolist() + vals2
     merge_first_tracker_32_and_simple = float32_tracker.merge(simple_tracker)
--- a/tests/unit/core/statistics/datatypes/test_floattracker.py
+++ b/tests/unit/core/statistics/datatypes/test_floattracker.py
@@ -1,4 +1,5 @@
 import numpy as np
+
 from whylogs.core.statistics.datatypes import FloatTracker
 
 
--- a/tests/unit/core/statistics/datatypes/test_floattracker.py
+++ b/tests/unit/core/statistics/datatypes/test_floattracker.py
@@ -1,3 +1,4 @@
+import numpy as np
 from whylogs.core.statistics.datatypes import FloatTracker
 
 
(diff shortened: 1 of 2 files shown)


tests/unit/core/statistics/datatypes/test_floattracker.py
```diff
@@ -43,3 +67,55 @@ def test_merge_floattrackers_should_addup():
...
+    float64_array = np.array([1.0, 2.0, 3.0], dtype=np.float64)
+    for val in float64_array:
+        float64_tracker.update(val)
+
+    simple_tracker = FloatTracker()
+    vals2 = [4.0, 5.0, 6.0]
+    for val in vals2:
+        simple_tracker.update(val)
+
+    assert simple_tracker.count == len(vals2)
```
murilommen: aren't these assertions already covered up on `test_values_are_min_max()`?

murilommen (commented):
everything looks good, only a very small comment on a possible duplicity. good job :)

tests/unit/core/statistics/datatypes/test_floattracker.py
TheMellyBee: I was thinking the same thing :) I was curious why the first test did it that way. I'll update as I was feeling weird about that as well. 

What I would like to do for the testing is break this up so that we can send in the type and test like and not have to have so much duplicating.

tests/unit/core/statistics/datatypes/test_floattracker.py
```diff
@@ -13,6 +15,28 @@ def test_values_are_min_max():
...
+    float32_array = np.array([1.0, 2.0, 3.0], dtype=np.float32)
+    for val in float32_array:
+        first.update(val)
+
+    assert first.count == len(float32_array)
+    assert first.max == max(float32_array)
+    assert first.min == min(float32_array)
+    assert first.sum == sum(float32_array)
+
+    second = FloatTracker()
```
jamie256: suggest naming something like 'tracker_2' to avoid overloaded term with unit of time 'second'

consider using parameterized test inputs e.g.
https://github.com/whylabs/whylogs/blob/mainline/tests/unit/core/types/test_typeddataconverter.py#L38-L39

tests/unit/core/statistics/datatypes/test_floattracker.py
```diff
@@ -13,6 +15,28 @@ def test_values_are_min_max():
     assert first.sum == sum(vals1)
 
 
+def test_np_float():
+    first = FloatTracker()
+    float32_array = np.array([1.0, 2.0, 3.0], dtype=np.float32)
+    for val in float32_array:
+        first.update(val)
+
+    assert first.count == len(float32_array)
```
jamie256: minor: consider creating a small test helper function: check_float_tracker(float_tracker, data) to wrap these max/min/sum checks to reduce duplication.

jamie256: looks like you did that in last revision!

tests/unit/core/statistics/datatypes/test_floattracker.py
TheMellyBee: How would you use parametrized tests in this case? Would it be to replace the setup like `    float32_array = np.array([1.0, 2.0, 3.0], dtype=np.float32)`

jamie256: You could do that and parameterize the data, or you could parameterize only the dtype:
`_TEST_NP_FLOAT_DTYPES = [np.float16, np.float32, np.float64, ]`

But what you have looks good and fixes a bug, let's merge.

jamie256 (approved):
LGTM!

PR #552, opened 2022-05-02
Adding profile visualizer screencap

dleybz:
## Description

Adding a profile visualizer screencap for the readme

### General Checklist

None of these checks apply :)


jamie256 (changes requested):
Does this need to be in our repo in mainline?

let's try using an "assets" branch: https://github.com/whylabs/whylogs/pull/554

jamie256:
already updated.

PR #553, opened 2022-05-02
Fix for webbrowser filepaths in WSL

### LINKED ISSUE #531: profile_viewer fails when run in a notebook in WSL: The sysem cannot find the file specified
### Summary

On a WSL setup (Windows Subsystem for Linux), if running the profile_viewer in a notebook, the logic tries to interact with files in the WSL but it can't find them from the browser and hits an error: "The system cannot find the file specified"

### Steps to Reproduce it
on a WSL:
1 use the profile_viewer to compare profiles and output a test.html file:
```
from whylogs.viz import profile_viewer
# create some profiles to pass in.
profile_viewer(profiles=target_profiles, reference_profiles=reference_test_profiles, output_path="test.html")
```

The viz logic could look for the files int the standard location mounted from windows for the WSL, and this would work.

jamie256:
## Description
Fixes #531 by detecting if in WSL and prefixing the output path accordingly before opening with webbrowser
+ minor updates to notebook while testing.

* verified on WSL2 running https://github.com/whylabs/whylogs/blob/mainline/examples/Dataset_Profiler_Viewer.ipynb in VSCode python 3.8

### General Checklist

* [ ] Documentation updated

### DIFF
--- a/examples/Dataset_Profiler_Viewer.ipynb
+++ b/examples/Dataset_Profiler_Viewer.ipynb
@@ -49,10 +49,20 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
    "id": "1232b2a8",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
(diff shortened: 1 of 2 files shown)


FelipeAdachi (approved):
LGTM!

Issue #441, opened 2022-02-15
Using both with_rotations_time and dataset_timestamp doesn't work


jamie256:
### Summary

whylogs logger should validate parameters to avoid problems setting both dataset_timestamp and with_rotation_time.

Specifically the problem here is the two params: 
```
dataset_timestamp=datetime(2022, 2, 12, 0, 0, 0),
                    with_rotation_time=rotation_time_unit
```

PR #557, opened 2022-05-06
Add warning if using both rotation and setting dataset_timestamp

jamie256:
## Description

log rotation doesn't use a passed in dataset_timestamp, add warning to help identify the pitfall. Mitigation fix for #441

### General Checklist

* [ ] Documentation updated


FelipeAdachi (approved):
This is a pitfall I fell into once! Very helpful.

PR #576, opened 2022-05-25
Adding pyspark example

murilommen:
## Description

Adding an example Jupyter Notebook to demonstrate how we've implemented our **experimental** pyspark integration. Feel free to read through the notebook and suggest anything that should be different.

### General Checklist

* [x] Documentation updated

### DIFF
--- a/python/whylogs/api/pyspark/experimental/__init__.py
+++ b/python/whylogs/api/pyspark/experimental/__init__.py
@@ -1,5 +1,6 @@
 from .profiler import (
     collect_column_profile_views,
+    collect_dataset_profile_view,
     column_profile_bytes_aggregator,
     whylogs_pandas_map_profiler,
 )
@@ -8,4 +9,5 @@ __ALL__ = [
     whylogs_pandas_map_profiler,
     column_profile_bytes_aggregator,
     collect_column_profile_views,
+    collect_dataset_profile_view,
 ]
--- a/python/whylogs/api/pyspark/experimental/profiler.py
+++ b/python/whylogs/api/pyspark/experimental/profiler.py
@@ -1,10 +1,12 @@
+from datetime import datetime
 from functools import reduce
 from logging import getLogger
-from typing import Dict, Iterable, Tuple
+from typing import Dict, Iterable, Optional, Tuple
 
 import whylogs as why
 from whylogs.core.stubs import pd
 from whylogs.core.view.column_profile_view import ColumnProfileView
+from whylogs.core.view.dataset_profile_view import DatasetProfileView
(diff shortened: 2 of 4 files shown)


python/whylogs/api/pyspark/experimental/__init__.py
```diff
@@ -1,5 +1,6 @@
 from .profiler import (
     collect_column_profile_views,
+    collect_dataset_profile_view,
```
TheMellyBee, line 3: Good for now, but I'd like to think through the API names here. They are a bit long for the API although they are accurate. something like DatasetCollecter or something like that may be a easier to use, but still accurate?

TheMellyBee (approved):
This looks great. Couple things to check on:
1. Can this have an Open in Collab button?
2. Check to see if there is anyway to resolving the warning/errors. Especially the two blanks ones.

python/whylogs/api/pyspark/experimental/profiler.py
```diff
@@ -56,3 +58,20 @@ def collect_column_profile_views(input_df: SparkDataFrame) -> Dict[str, ColumnPr
...
+) -> DatasetProfileView:
+    now = datetime.utcnow()
+
+    _dataset_timestamp = dataset_timestamp or now
+    _creation_timestamp = creation_timestamp or now
+
+    column_views_dict = collect_column_profile_views(input_df=input_df)
+
+    profile_view = DatasetProfileView(
+        columns=column_views_dict, dataset_timestamp=_dataset_timestamp, creation_timestamp=_creation_timestamp
```
jamie256, line 74: nice!

jamie256 (approved):
LGTM!

python/whylogs/api/pyspark/experimental/__init__.py
jamie256: I'm with you on these being names we likely change, which is partly why these are in "experimental" namespace. I'm not against longer explicit naming for advanced scenarios in particular. Maybe we can get some input from users in our RC on this?

jamie256:
> This looks great. Couple things to check on:
> 
> 1. Can this have an Open in Collab button?
> 2. Check to see if there is anyway to resolving the warning/errors. Especially the two blanks ones.

I also agree with Melly's suggestions, but we can do in a follow up too. The warnings I think Melly is referring to are the "red" cell output in the notebook example.

Also you might try this command to install whylogs[spark]:
```!pip install "whylogs[spark]" --pre```

andyndang:
FYI you can't run Pyspark in Collab. They don't have Java

PR #580, opened 2022-05-26
Examples: Local and s3 Writers

murilommen:
## Description

Adding a notebook to explain how to interact with our `local` and `s3` writers APIs.

### General Checklist

* [x] Documentation updated


python/examples/integrations/writing_profiles.ipynb
```diff
@@ -0,0 +1,695 @@
+{
```
FelipeAdachi, line 1: can you add this example in `examples.rst`?

python/examples/integrations/writing_profiles.ipynb
```diff
@@ -0,0 +1,695 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import whylogs as why\n",
+    "\n",
+    "profile = why.log(df)"
```
jamie256: Suggest we name the output here something like "profile_results"

In most cases it is equivalent to the "profile" conceptually, but the reason we have this is so that we can write a collection of profiles at once (if there are more than one profile returned, which currently we only get one, but in the case of segments we might have more).

jamie256 (approved):
LGTM, minor variable name suggestion but its ok as is too.

python/examples/integrations/writing_profiles.ipynb
```diff
@@ -0,0 +1,695 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Writing profiles\n",
+    "\n",
```
FelipeAdachi, line 8: I think this one can be run from colab, right? If so, let's add it!

python/examples/integrations/writing_profiles.ipynb
```diff
@@ -0,0 +1,695 @@
...
+    "Cool! So now that we have a proper profile created, let's see how we can persist is as a file."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Local writer\n",
+    "\n",
+    "The first and most straight-forward way of persisting a `whylogs` profile as a file is to write it directly to your disk. Our API makes it possible with the following commands. You could either write it from the `ProfileResultSet`"
```
FelipeAdachi, line 371: Maybe we could add the why.write(profile,"dest") way of writing, and then explain that this converts the profile into a profile view before writing it to disk.

FelipeAdachi (changes requested):

TheMellyBee (approved):
Looks great! Thanks for the readme :D

python/docs/examples.rst
```diff
@@ -16,6 +16,9 @@ Basic
     examples/basic/Notebook_Profile_Visualizer
     examples/basic/Schema_Configuration
     examples/basic/Logging_Different_Data
+    examples/integrations/Writing_Profiles
```
FelipeAdachi: Actually, can you put it under `integrations` section below, in this same file?
And I had already included the pyspark one in a previous PR 😄

FelipeAdachi (changes requested):
I think those binary files are not needed, right?

python/docs/examples.rst
murilommen: done :)

FelipeAdachi (approved):
LGTM!

PR #581, opened 2022-05-26
Add server side config check for whylabs v1 profiles enabled

jamie256:
## Description

Add a check for server side config so we can turn on WhylabsWriter when the backend support is enabled.

* Incidental fix to align env var names with what we used in v0: ["WHYLABS_DEFAULT_ORG_ID"] 

### General Checklist

* [ ] Documentation updated


andyndang (approved):
LGTM

PR #585, opened 2022-05-26
Readme refresh

dleybz:
Big update on readme


jamie256 (commented):
Danny, we generate the overview from the README.md in the whylogs/python when in the 1.0.x branch.

PR #586, opened 2022-05-26
Examples of custom Metrics

richard-rogers:
### General Checklist

* [ ] Documentation updated


murilommen (dismissed):
im a bit lost when the notebook begins with "If all of the state of the metric can be represented by subclasses of `MetricComponent`". I think some context, such as what Felipe's done for his examples [here](https://github.com/whylabs/whylogs/blob/1.0.x/python/examples/basic/Notebook_Profile_Visualizer.ipynb) is very welcome to introduce what a custom metric is and why it is important.

It's also hard to catch the endgame of having a custom metric tracked. How does it compare to the standard metrics we already have in a profile? And how can i see exactly what i've created with those custom classes?

FelipeAdachi:
@jamie256 what other examples do you have in mind for the `extensions` folder? we can keep it, but if there's only this one we can think of, maybe we should group it in an `advanced` folder?

jamie256:
@richard-rogers I think you need an:
```from whylogs.core import ColumnSchema```

might also be worth printing the `prof.view().to_pandas() ` at the end.

I agree with other feedback around explanation or changing the folder, but ok with iterating here.

File: python/examples/extensions/Custom_Metrics.ipynb [hunk 1]
```diff
@@ -0,0 +1,216 @@
...
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "HlFNH-H6-qKg"
+      },
+      "outputs": [],
+      "source": [
+        "class TestResolver(Resolver):\n",
+        "    def resolve(self, name: str, why_type: DataType, column_schema: ColumnSchema) -> Dict[str, Metric]:\n",
```
jamie256: This line giving me an error when run in colab, are you missing:
```from whylogs.core import ColumnSchema```

```NameError                                 Traceback (most recent call last)
[<ipython-input-4-7cba2972d2b3>](https://localhost:8080/#) in <module>()
----> 1 class TestResolver(Resolver):
      2     def resolve(self, name: str, why_type: DataType, column_schema: ColumnSchema) -> Dict[str, Metric]:
      3         return {"histogram": HistogramMetric.zero(column_schema.cfg),
      4                 "struct": StructMetric(0, "")}
      5 

[<ipython-input-4-7cba2972d2b3>](https://localhost:8080/#) in TestResolver()
      1 class TestResolver(Resolver):
----> 2     def resolve(self, name: str, why_type: DataType, column_schema: ColumnSchema) -> Dict[str, Metric]:
      3         return {"histogram": HistogramMetric.zero(column_schema.cfg),
      4                 "struct": StructMetric(0, "")}
      5 

NameError: name 'ColumnSchema' is not defined```

File: python/examples/extensions/Custom_Metrics.ipynb [hunk 2]
```diff
@@ -0,0 +1,216 @@
...
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "id": "WT8XU-Df065F"
+      },
+      "outputs": [],
+      "source": [
+        "!pip install whylogs --pre"
```
FelipeAdachi: we should remove the --pre and replace "!" with "%", and also comment it, to prevent it being runned every time the notebook is being tested:
`# %pip install -q whylogs`

jamie256 (changes requested):
Still hitting errors running this notebook mentioned in comments, but if we fix that we can merge and then have others add more explanation or build out the scenario more.
```
[<ipython-input-4-7cba2972d2b3>](https://localhost:8080/#) in TestResolver()
      1 class TestResolver(Resolver):
----> 2     def resolve(self, name: str, why_type: DataType, column_schema: ColumnSchema) -> Dict[str, Metric]:
      3         return {"histogram": HistogramMetric.zero(column_schema.cfg),
      4                 "struct": StructMetric(0, "")}
      5 

NameError: name 'ColumnSchema' is not defined```
```

FelipeAdachi:
Ok, I added `from whylogs.core import ColumnSchema` and fixed other minor issues.
I think we're good to merge it. As Jamie said, let's iterate on this later. Add explanation and motivation to customize your own metric, etc.

python/examples/extensions/Custom_Metrics.ipynb
FelipeAdachi: this was fixed in the last commit

jamie256 (approved):
Thanks for making the refactor fix @FelipeAdachi lets merge and iterate.

PR #588, opened 2022-05-27
Update to readme (in the right location)

dleybz:
Still need to add links to constrains and visualizer canonical examples, update icons for three properties, update examples/links for integrations, and (optional) upload profile visualizer gif in assets so that it's properly a gif. When image logger is available, will remove the comment currently hiding it


FelipeAdachi (approved):
LGTM!
Thanks, Danny!

PR #589, opened 2022-05-27
Images directory is redundant; can delete

dleybz:
Still need to add links to constrains and visualizer canonical examples, update icons for three properties, update examples/links for integrations, and (optional) upload profile visualizer gif in assets so that it's properly a gif. When image logger is available, will remove the comment currently hiding it


dleybz:
Oops, meant to do this on v1.0.x

PR #593, opened 2022-05-27
Populate usage_statistics.rst

dleybz:
Based on https://docs.whylabs.ai/docs/usage-statistics


python/docs/support/usage_statistics.rst
```diff
@@ -1,17 +1,22 @@
...
+Python version
+Execution environment (Sagemaker, Google Colab, Jupyter Notebook, etc.)
+Relevant libraries in the runtime environment (numpy, pandas, etc.)
+This data helps our developers to deliver the best possible product to our users. This information can be used to help inform our team of the best areas to focus development and better understand how our users are utilize whylogs.
 
-.. code-block:: bash
+Data is only collected at the time of importing the whylogs library or its modules. We do not collect any sensitive information, or user code.
 
-   pip install whylogs
+If users wish to opt out of this usage statistics collection, they can do so by setting the WHYLOGS_NO_ANALYTICS environment variable as follows:
```
murilommen: i would make the env var with the correct "variable syntax" (like `this`)

murilommen: or in rst should be :code:`WHYLOGS_NO_ANALYTICS`

murilommen (changes requested):

python/docs/support/usage_statistics.rst
dleybz: Good call

murilommen (approved):

PR #598, opened 2022-05-28
MetricConstraints with example notebook

jamie256:
## Description

Initial metric constraints API with lambda support and some examples.

### General Checklist

* [ ] Documentation updated


python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -0,0 +1,223 @@
...
+                selectors.append(MetricsSelector(metric_name=metric_path, column_name=column_name))
+        return selectors
+
+    def add_constraint(self, constraint: MetricConstraint, ignore_missing: bool = False) -> "ConstraintsBuilder":
+        # check that the column exists and we have a column profile view
+        column_name = constraint.metric_selector.column_name
+        column_profile_view = self._dataset_profile_view.get_column(column_name)
+        if column_profile_view is None and not ignore_missing and constraint.metric_selector.metrics_resolver is None:
+            raise ValueError(
+                f"{column_name} was not found in set of this profile, the lis of columns are: {self._dataset_profile_view.get_columns()}"
```
bernease: Typo: lis -> list

bernease (commented):
Thanks Jamie for churning well into the night on getting a draft for this! First I will say that while I have some high level design suggestions that follow, I do think this design is workable and I'm happy to disagree and commit. In that case, the PR looks pretty good, just some minor fixes in the section below plus some notebook improvements (omitted for now). Leaving as a comment for now, but happy to switch to approve once you glance it over / others weigh in.

High level design:
* I like the switch to standalone ConstraintsBuilder class!
* I think the API you've presented here is easy enough to understand, though quite verbose due to the many objects that need to be instantiated (ConstraintsBuilder, then a separate MetricsSelector for each column/metric combination, then a metrics component, then the add_component method on top of that). Upon reflection from our squad D&R, my theory on discoverability in Python goes methods (most discoverable) > classes >  parameters > string configurations but classes and parameters plummet rapidly on discoverability if you need multiple that have different purposes. 
* I know that this "custom" constraint example is just the building block for adding helper methods in the future, but it's hard to see the value of the MetricsSelector object other than being an object to pass a list of available column/metric combinations. 
* THe metrics_resolver parameter feels similar in spirit to the Projector which is used a few places in v1. But I'm not sure it works great for (summary) constraints, which are more analysis-driven and should have a more simple API. As a user, I'd rather repeat a column name or metric over and over specifying the metrics in list form as opposed to writing a custom function in the analysis use case especially. I'd honestly rather see one less class, but interested in opinions.

API suggestion if you're open to it:
* If you're open to reducing number of classes (fine if not), maybe something like: `ConstraintsBuilder().add_custom_constraint(constraint_name, column_name, condition) [or `.add_custom_constraint(constraint_name, column_name, metric, condition)` but I'm not sure why we need to specify the metric here vs. in the condition, maybe necessary for implementation] for the use case in this PR? Perhaps `ConstraintsBuilder().add_constraint(constraint_name, column_name, condition=MaxLessThanEqualConstraint(less_than_equal=7))` is a more generic API for this where condition: Union[Callable, ConstraintHelperMethod].

Minor stuff:
* We should rename the ipynb to have underscore (Metric_Constraints.ipynb) to match existing pattern.
* Instead of listing column names with `profile_view.to_pandas().index.values`, how about `profile_view.get_columns().keys()`, which seems a slightly more clear and efficient.

FelipeAdachi (approved):
LGTM, I don't have specific requests for change, I think we can merge and iterate.

In addition to what Bernease already suggested, just some thoughts for later on:
- Do we want to enable the user to remove added constraints? In an experimentation environment, these constraints will be updated often. The best way to update right now would be to call a new builder instance and start from scratch. If we want to enable removing constraints, we need a way to identify each constraint in the list.
- In the future, we'll need to think about how to add table-shape constraints.
- More complex constraints, like the Distributional Measure Ones, might have more elaborate logic that don't fit in a lambda expression. But if it accepts a callable, there should be ways to fit it into the current implementation.

python/whylogs/core/constraints/metric_constraints.py
jamie256: Typo fixed with Murilo's updated tests (thanks Murilo!)

jamie256:
> API suggestion if you're open to it:
> 
> * If you're open to reducing number of classes (fine if not), maybe something like: `ConstraintsBuilder().add_custom_constraint(constraint_name, column_name, condition) [or `.add_custom_constraint(constraint_name, column_name, metric, condition)`but I'm not sure why we need to specify the metric here vs. in the condition, maybe necessary for implementation] for the use case in this PR? Perhaps`ConstraintsBuilder().add_constraint(constraint_name, column_name, condition=MaxLessThanEqualConstraint(less_than_equal=7))` is a more generic API for this where condition: Union[Callable, ConstraintHelperMethod].

Hey Bernease, we chatted a little outside of this PR but trying to loop back for visibility. My explanation was that the number of classes is not ideal and I agree with your feedback that we should simplify this part of how people will define constraints, but that this is the core API that we will build the simplifications on top of, and this will remain a valid way to define constraints (for extensibility purposes). Some of the indirection here is so that you can re-use a constraint on new profiles (thats why we need a way to select which metric / and define an expression on it rather than just evaluate the expression on a value).

Murilo's test shows the kinds of helpers we could create, do you think we should just pull some of these test methods ```no_null``` and ```greater_than_number``` into the main API now?, e.g.

```
...
    builder.add_constraint(not_null(column_name="weight"))
    builder.add_constraint(not_null(column_name="animal"))
    builder.add_constraint(not_null(column_name="legs"))
    builder.add_constraint(greater_than_zero(column_name="weight"))
    builder.add_constraint(greater_than_zero(column_name="legs"))
    builder.add_constraint(greater_than_number(column_name="weight", number=10))
    builder.add_constraint(greater_than_number(column_name="legs", number=20))
    ...
```

I was thinking we could have the building blocks in initial launch and then scale out convenience methods to make this easier? Thoughts?

andyndang (approved):
LGTM!

PR #606, opened 2022-05-31
Make sure we are explicitly using a version of protoc in workflows

jamie256:
## Description

To avoid using an older version of protoc that causes trouble with protobuf 4.0 we should explicitly set the version in our workflows.

- poetry update
- Also updating to version rc4 and will publish new release candidate with updated protobuf messages.

### General Checklist

* [ ] Documentation updated


murilommen (approved):
LGTM

PR #616, opened 2022-05-31
Add unicode_range tracker to strings

jamie256:
Also allow overriding default config with passed in MetricConfig
Add tests showing how to disable frequent item tracking, and unicode_range metrics


python/whylogs/core/schema.py
```diff
@@ -123,6 +126,7 @@ def _resolve_pdf(self, df: pd.DataFrame, force_resolve: bool = False) -> bool:
             col_dtype = df.dtypes[col_name]
             self._columns[col_name] = ColumnSchema(
                 dtype=col_dtype,
+                cfg=self.default_configs,
```
jamie256, line 130: We need to pass the frozen MetricConfig to the ColumnSchema so that config will apply to each column.

python/whylogs/core/schema.py
```diff
@@ -102,6 +105,7 @@ def resolve(self, *, pandas: Optional[pd.DataFrame] = None, row: Optional[Mappin
 
                 self._columns[k] = ColumnSchema(
                     dtype=type(v),
+                    cfg=self.default_configs,
```
jamie256, line 108: pass the frozen MetricConfig to the ColumnSchema so that config will apply to each column.

python/tests/api/logger/test_logger.py
```diff
@@ -7,6 +7,8 @@
 
 import whylogs as why
 from whylogs.core import ColumnProfileView
+from whylogs.core.metrics.metrics import MetricConfig
```
andyndang: Should `MetricConfig` be a the `whylogs.core` import level?

python/whylogs/core/metrics/metrics.py
```diff
@@ -40,6 +40,7 @@ class MetricConfig:
     kll_k: int = 256
     fi_lg_max_k: int = 10  # 128 entries
     fi_disabled: bool = False
+    track_unicode_ranges: bool = True
```
andyndang: Do we have some quantifiable measurements on how much perf / memory footprint on the profiling step?

python/whylogs/core/schema.py
```diff
@@ -61,9 +61,12 @@ class DatasetSchema:
     cache_size: int = 1024
     schema_based_automerge: bool = False
 
-    def __init__(self) -> None:
+    def __init__(self, config: MetricConfig = None) -> None:
```
andyndang:
```suggestion
    def __init__(self, config: Optional[MetricConfig] = None) -> None:
```

Also we should consider other params here as well (not a blocking comment)

andyndang (commented):
Leaving some comment above. Most notable is I would like to understand the perf impact before turning this on by default

python/tests/api/logger/test_logger.py
richard-rogers: I like it next to `Metric` where it's used. It would be fishy to import it from a level above `core/metrics`, but it could be made a peer of `metrics.py`. It would be a small file though.

python/tests/api/logger/test_logger.py
```diff
@@ -144,3 +146,64 @@ def test_bool_count():
     prof_view = results.profile().view()
     assert prof_view._columns.get("fly")._metrics.get("types").boolean.value == 4
     assert prof_view._columns.get("fly")._metrics.get("types").integral.value == 0
+
+
+def test_unicode_range_defaults() -> None:
+    strings = {
+        "words": ["1", "12", "123", "1234a", "abc", "abc123", "I😍emoji"],
+    }  # fix ❤️  /u+fe0f
```
richard-rogers: We should perhaps `s/emoji/emoticon/` for technical correctness. "emoji" is complicated, and it would be unwieldy to cover them fully.

python/whylogs/core/metrics/metrics.py
```diff
@@ -40,6 +40,7 @@ class MetricConfig:
     kll_k: int = 256
     fi_lg_max_k: int = 10  # 128 entries
     fi_disabled: bool = False
```
richard-rogers, line 42: If I were me, I'd move the metric enable/disable to a separate class (`ColumnSchema`? or maybe a separate `ColumnConfig`?) and have `MetricConfig` just hold actual parameters for `Metric`s.  It also seems like these overlap with resolvers... Is there guidance on when to `resolve` vs `MetricConfig`?  Also, `UnicodeRangeMetric` tracks string lengths in addition to range counts, so it might be unclear that `track_unicode_ranges` turns the non-rangy parts of the metric on/off. There's also the case of `MetricConfig.unicode_ranges = {}`, which will still track a single `UNKNOWN` range.

python/tests/api/logger/test_logger.py
jamie256: Either way is better than what my IDE did above.

While the MetricConfig live in whylogs.core.metrics it is also a top level config mechanism, and I would say makes sense to import alongside the profile classes.

```
from whylogs.core import MetricConfig
```

python/whylogs/core/schema.py
jamie256: Fixed

python/tests/api/logger/test_logger.py
jamie256: Agreed, done.

python/whylogs/core/metrics/metrics.py
jamie256: Good call out to hold off on enabling by default before we have some better numbers. Splitting up this change:
1) Fixes to allow enabling unicode_range tracker turned on as part of profiling (keeping this in this CR)

2) Enabling unicode_range tracker by default (rolled back for now, to limit risk pending better understanding of impact after some benchmarking).

python/whylogs/core/metrics/metrics.py
jamie256: Those refactors do sound reasonable. I vote to keep this change small and move forward with the bug fix portion here first + tests. Let's follow up with the refactor around ColumnConfig vs MetricConfig in a separate PR.

richard-rogers (approved):

PR #623, opened 2022-06-01
Log Rotation Example

TheMellyBee:
This is a work in progress for log rotation. There is also a flask integration in progress, but I wanted to get this out for the launch. Right now it's really rough cause I'm still trying to figure out how using the logger work. I'm getting streaming data from the cpu. 


TheMellyBee:
@FelipeAdachi and @jamie256 here is an update. There are two major places that need fixed before it's ready, the Google Collab won't let me add in a gist for it, if someone could show me how that would be great. Second, how do we merge from the point of reading in .bin? Can it only be the views? 

I addressed most of the feedback I believe

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,332 @@
...
+      "\u001B[0;32m~/opt/anaconda3/lib/python3.9/os.py\u001B[0m in \u001B[0;36mmakedirs\u001B[0;34m(name, mode, exist_ok)\u001B[0m\n\u001B[1;32m    223\u001B[0m             \u001B[0;32mreturn\u001B[0m\u001B[0;34m\u001B[0m\u001B[0;34m\u001B[0m\u001B[0m\n\u001B[1;32m    224\u001B[0m     \u001B[0;32mtry\u001B[0m\u001B[0;34m:\u001B[0m\u001B[0;34m\u001B[0m\u001B[0;34m\u001B[0m\u001B[0m\n\u001B[0;32m--> 225\u001B[0;31m         \u001B[0mmkdir\u001B[0m\u001B[0;34m(\u001B[0m\u001B[0mname\u001B[0m\u001B[0;34m,\u001B[0m \u001B[0mmode\u001B[0m\u001B[0;34m)\u001B[0m\u001B[0;34m\u001B[0m\u001B[0;34m\u001B[0m\u001B[0m\n\u001B[0m\u001B[1;32m    226\u001B[0m     \u001B[0;32mexcept\u001B[0m \u001B[0mOSError\u001B[0m\u001B[0;34m:\u001B[0m\u001B[0;34m\u001B[0m\u001B[0;34m\u001B[0m\u001B[0m\n\u001B[1;32m    227\u001B[0m         \u001B[0;31m# Cannot rely on checking for EEXIST, since the operating system\u001B[0m\u001B[0;34m\u001B[0m\u001B[0;34m\u001B[0m\u001B[0m\n",
+      "\u001B[0;31mFileExistsError\u001B[0m: [Errno 17] File exists: '/Users/melanie/Dev/whylogs/python/examples/advanced/example_output'"
+     ]
+    }
```
jamie256: General comment, when reviewing the notebook there are multiple screens worth of red for the error output, can we remove the outputs that aren't helpful to understanding the example?

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,332 @@
...
+      "\u001B[0;31mKeyboardInterrupt\u001B[0m: "
+     ]
+    }
+   ],
+   "source": [
+    "url = \"https://blockchain.info/ticker\"\n",
```
jamie256: While this is a neat example as a demo, polling blockchain seems a bit distracting from the topic of how to structure a rolling logger. Can we use our demo tour datasets or just two json messages emitted by code as placeholders for real inputs?

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,332 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "source": [
+    "# Rolling Logs with Streaming Data\n",
+    "[![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/whylabs/whylogs/blob/mainline/python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb)\n",
+    "\n",
+    "Now that you've become family with the [\"Getting Started\"](https://github.com/whylabs/whylogs/blob/9618e5dd6570bc484579ec1325f2f512ff56977f/python/examples/basic/Getting_Started.ipynb) and the basic examples, let's see what else whylogs can be used for! So far, you've seen it ingest rows and dataframes during the logging process, but now let's look at ways to handle large amounts of changing data such as streaming with ... rolling logs! (sometimes also called log rotation)\n",
```
FelipeAdachi: Can we point to the getting started example at mainline?

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,332 @@
...
+      "\u001B[0;32m~/opt/anaconda3/lib/python3.9/os.py\u001B[0m in \u001B[0;36mmakedirs\u001B[0;34m(name, mode, exist_ok)\u001B[0m\n\u001B[1;32m    223\u001B[0m             \u001B[0;32mreturn\u001B[0m\u001B[0;34m\u001B[0m\u001B[0;34m\u001B[0m\u001B[0m\n\u001B[1;32m    224\u001B[0m     \u001B[0;32mtry\u001B[0m\u001B[0;34m:\u001B[0m\u001B[0;34m\u001B[0m\u001B[0;34m\u001B[0m\u001B[0m\n\u001B[0;32m--> 225\u001B[0;31m         \u001B[0mmkdir\u001B[0m\u001B[0;34m(\u001B[0m\u001B[0mname\u001B[0m\u001B[0;34m,\u001B[0m \u001B[0mmode\u001B[0m\u001B[0;34m)\u001B[0m\u001B[0;34m\u001B[0m\u001B[0;34m\u001B[0m\u001B[0m\n\u001B[0m\u001B[1;32m    226\u001B[0m     \u001B[0;32mexcept\u001B[0m \u001B[0mOSError\u001B[0m\u001B[0;34m:\u001B[0m\u001B[0;34m\u001B[0m\u001B[0;34m\u001B[0m\u001B[0m\n\u001B[1;32m    227\u001B[0m         \u001B[0;31m# Cannot rely on checking for EEXIST, since the operating system\u001B[0m\u001B[0;34m\u001B[0m\u001B[0;34m\u001B[0m\u001B[0m\n",
+      "\u001B[0;31mFileExistsError\u001B[0m: [Errno 17] File exists: '/Users/melanie/Dev/whylogs/python/examples/advanced/example_output'"
+     ]
+    }
+   ],
+   "source": [
+    "import pandas as pd\n",
+    "!pip install psutil\n",
```
FelipeAdachi: can we change all occurences of "!pip install" to "%pip install"?

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,332 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "10\n"
+     ]
+    }
+   ],
+   "source": [
+    "def count_files(tmp_path):\n",
```
FelipeAdachi: I think this works only if you have the folder already created. Maybe running a mkdir first?

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,332 @@
...
+    "\n",
+    "    #  You may prefer to do a \"while true\" or until a certain time.\n",
+    "    #  log_number < 1000 is limited for this example\n",
+    "    while log_number < 1000:\n",
+    "        log_number += 1\n",
+    "        bitcoin_ticker_df = pd.read_json(url)\n",
+    "        logger.log(bitcoin_ticker_df)\n",
+    "        if log_number % 100 == 0:\n",
+    "            print(\"Log number:\" + str(log_number) +\n",
+    "              \"     Logged Files: \" + str(count_files(tmp_path)))\n",
```
jamie256: I think this is part of the key to understanding what the rolling logger does and why someone might use it:
1) Rather than "Log number" we should name this something like "inputs processed" or "message count", so that a user sees that we process zero to many inputs and periodically produce a dataset profile view.
2) rephrase "Logged Files:" to something like "dataset profiles" created and written by the "local" writer? We could encourage users to do something like merge them, but it might be better to just describe what they are and show the file names, then it might be clear that this produces a timeseries of dataset profiles that are produced every "interval"?

Merging is interesting, but can be mentioned after describing what the files are (rather than suggesting what someone can do with them?). The existing documentation for rolling logger has a minimal code snippet: https://github.com/whylabs/whylogs/blob/mainline/python/docs/migration/log_rotation.md that shows an excerpt that configures a rolling logger, but its missing description and surrounding setup, but we don't need to go much further since we have pub/sub and real app examples covered with the Flask (coming) and Kafka example notebooks.

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,332 @@
...
+      "\u001B[0;31mKeyboardInterrupt\u001B[0m: "
+     ]
+    }
+   ],
+   "source": [
+    "url = \"https://blockchain.info/ticker\"\n",
+    "\n",
+    "with why.logger(mode=\"rolling\", interval=30, when=\"S\", base_name=\"cpu_streaming_data\") as logger:\n",
```
jamie256: The "with" clause here is worth explaining further: one thing a whylogs integration in production should do is hook into context manager lifecycle (which the rolling logger does here I think), so that when an application shuts down or the logger otherwise goes out of scope, the logger triggers an "early, out of schedule" one last flush of the dataset profile (so that we don't lose the data that was profiled since the last "interval" triggered.

This might be an interesting focus not covered by other content / examples: a service might get zero to many requests, and we can generate a profile on a schedule to periodically gather telemetry for a prod service / ML model, and also we flush on app exit if you follow this pattern.

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,332 @@
...
+    "    logger.append_writer(\"local\", base_dir=tmp_path)\n",
+    "    log_number=0\n",
+    "\n",
+    "    #  You may prefer to do a \"while true\" or until a certain time.\n",
+    "    #  log_number < 1000 is limited for this example\n",
+    "    while log_number < 1000:\n",
+    "        log_number += 1\n",
+    "        bitcoin_ticker_df = pd.read_json(url)\n",
+    "        logger.log(bitcoin_ticker_df)\n",
+    "        if log_number % 100 == 0:\n",
```
FelipeAdachi: at least for me, it takes a long time to print the first time. maybe reducing from 100 to a lower number?

jamie256 (commented):
Thanks for building this out and describing the scenario. A few comments below: I think we can focus the scope a bit to help explain the core concepts in the rolling logger (since it is a bit tricky), but this is a neat scenario (maybe keep for a demo - example extension)?

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
TheMellyBee: Sure! Huh that's interesting that there are screens worth. It also shouldn't have the file error any more so I'll fix that

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
TheMellyBee: Absolutely!

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
TheMellyBee: Ok :) I haven't seen the % before! Thanks!

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
TheMellyBee: I do run mkdir up at the top when creating a the tmp_Path

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
TheMellyBee: I disagree with that as having a live data pull allows them to actually see what you might use a data logger on vs the static datasets that I was seeing. Plus in a real world example they would be pulling from something. As a user I would find this much more clear to help figure out how and why to use this than the old logger which pulled a static. 

I'm curious more on your thoughts and on @FelipeAdachi here. :)

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
FelipeAdachi: I'm sorry, it seems I pulled an outdated version. I think you're already doing that

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,332 @@
...
+    "pycharm": {
+     "name": "#%%\n"
+    }
+   }
+  },
+  {
+   "cell_type": "markdown",
+   "source": [
+    "# What's next?\n",
+    "- Check out a deeper dive into rolling logs with a flask app here. # In progress\n",
```
FelipeAdachi: Let's remove this until we actually have the flask example done?

TheMellyBee: Deal

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
TheMellyBee: Love those name suggestions! 

That description would be good. What do you think of something along these lines? 

---

Congrats! Now you've got data safely stored away, but what exactly are these .bin files? As you are logging datasets the session tracks many inputs (done through `why.log()`) into a dataset profile. When we use the rolling logger it will write out the dataset profile to the .bin then flush it to start logging again. This allows you to have that data safely stored in an incremental fashion which you can then merge back together as one piece, individually, or any number.

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
TheMellyBee: Sounds great. Does that happen automatically with the with? or does that need to be added in by the user? (I was expected that we do that on our end)

Absolutely! Do we think this is one that we should in here, or would that be better served in the flask app example or something else?

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
TheMellyBee: Sure thing! I'll try it with 25 and see how that goes

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
FelipeAdachi: This happens automatically, and I think we could add that explanation on this example!

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
FelipeAdachi: Yes, I think this example could have shown something simpler, but this is a nice use case, and I think there'd be no harm keeping this as it is. And I do agree that a dynamic source of data makes more sense for the rolling logger.

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
FelipeAdachi: Yes, I agree we could show a list of the generated binary files, and maybe just read the first file and show what's in there with a `to_pandas()`. Then we could just mention that these multiple profiles could be merged (and point to the merging example), so you are able to inspect them with different time granularities with resolution of 30s, which is the defined logging interval.

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
TheMellyBee: Ok, so today let's make sure to chat data examples to see if there is a better compromise.

jamie256: Part of my confusion in this scenario (trying to put on my "new person to whylogs hat") is around the poll-and-read-one-message from a URL structure in the example. Having an event driven flow or otherwise varying rate of input over time might better clarify why a rolling logger is well suited for a production use case? The idea of live data is also helpful, which your example shows.

Using the context of a message consumer or web service in an example might both cover the live data idea (without having to actually handle live data) as well as the expectation of varying rates of traffic, needs for scaling characteristics that datasketching is well suited for and focuses on the integration bits. We could simulate sending web traffic or message publishing, a mock function you define in the notebook to produce the data in a mocked way might be fine. In the context of a handle_request method is where the relevant whylogs integration would occur (e.g. given a message and an "ML method" whylogs logger profiles the inputs/outputs with the rolling logger initialized outside of handle_request. Having a rolling logger configured in this case enables the whylogs integration to decouple dataset profile writing from the message flow or web traffic. If two messages come in within one "interval" we get 1 dataset profile written. If we see 1000 messages next interval, still 1 dataset profile written for that interval. The output size of the dataset profiles is relatively predictable even if the traffic might be spiky.

Let's discuss further and chat data sources.

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
jamie256: Correct the base class for the Loggers implements this here:
https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/api/logger/logger.py#L79-L81

There is debug output written when this occurs on the rolling logger: https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/api/logger/rolling.py#L196

See: https://docs.python.org/3/library/contextlib.html and https://docs.python.org/3/reference/datamodel.html#context-managers

If the example is using a `with` statement it is worth mentioning why we are doing that.

jamie256:
If you run ```poetry run pre-commit run --all-files``` locally before pushing the branch pre-commit can fix up some of these issues e.g. the end-of-file-fixer that appears to be failing the whylogs-ci: https://github.com/whylabs/whylogs/runs/6910942736?check_suite_focus=true

python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb
TheMellyBee: Maybe I can have the mock push data at random intervals from 0 to 10 secs or something? Basically mimic a subscription model

TheMellyBee:
> 

Btw

python/examples/advanced/Log_Rotation_for_Streaming_Data/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,1221 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "#### Why would you want this?\n",
+    "Well, logging data throughout a given time period allows a higher grain of precision to your statistical profiles, and having these logs written regularly not only ensures their safety but also allows more options for merging profiles when it comes time for analysis. We'll go into that in depth in the [\"Merging Profiles\"](https://github.com/whylabs/whylogs/blob/9618e5dd6570bc484579ec1325f2f512ff56977f/python/examples/basic/Merging_Profiles.ipynb) notebook, but you can also see a simple example of it at the end of this notebook.\n",
```
FelipeAdachi: I think the motivation should be more related to the streaming scenario itself, by stating that during production there are several cases in which data flows in real-time, unbounded by business hours or other natural boundaries. Maybe take inspiration from the first paragraphs in this page:
https://docs.whylabs.ai/docs/usecases-streaming

python/examples/advanced/Log_Rotation_for_Streaming_Data/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,1221 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "pycharm": {
+     "name": "#%%\n"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "%pip install psutil;\n",
```
FelipeAdachi: remove this if it's not being used

python/examples/advanced/Log_Rotation_for_Streaming_Data/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,1221 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "##  Example\n",
+    "This example will be more like what you'll see in an environment. Imagine we want to see our data **every day** you'll want to have at logs **every thirty**, but you don't want to roll the logger over with only one data point, so we'd work with every 5 hours for instance.\n",
```
FelipeAdachi: Both paragraphs here are really confusing. Every thirty what? the rolling logger is every 5 hours or every hour? Either way, the code below uses 15 minutes as rolling interval, so it should be changed as well

python/examples/advanced/Log_Rotation_for_Streaming_Data/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,1221 @@
...
+   "metadata": {
+    "pycharm": {
+     "name": "#%%\n"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "class MyApp:\n",
+    "    def __init__(self):\n",
+    "        self.logger = why.logger(mode=\"rolling\", interval=15, when=\"M\", \n",
```
FelipeAdachi, line 424: here the interval is 15 min, but from the text above it should be something like 1 hour or 5 hours.

python/examples/advanced/Log_Rotation_for_Streaming_Data/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,1221 @@
...
+    "        message = ['moc_1655923486.322933.json', 'moc_1655926186.340832.json']\n",
+    "        example_path = os.path.join(\"mock_input\", random.choice(message))\n",
+    "        data_df = pd.read_json(example_path)\n",
+    "    return data_df\n",
+    "\n",
+    "\n",
+    "def test_driver():\n",
+    "    for i in range(100):\n",
+    "        data_df = data_feeder()\n",
+    "        consume(data_df)\n",
```
FelipeAdachi: should this be using `app`'s consume?

python/examples/advanced/Log_Rotation_for_Streaming_Data/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,1221 @@
...
+      "\u001B[0;32m/var/folders/l2/gclnndpj2bv1x_qyjts0c7tc0000gn/T/ipykernel_31676/1485609349.py\u001B[0m in \u001B[0;36m<cell line: 2>\u001B[0;34m()\u001B[0m\n\u001B[1;32m      1\u001B[0m \u001B[0;31m# Get the first file\u001B[0m\u001B[0;34m\u001B[0m\u001B[0;34m\u001B[0m\u001B[0m\n\u001B[0;32m----> 2\u001B[0;31m \u001B[0mfirst_file\u001B[0m \u001B[0;34m=\u001B[0m \u001B[0mlistdir\u001B[0m\u001B[0;34m(\u001B[0m\u001B[0mtmp_path\u001B[0m\u001B[0;34m)\u001B[0m\u001B[0;34m[\u001B[0m\u001B[0;36m0\u001B[0m\u001B[0;34m]\u001B[0m\u001B[0;34m\u001B[0m\u001B[0;34m\u001B[0m\u001B[0m\n\u001B[0m\u001B[1;32m      3\u001B[0m \u001B[0mprint\u001B[0m\u001B[0;34m(\u001B[0m\u001B[0mfirst_file\u001B[0m\u001B[0;34m)\u001B[0m\u001B[0;34m\u001B[0m\u001B[0;34m\u001B[0m\u001B[0m\n\u001B[1;32m      4\u001B[0m \u001B[0mpath\u001B[0m \u001B[0;34m=\u001B[0m \u001B[0mos\u001B[0m\u001B[0;34m.\u001B[0m\u001B[0mpath\u001B[0m\u001B[0;34m.\u001B[0m\u001B[0mjoin\u001B[0m\u001B[0;34m(\u001B[0m\u001B[0mtmp_path\u001B[0m\u001B[0;34m,\u001B[0m \u001B[0mfirst_file\u001B[0m\u001B[0;34m)\u001B[0m\u001B[0;34m\u001B[0m\u001B[0;34m\u001B[0m\u001B[0m\n\u001B[1;32m      5\u001B[0m \u001B[0;34m\u001B[0m\u001B[0m\n",
```
FelipeAdachi: Let's remove the errors in the output

python/examples/advanced/Log_Rotation_for_Streaming_Data/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,1221 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "## Merging Profiles from .bin\n",
+    "Ok, so we have saved .bin!! Huzzah! .... and what do we do with them?\n",
```
FelipeAdachi: I personally think we don't need the code to show merging here.
We could just say something like "Now that you have several profiles for different time intervals, you are free to merge them as you see fit, in order to inspect data at different time granularities. For more information on merging see (points to the merging notebook)." and wrap it up.

python/examples/advanced/Log_Rotation_for_Streaming_Data/Streaming_Data_with_Log_Rotation.ipynb
FelipeAdachi: We should also explain why is there a timer with random seconds, making it clear that is to reflect a production environment, where data arrives at random intervals.

FelipeAdachi (changes requested):
The changes made are great! The class to consume data really makes the use case closer to what is intended for the rolling logger. thanks, Melly!

Added some comments!

python/examples/advanced/Log_Rotation_for_Streaming_Data/Streaming_Data_with_Log_Rotation.ipynb
TheMellyBee: Absolutely thanks! Will rework

python/examples/advanced/Log_Rotation_for_Streaming_Data/Streaming_Data_with_Log_Rotation.ipynb
TheMellyBee: hmm my fault for making the above confusing. If we are analysising this every hour we would want to roll it ever 5-15 min.

python/examples/advanced/Log_Rotation_for_Streaming_Data/Streaming_Data_with_Log_Rotation.ipynb
TheMellyBee: Good point! I'll change that.

FelipeAdachi:
Nice Melly!

One last minor thing:
We don't want to have 15 seconds on the code, because if people copy the code as is, it will flood the platform with profiles.
Maybe put 15 minutes, as described in the text, and add a comment that the user can change it for testing purposes.

TheMellyBee:
Good catch, that was just to run it for the output! Forgot to move it back

FelipeAdachi (approved):

python/examples/advanced/Log_Rotation_for_Streaming_Data/Streaming_Data_with_Log_Rotation.ipynb
```diff
@@ -0,0 +1,1571 @@
...
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "# Rolling Logs with Streaming Data\n",
+    "[![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/whylabs/whylogs/blob/mainline/python/examples/advanced/Streaming_Data_with_Log_Rotation.ipynb)\n",
+    "\n",
+    "Now that you've become family with the [\"Getting Started\"](https://github.com/whylabs/whylogs/blob/mainline/python/examples/basic/Getting_Started.ipynb) and the basic examples, let's see what else whylogs can be used for! So far, you've seen it ingest rows and dataframes during the logging process, but now let's look at ways to handle large amounts of changing data such as streaming with ... rolling logs! (sometimes also called log rotation)\n",
```
jamie256, line 14: "Now that you've become family" --> "familiar"

jamie256 (approved):
Looks great!

python/examples/advanced/Log_Rotation_for_Streaming_Data/Streaming_Data_with_Log_Rotation.ipynb
jamie256: can do this in follow up.

PR #628, opened 2022-06-02
[feature] Mlflow Writer

murilommen:
## Description

This repository needs an Mlflow writer to enable users to write their `whylogs`' profile along with their training experiments.


python/tests/api/writer/test_mlflow.py
```diff
@@ -0,0 +1,61 @@
...
+        shutil.rmtree("/tmp/artifact_downloads")
+
+    @pytest.fixture
+    def mlflow_writer(self):
+        writer = MlflowWriter()
+        return writer
+
+    def test_writes_profile_to_mlflow_experiment(self, profile_view, mlflow_writer):
+        mlflow_writer.write(profile_view)
+        assert os.getcwd() == "/Users/murilomendonca/Documents/repos/whylogs"  #
```
FelipeAdachi: this will fail in CI, right?

murilommen: yes, good catch! hahah forgot to remove :(

FelipeAdachi (approved):
LGTM!

python/pyproject.toml
```diff
@@ -35,6 +35,7 @@ flake8 = { version = "4.0.1", optional = true }
 nbsphinx = { version = "^0.8.8", optional = true }
 boto3 = {version = "^1.22.13", optional = true }
 whylabs-client = {version = "^0.3.0", optional = true}
+mlflow-skinny = {version = "^1.26.1", optional = true}
```
jamie256, line 40: minor but this is in the # Doc dependencies section, maybe move it down? We could add a comment for # integrations?

python/tests/api/writer/test_mlflow.py
```diff
@@ -0,0 +1,59 @@
...
+import whylogs as why
+from whylogs.api.logger.result_set import ViewResultSet
+from whylogs.api.writer.mlflow import MlflowWriter
+
+
+class TestMlflowWriter(object):
+    @classmethod
+    def teardown_class(cls):
+        shutil.rmtree("mlruns")
+        shutil.rmtree("/tmp/artifact_downloads")
```
jamie256: why not also use ```tempfile``` in the tests?

python/whylogs/api/writer/mlflow.py
```diff
@@ -1,12 +1,43 @@
...
+        if profile_name:
+            self._profile_name = profile_name
 
-    def option(self, param: Optional[str] = None) -> "MlflowWriter":  # type: ignore
-        pass
+    def _get_temp_directory(self, run_id):
+        tmp_dir = mkdtemp()
+        output_dir = os.path.join(tmp_dir, self._profile_dir)
+        os.makedirs(output_dir, exist_ok=True)
+        output = os.path.join(output_dir, f"{self._profile_name}_{run_id}.bin")
```
jamie256, line 42: How does the naming compare to the mlflow writer in v0? does it matter if they are different?

jamie256 (commented):
Minor comment and a question.

jamie256 (approved):

python/tests/api/writer/test_mlflow.py
murilommen: the tempfile on the implementation is writing on local temporarily, then logging on to mlflow. 
i'm not sure if you can not persist a file with mlflow's `log_artifact`

murilommen: `mlruns` is generated automatically by mlflow when creating a run and `/tmp/artifact_downloads` was created to download back a profile with the Client

python/whylogs/api/writer/mlflow.py
murilommen: Since the api changes from `mlflow.whylogs.log_pandas` to `profile.writer("mlflow").write()` I didn't think about parameters naming tbh. But it's a valid question i don't have a strict answer to. 
Could change params to `output_path, path_template, filename_template` although it's not exactly clear the difference between `output_path` and `path_template` to me. And thinking through, the other writers don't exactly follow the same naming on v0 as well. Thoughts?

python/pyproject.toml
murilommen: will do!

python/tests/api/writer/test_mlflow.py
jamie256: Right, I was thinking that a utility (like tempfile) can insulate us from the complexities of file paths on different systems. We don't have it yet, but I'd like to include windows in our test matrix, so I'm trying to reduce the work of enabling that soon.

python/whylogs/api/writer/mlflow.py
jamie256: You don't need to change it, and this is a new API around mlflow in v1, but wanted to think through if the migration some from a user's perspective.

python/tests/api/writer/test_mlflow.py
murilommen: mlruns is standard for both systems, but maybe i can change the /tmp to be a directory created at the same level as mlruns. how does that sound?

PR #629, opened 2022-06-02
Restrict max python version to less than 3.10 aligned with test matrix

### LINKED ISSUE #627: Install with pipenv fails
## Description

When trying to install with pipenv whylogs fails with an error in a package "whylogs-sketching"

- System info
  - Operating system and version: Windows 10 Enterprise 19042.1706
  - whylogs v1.0.0- python 3.10.4
  - pipenv 2022.5.2
- Steps to reproduce:
  - Open up a new folder
  - Try to install in a terminal with the line: "pipenv install whylogs"

whylogs-pipenv-fail

## Suggestions

Maybe this is a windows problem. I haven't checked if it can be installed on a linux build. I can install it with just pip so maybe it's just how pipenv install things that is messing it up.

## Related

jamie256:
## Description

This repository needs to describe the max python version in a way consistent with our tests which support 3.9.

3.10 has some dependencies missing, and we can allow that after we build whylogs-datasketching for those and turn on python 3.10 in our test matrix.

mitigates #627

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -11,7 +11,7 @@ include = ["whylogs/core/proto/v0/*.py*", "whylogs/core/proto/*.py*"]
 
 [tool.poetry.dependencies]
 # core dependencies. Be REALLY mindful when touching this list
-python = ">=3.7.1, <4.0"
+python = ">=3.7.1, <3.10"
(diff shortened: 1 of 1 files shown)


andyndang (approved):
LGTM!

PR #632, opened 2022-06-03
fixing mean and stddev calculation for single-value updates

FelipeAdachi:
## Description

When updating distribution metrics with a single new value, welford algorithm is used. The algorithm expects a `new_value`, but    it was being passed instead a variance result from the already tracked data, and not the incoming value.

It was also attempting to use the algorithm when there is no value to be tracked, so changed the `else` to check for `len`==1

## Changes

## Related


andyndang (approved):
LGTM

PR #638, opened 2022-06-10
Kafka example notebook

murilommen:
## Description

Adding the example notebook that references the previous `v0` example: https://github.com/whylabs/whylogs-examples/blob/mainline/python/Kafka.ipynb


python/examples/integrations/kafka-example/Kafka_Example.ipynb
```diff
@@ -0,0 +1,832 @@
...
+    "while True:\n",
+    "    finished = True\n",
+    "    record = consumer.poll(timeout_ms=500, max_records=100, update_offsets=True)\n",
+    "    for k,v in record.items():\n",
+    "        print(f'{k} - {len(v)}')\n",
+    "        df = pd.DataFrame([row.value for row in v])\n",
+    "        if counter == 0:\n",
+    "            profile = why.log(df).profile()\n",
+    "        else:\n",
+    "            profile.track(df)\n",
```
jamie256: Can we configure a rolling logger here rather than guiding users to manage the profile themselves? Ideally users don't have to know about `profile.track()`

murilommen: see if that is better now @jamie256

jamie256 (approved):
Looks good!

Two suggestions:
1) increase the "interval" in the rolling logger so that if this code is copied it gives more sensible defaults (you can make a note that you ran with a shorter interval when testing).
2) minor: add some comments around the While True: finished semantics. Its looks good but is just tricky enough that a code comment might help

python/examples/integrations/kafka-example/Kafka_Example.ipynb
```diff
@@ -0,0 +1,764 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 60,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from functools import reduce\n",
+    "\n",
+    "merged_profile = reduce((lambda x, y: x.merge(y)), profiles_list)"
```
jamie256, line 202: nice example of merging profile views from disk!

PR #640, opened 2022-06-11
Decimal handling fixes

### LINKED ISSUE #620: pyspark.sql.types.DecimalType recognized as integral by whylogs.api.pyspark.experimental.collect_dataset_profile_view
## Description

`pyspark.sql.types.DecimalType` columns are recognized as `integral` in profile generated by `whylogs.api.pyspark.experimental.collect_dataset_profile_view`.

- System info
  - Spark Version: 3.0.3
  - PySpark Version: 3.0.3
  - Hadoop Version: 2.7
  - JDK 8
  - Python 3.7.13
  - whylogs 1.0.0

- Steps to reproduce:
  - Colab Notebook to reproduce the observation (end-to-end executable): https://gist.github.com/stiebels/d4941387be46b60d329fb98e3388d7ba

- Plain Code:
```python
from sklearn.datasets import load_wine
from pyspark.sql import SparkSession
spark = SparkSession.builder.master("local[*]").getOrCreate()

df = load_wine(as_frame=True)['data']
df_spark = spark.createDataFrame(df)
df_spark_b = df_spark.withColumn("alcohol",col("alcohol").cast("decimal(10,5)"))
dataset_profile_view = collect_dataset_profile_view(df_spark_b)
dataset_profile_view.to_pandas().transpose()['alcohol']
```

## Suggestions

`pyspark.sql.types.DecimalType` should rather be treated the same as `pyspark.sql.types.DoubleType` and show up as `fractional` in the whylogs profile.

jamie256:
## Description

Until pandas has a dtype value for Decimal to inform that the values are numeric, whylogs will allow distribution metrics on object columns (which will be empty if the column contains no numeric values). Handling of empty distribution metrics was not correct in a few places (e.g. we needed to detect when we have an empty distribution before reading quantiles from it), also constraint conditions can surface surprising results when nan is compared to a threshold, so updated some of the tests.

Fix Decimal handling:
* Allow distribution metric on object column types to cover Decimal in pandas scenario.
* Add some better test validation of output metrics
* Update constraints to work against nan values
* Add tests for empty distribution metrics

### DIFF
--- a/python/tests/core/constraints/test_metric_constraints.py
+++ b/python/tests/core/constraints/test_metric_constraints.py
@@ -73,7 +73,7 @@ def test_constraints_builder(pandas_constraint_dataframe) -> None:
 
     legs_less_than_12_constraint = MetricConstraint(
         name="legs less than 12",
-        condition=lambda x: x.max < 12,
+        condition=lambda x: not x.max >= 12,
         metric_selector=MetricsSelector(metric_name="distribution", column_name="legs"),
     )
 
@@ -83,7 +83,7 @@ def test_constraints_builder(pandas_constraint_dataframe) -> None:
     )
     no_negative_numbers = MetricConstraint(
         name="no negative numbers",
-        condition=lambda x: x.min >= 0,
+        condition=lambda x: not x.min < 0,
         metric_selector=distribution_selector,
         require_column_existence=False,
     )
--- a/python/tests/api/pyspark/experimental/test_profiler_function.py
+++ b/python/tests/api/pyspark/experimental/test_profiler_function.py
@@ -102,3 +102,12 @@ class TestPySpark(object):
         pandas_type_counts = pandas_decimals_profile_view.get_column("wine").get_metric("types")
         assert pandas_type_counts.integral.value == 0
         assert pandas_type_counts.fractional.value == 2
+
+        wine_column = dataset_profile_view.get_column("wine")
+
+        distribution_metric = wine_column.get_metric("distribution")
+        if distribution_metric is None:
+            TEST_LOGGER.info(
+                f"Could not find distribution metric on decimal series: metrics are {wine_column.get_metric_names()}"
+            )
+        assert distribution_metric is not None
--- a/python/tests/core/test_datatypes.py
+++ b/python/tests/core/test_datatypes.py
@@ -1,3 +1,5 @@
+from decimal import Decimal
+
 import numpy as np
 import pandas as pd
 import pytest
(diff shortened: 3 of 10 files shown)


python/tests/core/constraints/test_metric_constraints.py
```diff
@@ -73,7 +73,7 @@ def metric_resolver(profile_view) -> List[Metric]:
 
     legs_less_than_12_constraint = MetricConstraint(
         name="legs less than 12",
-        condition=lambda x: x.max < 12,
+        condition=lambda x: not x.max >= 12,
```
jamie256, line 76: Make the condition less surprising when x.max is nan, because: 
```nan < 12``` evaluates to False

Comment (richard-rogers) [reply python/tests/core/constraints/test_metric_constraints.py:76]: Why shouldn't `NaN < 12` be false?

Comment (andyndang) [reply python/tests/core/constraints/test_metric_constraints.py:76]: FYI  `NaN > 12` is also False 💀

File: python/tests/core/constraints/test_metric_constraints.py [hunk 2]
```diff
@@ -73,7 +73,7 @@ def metric_resolver(profile_view) -> List[Metric]:
 
     legs_less_than_12_constraint = MetricConstraint(
```
FelipeAdachi, line 74: I think this is good for a temporary fix, but for the future, if it's a `nan`, shouldn't this raise an error when building the metric?

FelipeAdachi (approved):
LGTM!
minor comment added, but as something to consider in the future

python/tests/core/constraints/test_metric_constraints.py
richard-rogers: Yes, they're not really comparable. NaNs are meant to indicate a problem in a computation, so it seems reasonable that NaN should be treated as a constraint failure.

jamie256: or NaN is an unset value in this case. The NaN here is returned by the kll datasketch:
https://github.com/apache/datasketches-cpp/blob/dfd94962367207a13870e6b3bca5e6ed5a5ca3c6/python/src/kll_wrapper.cpp#L171-L172

I think we could change the DistributionMetric to return something else in this case rather than just passing back what we get from kll sketch, but the condition update was just incidental as the tests now cover empty distribution metrics.

The "not more than 12" sounds like a reasonable way of phrasing the condition?

python/tests/core/constraints/test_metric_constraints.py
jamie256: This test is an arbitrary example of a metric constraint, I think we can follow up and remove the possibility of seeing `nan` in this metric, but I'm scoping that out of this fix.

python/tests/core/constraints/test_metric_constraints.py
richard-rogers: I'm only a pretend numerical analyst, but using NaN for that frightens me :)  Maybe `None` would be a little better?

The `not`s are horrible :) but we don't have to hold up this PR...

jamie256: I mean its an arbitrary condition here for the test, we could alternatively update the expected constraint results or test in other ways, I'm *not* saying you have to use the negation pattern, but this condition is valid.

We should handle these kinds of things (NaN and None aware comparisons against metrics) in our constraint helper methods. I'm not saying we should stop here but that its out of scope for the Decimals fix.

andyndang:
LGTM!

PR #643, opened 2022-06-13
Zero padding the differences between two sets of frequent items

FelipeAdachi:
In some cases, items (or categories) are present in one profile (target or reference) but absent in the other.

This currently causes a problem in two areas:
- `viz`'s `distribution_chart` plots: Items that are present on the reference profile but absent in the target profile are not shown in the plot
- drift calculations: `chi-squared`'s implementation ignores items that appear exclusively on the reference profile.

To address these issues, an additional padding step is included, where items that don't appear in the current profile but are present in the other one are created with an estimate of 0. This is done for both profiles.

## Description

This repository needs 

## Changes

## Related


jamie256 (approved):
Looks better, thanks for fixing.

Can we add a test as a follow up?

PR #661, opened 2022-06-27
Decoupling Notebook Profile Vizualizer 

murilommen:
## Changes

- Created extensions module, with the intention to make it the standard design pattern for the `NotebookProfileViz` in the future
- Now it is possible to generate a report by injecting a `SummaryDriftReport.report()` onto the `rendered_html` argument
- separated some functionalities onto utils
- Included `DeprecationWarning` for the `preferred_path` argument, favoring `path` (to be aligned with what we use for the `DatasetProfileView` api)
- Have a cleaner extensions Writer 
- Mimic the same `ResultSet` writers logic, to be able to write HTMLs to s3 and so on

## Obs:


murilommen:
API is in but still not 100% working, since some of the logic is still tied to `DatasetProfileView`s

araker:
@murilommen, would this make it easier to create custom visualizations and/or to customize the existing ones?

murilommen:
@araker this is the beggining for us to be able to extend the current `NotebookProfileViz` workflow. and therefore, yes, each report might implement their own logics in the future. i have also tried to use the same API as we have for `ResultSet`s (the product object from `why.log(data)`) to be able to write HTMLs to the local path, s3, mlflow (in the future) and so on..

python/whylogs/api/writer/local.py
```diff
@@ -15,16 +20,26 @@ def __init__(self, base_dir: Optional[str] = None, base_name: Optional[str] = No
         self._base_dir = base_dir
         self._base_name = base_name
 
-    def write(self, profile: DatasetProfileView, dest: Optional[str] = None) -> None:
-        if dest is None:
+    def write(
+        self,
+        file: Optional[Writable] = None,
```
richard-rogers: Changing the argument order can break client code that passes arguments in order rather than by name. It would be safer to do `def write(self, profile, dest, file, **kwargs)`

python/whylogs/api/writer/local.py
```diff
@@ -15,16 +20,26 @@ def __init__(self, base_dir: Optional[str] = None, base_name: Optional[str] = No
...
+        profile: Optional[DatasetProfileView] = None,
+        dest: Optional[str] = None,
+        **kwargs,
+    ) -> None:
+        if profile:
+            logger.warning("`profile will be deprecated, you should use `file` instead")
```
richard-rogers: I think there are packages that supply decorators to deprecate function arguments.

python/whylogs/api/writer/local.py
```diff
@@ -15,16 +20,26 @@ def __init__(self, base_dir: Optional[str] = None, base_name: Optional[str] = No
...
+            file = profile
+        if isinstance(file, HTMLReport) and dest is None:
```
richard-rogers: `Writable`-specific default destination logic should live in the `Writable`

python/whylogs/api/writer/mlflow.py
```diff
@@ -1,25 +1,39 @@
...
-    def write(self, profile: DatasetProfileView) -> None:
+    def write(
+        self,
+        file: Optional[Writable] = None,
+        profile: Optional[DatasetProfileView] = None,
+        dest: Optional[str] = None,
+        **kwargs,
+    ) -> None:
+        if profile:
+            logger.warning("`profile` will be deprecated in the future, use `file` instead")
```
richard-rogers: Move this to the `Writer` base class to avoid repeating it in every subclass.

python/whylogs/api/writer/writer.py
```diff
@@ -6,7 +7,26 @@
...
+    @abstractmethod
+    def write(self, **kwargs):
+        pass
+
+    @abstractmethod
+    def option(self, **kwargs):
+        return self
+
+
+class Writer(Writable, ABC):
```
richard-rogers: Why is `Writer` inheriting from `Writable`?

python/whylogs/viz/extensions/reports/html_report.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        _rendered_html = self.report()
+        with self._safe_open_write(path) as file:
+            file.write(_rendered_html)
+
+    def option(self, **kwargs: Any):
+        return self
+
+
+class HTMLReportWriter(Writable):
```
richard-rogers: Shouldn't any `Writer` be able to write an `HTMLReport` since the report is `Writable`?

richard-rogers (changes requested):

python/whylogs/api/writer/local.py
murilommen: good point. will search on SO to implement a decorator here :)

python/whylogs/api/writer/writer.py
murilommen: forgot it there when i was trying to understand what the api should look like, my bad

python/whylogs/viz/extensions/reports/html_report.py
murilommen: not all writers, since there is no use for writing HTMLs to whylabs, for example. will try to deal with it specifically in that case

python/whylogs/api/writer/local.py
murilommen: good point. not sure it will work for DatasetProfileView though, since each writer have a minor specification to where they will store the profile. i.e.: s3 treats the paths with `prefix` and `object_name`, while local works with `base_dir` and `file_name`. but i agree it should be standard

murilommen: yea, so it's a bit harder than i thought, specially with how mlflow writer's built at the moment. in order to write the artifact, we use a helper `_get_temp_directory` method, where we have to know beforehand the "proper" path to write to a `tempfile.NamedTemporaryFile`. it is doable for most cases, but the thing is that specially for mlflow, `file.write()` expects a valid path to be there and ready.

murilommen: happy to read your thoughts on this @richard-rogers

python/whylogs/viz/extensions/reports/html_report.py
richard-rogers: In general, an `Writer` should be able to write any `Writable`. In the whylabs writer case, it makes sense for the writer to verify  the `Writable` is something whylabs can swallow. But for the most part a `Writer` should only be concerned with the destination, not the format of the data.

python/whylogs/api/writer/local.py
richard-rogers: `Writable` should have a method to return the path to the data the `Writer` should send to it's destination. Any default logic could be handled there.

richard-rogers (changes requested):

python/whylogs/api/writer/local.py
```diff
@@ -15,16 +21,25 @@ def __init__(self, base_dir: Optional[str] = None, base_name: Optional[str] = No
         self._base_dir = base_dir
         self._base_name = base_name
 
-    def write(self, profile: DatasetProfileView, dest: Optional[str] = None) -> None:
-        if dest is None:
-            dest = f"{self._base_name}_{profile.creation_timestamp}.bin"
+    @deprecated_alias(profile="file")
+    def write(
+        self,
+        file: Optional[Union[Writable, DatasetProfileView]] = None,
```
richard-rogers: `DatasetProfileView` should be a `Writable`, so this could just be `file: Writable` (or is that interacting badly with the decorator?)

python/whylogs/api/writer/local.py
```diff
@@ -15,16 +21,25 @@ def __init__(self, base_dir: Optional[str] = None, base_name: Optional[str] = No
...
+        dest: Optional[str] = None,
+        **kwargs,
+    ) -> None:
+        if isinstance(file, HTMLReport) and dest is None:
```
richard-rogers: Default `Writable` path logic should live in the `Writable`
```
source_path = file.get_path()
if source_path is None:
    dest = # writer-specific default
file.write(source_path)
# copy source_path to writer's dest
```

python/whylogs/api/writer/mlflow.py
```diff
@@ -14,12 +20,19 @@ def __init__(self) -> None:
...
+        file: Optional[Union[Writable, DatasetProfileView]] = None,
+        dest: Optional[str] = None,
+        **kwargs,
+    ) -> None:
         run = mlflow.active_run() or mlflow.start_run()
         self._run_id = run.info.run_id
-
-        output = self._get_temp_directory(run_id=self._run_id)
-        profile.write(path=output)
+        if isinstance(file, HTMLReport) and dest is None:
```
richard-rogers: Default `Writable` path logic should live in the `Writable`

python/whylogs/api/writer/s3.py
```diff
@@ -63,15 +66,25 @@ def __init__(
...
-            dest = f"{self.base_prefix}_{profile.creation_timestamp}.bin"
+    @deprecated_alias(profile="file")
+    def write(
+        self,
+        file: Optional[Union[Writable, DatasetProfileView]] = None,
+        dest: Optional[str] = None,
+        **kwargs,
+    ) -> None:
+        if isinstance(file, HTMLReport) and dest is None:
+            dest = "html_reports/ProfileViz.html"
```
richard-rogers: Default `Writable` path logic should live in the `Writable`

python/whylogs/api/writer/writer.py
```diff
@@ -16,7 +35,12 @@ def check_interval(self, interval_seconds: int) -> None:
         non-negative interval."""
 
     @abstractmethod
-    def write(self, profile: DatasetProfileView, dest: Optional[str] = None) -> None:
+    def write(
+        self,
+        file: Union[Writable, DatasetProfileView],
```
richard-rogers: We should make `DatasetProfileView <: Writable` so this can just be `file: Writable`

python/whylogs/viz/extensions/reports/html_report.py
```diff
@@ -0,0 +1,79 @@
...
+        _rendered_html = self.report()
+        with self._safe_open_write(path) as file:
+            file.write(_rendered_html)
+
+    def option(self, **kwargs: Any):
+        return self
+
+
+class HTMLReportWriter(Writer):
+    def __init__(self, report: HTMLReport, writer: Writer) -> None:
```
richard-rogers, line 74: I don't see why this is necessary. `HTMLReport` is `Writable`, so any `Writer` (except whylabs) should be able to handle it?

richard-rogers (changes requested):

python/whylogs/viz/enums/enums.py
```diff
@@ -0,0 +1,17 @@
...
+from enum import Enum
+
+
+@dataclass
+class PageSpec:
+    html: str
+    height: str
+
+
+class PageSpecEnum(Enum):
```
TheMellyBee, line 11: Someday I'd love to break these out a bit differently so they are less copy and paste of each other, but not needed in this iteration.

python/whylogs/viz/extensions/reports/histograms.py
```diff
@@ -0,0 +1,17 @@
+from whylogs.viz.extensions.reports.html_report import HTMLReport
+
```
TheMellyBee, line 2: Why are all of these here but passed?

python/whylogs/viz/extensions/reports/constraints.py
```diff
@@ -0,0 +1,5 @@
+from whylogs.viz.extensions.reports.html_report import HTMLReport
+
+
+class ConstraintsReport(HTMLReport):
+    pass
```
TheMellyBee, line 5: Why is this constraint here but passed?

TheMellyBee (approved):
I agree with Richard on his points and defer to his approval for submission. Otherwise, nice refactor. This will make it much easier to refactor some things like how the html files are set up at a laster time.

python/whylogs/viz/extensions/reports/histograms.py
murilommen: to be implemented in the future :)

python/whylogs/api/writer/writer.py
```diff
@@ -1,13 +1,32 @@
...
+    def _safe_open_write(path):
+        """Open `path` for writing, creating any parent directories as needed."""
+        try:
+            os.makedirs(os.path.dirname(path), exist_ok=True)
+        except FileExistsError:
+            pass
+        return open(path, "w")
+
+    @abstractmethod
+    def write(self, **kwargs):
```
richard-rogers: I would put `def write(self, path: Optional[str]=None)` here to make the `Writer <-> Writable` interface explicit.

python/whylogs/viz/extensions/reports/html_report.py
murilommen: this is analog to `ResultSetWriter`. the main difference is what `HTMLReportWriter.write` has as its argument

python/whylogs/api/writer/writer.py
murilommen: done!

python/whylogs/api/writer/local.py
```diff
@@ -1,30 +1,36 @@
...
         self._base_dir = base_dir
         self._base_name = base_name
 
-    def write(self, profile: DatasetProfileView, dest: Optional[str] = None) -> None:
-        if dest is None:
-            dest = f"{self._base_name}_{profile.creation_timestamp}.bin"
+    @deprecated_alias(profile="file")
+    def write(
+        self,
+        file: Optional[Union[Writable, DatasetProfileView]] = None,
```
richard-rogers: Perhaps just `file: Optional[Writable] = None`, though this will crash if the file is `None`?

richard-rogers: Yeah, looking at `Writer::write()` this should just be `file: Writable`

python/whylogs/api/writer/mlflow.py
```diff
@@ -1,43 +1,59 @@
...
+        self._file_name = None
         self._end_run = True
 
-    def write(self, profile: DatasetProfileView) -> None:
+    @deprecated_alias(profile="file")
+    def write(
+        self,
+        file: Writable,
+        dest: Optional[str] = None,
+        **kwargs,
```
richard-rogers: `**kwargs: Any`

python/whylogs/api/writer/s3.py
```diff
@@ -63,15 +63,20 @@ def __init__(
...
 
-    def write(self, profile: DatasetProfileView, dest: Optional[str] = None) -> None:
-        if dest is None:
-            dest = f"{self.base_prefix}_{profile.creation_timestamp}.bin"
-        if self.object_name is None:
-            self.object_name = os.path.basename(dest)
+    @deprecated_alias(profile="file")
+    def write(
+        self,
+        file: Optional[Writable] = None,
```
richard-rogers: `file: Writable`

python/whylogs/api/writer/s3.py
```diff
@@ -63,15 +63,20 @@ def __init__(
...
+        dest: Optional[str] = None,
+        **kwargs,
```
richard-rogers: `**kwargs: Any`

python/whylogs/api/writer/whylabs.py
```diff
@@ -80,21 +81,27 @@ def option(
         if api_key is not None:
             self._api_key = api_key
 
-    def write(self, profile: Union[DatasetProfileView, DatasetProfile], dataset_id: Optional[str] = None) -> Any:
+    @deprecated_alias(profile="file")
+    def write(self, file: Union[DatasetProfileView, DatasetProfile], dataset_id: Optional[str] = None, **kwargs) -> Any:
```
richard-rogers: It looks like we can make this `file: Writable` and make `DatasetProfile <: Writable

richard-rogers: Can we pass `dataset_id` through `kwargs`?

python/whylogs/api/writer/whylabs.py
```diff
@@ -80,21 +81,27 @@ def option(
...
         # check if the server supports ingesting whylogs 1.0.x profiles:
         if self._check_if_whylabs_disabled_v1_profiles():
             raise ValueError("The Whylabs writer is not yet supported on whylogs 1.0.x!")
 
+        profile_view = file.view() if isinstance(file, DatasetProfile) else file
+
+        if not isinstance(profile_view, DatasetProfileView):
+            raise BadConfigError(
```
richard-rogers: Perhaps `ValueError`?

python/whylogs/core/view/dataset_profile_view.py
```diff
@@ -84,8 +85,12 @@ def get_columns(self, col_names: Optional[List[str]] = None) -> Dict[str, Column
         else:
             return {k: self._columns.get(k) for k in self._columns}
 
-    def write(self, path: str) -> None:
+    def get_default_path(self) -> str:
+        return f"profile_{self.creation_timestamp}.bin"
+
+    def write(self, path: Optional[str] = None) -> None:
```
richard-rogers: add `**kwargs: Any`

python/whylogs/api/writer/writer.py
```diff
@@ -1,13 +1,30 @@
...
+        except FileExistsError:
+            pass
+        return open(path, "w")
+
+    @abstractmethod
+    def get_default_path(self) -> str:
+        pass
 
+    @abstractmethod
+    def write(self, path: Optional[str] = None, **kwargs):
```
richard-rogers: `**kwargs: Any`

python/whylogs/viz/extensions/reports/html_report.py
```diff
@@ -0,0 +1,83 @@
...
+        iframe = f"""<div></div><iframe srcdoc="{html.escape(template)}" width=100% height={self.height}
+        frameBorder=0></iframe>"""
+        display = HTML(iframe)
+        return display
+
+    @abstractmethod
+    def report(self) -> HTML:
+        pass
+
+    def write(self, path: Optional[str] = None, **kwargs) -> None:
```
richard-rogers: `**kwargs: Any`

python/whylogs/viz/extensions/reports/html_report.py
richard-rogers: If the arguments need to be different, it probably should not inherit from `Writer`

richard-rogers (changes requested):

python/whylogs/api/writer/whylabs.py
murilommen: got it :)

python/whylogs/api/writer/local.py
```diff
@@ -1,30 +1,35 @@
...
+    def write(
+        self,
+        file: Optional[Writable] = None,
+        dest: Optional[str] = None,
+        **kwargs,
+    ) -> None:
+        dest = dest or self._base_name or file.get_default_path()  # type: ignore
         full_path = os.path.join(self._base_dir, dest)
-        profile.write(full_path)
+        file.write(full_path)  # type: ignore
```
richard-rogers, line 28: `file` should not be `Optional`

python/whylogs/api/writer/local.py
```diff
@@ -1,30 +1,35 @@
...
 
-    def write(self, profile: DatasetProfileView, dest: Optional[str] = None) -> None:
-        if dest is None:
-            dest = f"{self._base_name}_{profile.creation_timestamp}.bin"
+    @deprecated_alias(profile="file")
+    def write(
+        self,
+        file: Optional[Writable] = None,
+        dest: Optional[str] = None,
+        **kwargs,
```
richard-rogers: `**kwargs: Any`

python/whylogs/api/writer/whylabs.py
```diff
@@ -79,22 +81,23 @@ def option(
...
-    def write(
-        self,
-        profile: Union[DatasetProfileView, DatasetProfile],
-        dataset_id: Optional[str] = None,
-        **kwargs,
-    ) -> Any:
-        if dataset_id is not None:
-            self._dataset_id = dataset_id
+    @deprecated_alias(profile="file")
+    def write(self, file: Writable, **kwargs: Any) -> requests.Response:
```
richard-rogers: `-> None`

python/whylogs/api/writer/whylabs.py
```diff
@@ -79,22 +81,23 @@ def option(
...
+            self._dataset_id = kwargs.get("dataset_id")
 
         with tempfile.NamedTemporaryFile() as tmp_file:
             profile_view.write(path=tmp_file.name)
             tmp_file.flush()
 
-            dataset_timestamp = profile.dataset_timestamp or datetime.datetime.now(datetime.timezone.utc)
+            dataset_timestamp = profile_view.dataset_timestamp or datetime.datetime.now(datetime.timezone.utc)
             dataset_timestamp = int(dataset_timestamp.timestamp() * 1000)
             return self._upload_whylabs(dataset_timestamp=dataset_timestamp, profile_path=tmp_file.name)
```
richard-rogers: Maybe add a `get_response()` method if this is necessary?

python/whylogs/api/writer/writer.py
```diff
@@ -1,13 +1,30 @@
...
+        except FileExistsError:
+            pass
+        return open(path, "w")
+
+    @abstractmethod
+    def get_default_path(self) -> str:
+        pass
 
+    @abstractmethod
+    def write(self, path: Optional[str] = None, **kwargs: Any):
```
richard-rogers: `-> None` ?

python/whylogs/core/dataset_profile.py
```diff
@@ -140,12 +143,15 @@ def flush(self) -> None:
...
-        else:
-            output_path = path_or_base_dir
+    @staticmethod
+    def get_default_path(path) -> str:
+        if not path.endswith("bin"):
+            path = os.path.join(path, f"profile.{int(round(time.time() * 1000))}.bin")
+        return path
 
+    @deprecated_alias(path_or_base_dir="path")
+    def write(self, path: str) -> None:
```
richard-rogers: `path: Optional[str], **kwargs: Any) -> None`

richard-rogers (changes requested):

python/whylogs/api/writer/whylabs.py
murilommen: actually i didnt have to return it in the `write` method in the first place. good catch!

murilommen:
@richard-rogers thanks for the comments there. i really wish that mypy on pre-commit could catch these ABC signature mismatches somehow.

python/whylogs/api/writer/s3.py
```diff
@@ -63,15 +63,20 @@ def __init__(
         self.bucket_name = bucket_name or None
```
richard-rogers: These will set `bucket_name` and `object_name` to `None` if they're passed the empty string. Not sure if that's the intended behavior?

richard-rogers (changes requested):

richard-rogers:
> @richard-rogers thanks for the comments there. i really wish that mypy on pre-commit could catch these ABC signature mismatches somehow.



> @richard-rogers thanks for the comments there. i really wish that mypy on pre-commit could catch these ABC signature mismatches somehow.

I think the `@override` decorator will catch a lot of those issues.

python/whylogs/api/writer/s3.py
murilommen: added in tests that cover these two cases :)

python/whylogs/viz/notebook_profile_viz.py
```diff
@@ -275,18 +241,19 @@ def summary_drift_report(self, cell_height: str = None) -> HTML:
             visualization.summary_drift_report()
 
         """
+        if not self._target_view and not self._ref_view:
```
richard-rogers: Should this be `or` instead of `and` ?  `if not (self._target_view and self._ref_view):` would match the error message.

richard-rogers (dismissed):

python/whylogs/viz/__init__.py
```diff
@@ -1,8 +1,8 @@
 import IPython  # type: ignore  # noqa  # import ipython to make
 
-from whylogs.viz.jupyter_notebook_viz import NotebookProfileVisualizer
```
jamie256, line 3: Will this break integrations that reference `whylogs.viz.jupyter_notebook_viz` ? NotebookProfileVisualizer should be the only thing integrations reference and should use whylogs.viz so maybe it is ok, but we might consider putting a deprecation warning there.

jamie256 (approved):
LGTM!

python/whylogs/viz/__init__.py
jamie256: discussed, and usage should be using whylogs.viz which remains the same, risk is acceptable.

PR #662, opened 2022-06-28
CI Notebook testing

FelipeAdachi:
## Description

- Added `test-notebook` workflow for notebook testing
- Added make command for testing notebooks
- Added `nbconvert` as dev dependency

## Changes

## Related


.github/workflows/test-notebook.yml
```diff
@@ -0,0 +1,94 @@
+name: notebook test
```
FelipeAdachi, line 1: This could be integrated into `whylogs-ci.yml` as extra steps in the `pithon-ci` job. Not sure which approach is best. Thoughts?

jamie256: I think this is better as standalone since it takes longer to run. I don't think we need to run it over the test matrix of python versions for example.

jamie256 (approved):
Thanks for picking this up, LGTM!

PR #675, opened 2022-07-06
Make Profile/Profile Views retain the same order of columns as original DataFrame

### LINKED ISSUE #641: NotebookProfileVisualizer summary drift report rerun changes
## Description
Filing issue based on feedback:
"If I rerun the notebook, the summary drift report displays features in a different order each time I run it. This is disconcerting."

Let's make the feature order and rendering stable.

FelipeAdachi:
When logging a dataframe, order of the columns was not retained in profile/profile_view.

The problem is that `frozenset` doesn't retain order.

- Changed `schema's` `get_col_names` to return a tuple instead of frozenset (to ensure order and immutability)
- Replaced `difference()` (that only works with sets) with logic that works with tuples
- Added test to assert that order is retained

## Description

This repository needs 

## Changes

## Related

### DIFF
--- a/python/whylogs/core/dataset_profile.py
+++ b/python/whylogs/core/dataset_profile.py
@@ -108,7 +108,9 @@ class DatasetProfile(object):
         # TODO: do this less frequently when operating at row level
         dirty = self._schema.resolve(pandas=pandas, row=row)
         if dirty:
-            new_cols = self._schema.get_col_names().difference(self._columns.keys())
+            schema_col_keys = self._schema.get_col_names()
+            col_keys = tuple(self._columns.keys())
+            new_cols = (col for col in schema_col_keys if col not in col_keys)
             for col in new_cols:
                 col_schema = self._schema.get(col)
                 if col_schema:
--- a/python/whylogs/core/dataset_profile.py
+++ b/python/whylogs/core/dataset_profile.py
@@ -109,8 +109,8 @@ class DatasetProfile(object):
         dirty = self._schema.resolve(pandas=pandas, row=row)
         if dirty:
             schema_col_keys = self._schema.get_col_names()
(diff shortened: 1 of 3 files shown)


TheMellyBee:
One thing I'd recommend is to go with an OrderedDict from collections (which is almost the same as the default dict 3.7+ but good to use from collections for capability). This will maintain the O(1) look up we had with the set which will make your list comprehension more efficient while maintaining the order.  

Now that does remove the immutability which I'm glad you pointed out. In this case since we are getting the names of something are we super worried about getting the immutability? If so then Tuples would be best way to go. 

Personally, it's a pretty common practice to return a list of keys which is mutable. I personally think the api would be in line with others. What do you think?

python/whylogs/core/dataset_profile.py
```diff
@@ -108,7 +108,9 @@ def _do_track(
         # TODO: do this less frequently when operating at row level
         dirty = self._schema.resolve(pandas=pandas, row=row)
         if dirty:
-            new_cols = self._schema.get_col_names().difference(self._columns.keys())
+            schema_col_keys = self._schema.get_col_names()
+            col_keys = tuple(self._columns.keys())
```
TheMellyBee: Actually instead of changing the types, you can change up lines 112 and 113 to be able to use a hash more optimally for the difference using by using this:

`new_cols = (col for col in schema_col_keys if col not in self._columns)`

TheMellyBee (changes requested):

python/whylogs/core/dataset_profile.py
FelipeAdachi: yes, that's much better!
Thanks, Melly!
changed in the last commit!

TheMellyBee (approved):
Looks good! :D

PR #677, opened 2022-07-06
Add options to WhylabsWriter to unblock use with rolling logger

### LINKED ISSUE #676: Whylabs writer breaks when dest parameter is passed to write by the rolling logger
## Description

While the dest parameter is intended for a filename which is not used by WhylabsWriter it is passed by the rolling logger which then raises an error.

## Mitigation
Allow other parameters to WhylabsWriter to unblock.

jamie256:
## Description

Add options to WhylabsWriter to unblock use with rolling logger, which passes a dest parameter and breaks WhylabsWriter's write call.

### DIFF
--- a/python/tests/api/writer/test_whylabs.py
+++ b/python/tests/api/writer/test_whylabs.py
@@ -76,3 +76,10 @@ class TestWhylabsWriter(object):
         os.environ["WHYLABS_API_KEY"] = "0123456789.any"
         with pytest.raises(ValueError):
             results.writer("whylabs").option(org_id="org_id", api_key="api_key_123.foo").write(dataset_id="dataset_id")
+
+    def test_writer_accepts_dest_param(self, results, caplog):
+        self.responses = responses.RequestsMock()
+        self.responses.start()
+        self.responses.add(PUT, url="https://api.whylabsapp.com", body=results.view().to_pandas().to_json())
+        with pytest.raises(ValueError):
+            results.writer("whylabs").option(api_key="bad_key_format").write(dataset_id="dataset_id", dest="tmp")
--- a/python/tests/api/writer/test_whylabs.py
+++ b/python/tests/api/writer/test_whylabs.py
@@ -78,8 +78,6 @@ class TestWhylabsWriter(object):
             results.writer("whylabs").option(org_id="org_id", api_key="api_key_123.foo").write(dataset_id="dataset_id")
 
     def test_writer_accepts_dest_param(self, results, caplog):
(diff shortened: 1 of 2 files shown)


python/whylogs/api/writer/whylabs.py
```diff
@@ -79,7 +79,12 @@ def option(
         if api_key is not None:
             self._api_key = api_key
 
-    def write(self, profile: Union[DatasetProfileView, DatasetProfile], dataset_id: Optional[str] = None) -> Any:
+    def write(
+        self,
+        profile: Union[DatasetProfileView, DatasetProfile],
+        dataset_id: Optional[str] = None,
+        **options,
```
jamie256: rolling logger passes in unused 'dest' parameter to write()

python/tests/api/writer/test_whylabs.py
```diff
@@ -76,3 +76,10 @@ def test_api_key_prefers_parameter_over_env_var(self, results, caplog):
         os.environ["WHYLABS_API_KEY"] = "0123456789.any"
         with pytest.raises(ValueError):
             results.writer("whylabs").option(org_id="org_id", api_key="api_key_123.foo").write(dataset_id="dataset_id")
+
+    def test_writer_accepts_dest_param(self, results, caplog):
+        self.responses = responses.RequestsMock()
+        self.responses.start()
+        self.responses.add(PUT, url="https://api.whylabsapp.com", body=results.view().to_pandas().to_json())
+        with pytest.raises(ValueError):
+            results.writer("whylabs").option(api_key="bad_key_format").write(dataset_id="dataset_id", dest="tmp")
```
murilommen, line 83: it prob raises the `ValueError` with the api_key, right? wouldnt it be best to have an `assert true` happening with `dest`? or maybe catch the ValueError log to make sure it's "dest unrelated"

jamie256: That is correct, the parameter mismatch that occurs without the change to the WhylabsWriter fails before entering the write method. This test expects an api key format `ValueError`. Without my whylabs writer changes this test causes a
 `TypeError` failing to call write()

I was looking for a cheap way to verify we enter the write() call, but don't actually make a network call to whylabs. We can follow up and fix the mocks a bit more to make sure this test is better focused.

Note that this change is being more permissive than allowing `dest`, any extra parameters are ignored.

jamie256: Added TODO for test follow up and simplified a bit.

murilommen (approved):

PR #672, opened 2022-07-05
Add WhyLabs Writer example notebook

FelipeAdachi:
- Add WhyLabsWriter example notebook
- changing both writer examples into a `writer` folder under `integrations`
- updating READMES and other markdown files in documentation

## Description

This repository needs 

The commits in this pull request will 

## Changes

- Imperative commit or change title (commit ID) 
  - Add some bullet points to explain the change
- Next commit or change (commit ID)

## Related

Relates to organization/repo#number

Closes organization/repo#number

- [ ] I have reviewed the [Guidelines for Contributing](CONTRIBUTING.md) and the [Code of Conduct](CODE_OF_CONDUCT.md).

PR #685, opened 2022-07-12
Flask Integration of whylogs and WhyLabs Notebook

TheMellyBee:
## Description
Adding an example of integrating a Flask app with whylogs log rotation viewing from local writer or with WhyLabs

## Changes
- Removes mypy checks from example folder 
- ports from the v0
- changes the dataset to use IRIS from scikit-learn
- Attempted mypy fixes, but there are many # type : ignore due to how the docker file was originally set up. Some of these I think we can get rid of with some refactoring, but I wanted to get the example up

## Relasted
#672 


python/pyproject.toml
```diff
@@ -43,6 +43,7 @@ mlflow-skinny = {version = "^1.26.1", optional = true}
 # Pyspark related dependencies
 pyarrow = {version = "^8.0.0", optional = true}
 pyspark = {version = "^3.0.0", optional = true}
+Flask = "^2.1.2"
```
FelipeAdachi: will this always install Flask?
Since Flask is needed only for the example, maybe we shouldn't require in the `.toml`, and only ask for the user to install it directly on the example?

python/examples/integrations/flask_streaming/flask_with_whylogs.ipynb
```diff
@@ -0,0 +1,836 @@
...
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "# Integrating Whylogs into your Flask Flow\n",
+    "\n",
+    "[![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/whylabs/whylogs/blob/mainline/python/examples/integrations/flask_with_whylogs.ipynb)\n",
+    "\n",
+    "Now that you've gone through the basics, let's get into how to integrate whylogs into your current work flow. Today, let's look at Flask to build a web app that wil get data from IRIS and us it for a prediction. Notice that we will log both the imput received and the prediction! "
```
FelipeAdachi: __will__ get data.....and __use__ it for a prediction. Notice that we will log both the __input__

python/examples/integrations/flask_streaming/flask_with_whylogs.ipynb
```diff
@@ -0,0 +1,836 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Note: you may need to restart the kernel to use updated packages.\n"
+     ]
+    }
+   ],
+   "source": [
+    "%pip install -q pandas utils joblib scikit-learn 'whylogs[whylabs]'"
```
FelipeAdachi: maybe we could install the other packages, but let's make the `pip install whylogs[whylabs]` commented, so we don't install this every time we run the test on it (even though this will probably be skipped, since we need Flask, docker, etc.)

python/examples/integrations/flask_streaming/flask_with_whylogs.ipynb
FelipeAdachi: We can have at the start of the example a "what you'll need" paragraph:

For this example, you'll have to:
- have Docker
- (other requirements)

python/examples/integrations/flask_streaming/flask_with_whylogs.ipynb
```diff
@@ -0,0 +1,836 @@
...
+    "\n",
+    "While there you can visualize the results by following the instructions of the [\"Notebook Profile Visualizer\"](https://github.com/whylabs/whylogs/blob/9618e5dd6570bc484579ec1325f2f512ff56977f/python/examples/basic/Notebook_Profile_Visualizer.ipynb)\n",
+    ". Some more information can be found in these notebooks [\"Merging Profiles\"](https://github.com/whylabs/whylogs/blob/9618e5dd6570bc484579ec1325f2f512ff56977f/python/examples/basic/Merging_Profiles.ipynb) and [\"Streaming_Data_with_Log_Rotation\"](https://github.com/whylabs/whylogs/blob/mainline/python/examples/advanced/Log_Rotation_for_Streaming_Data/Streaming_Data_with_Log_Rotation.ipynb).\n",
+    "\n",
+    "## Using WhyLabs\n",
+    "There is another way to look at the visualization by integrating with WhyLabs. Follow the directions in \"WhyLabs Writer\" to change the .env file in this directory. Then rerun the above. This time it will populate your WhyLabs dataset to be able to see the drifts on the WhyLabs portal. \n",
```
FelipeAdachi: The first sentence makes it look like WhyLabs is just another way to visualize the profile. Maybe you could change to make it clear that the profiles can be sent to your observability dashboard at WhyLabs directly, or something like that.

jamie256:
The whylogs-ci is not passing, see https://github.com/whylabs/whylogs/runs/7293207003?check_suite_focus=true

python/examples/integrations/flask_streaming/flask_with_whylogs.ipynb
```diff
@@ -0,0 +1,836 @@
...
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "# Integrating Whylogs into your Flask Flow\n",
+    "\n",
+    "[![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/whylabs/whylogs/blob/mainline/python/examples/integrations/flask_with_whylogs.ipynb)\n",
+    "\n",
+    "Now that you've gone through the basics, let's get into how to integrate whylogs into your current work flow. Today, let's look at Flask to build a web app that wil get data from IRIS and us it for a prediction. Notice that we will log both the imput received and the prediction! "
+   ]
```
FelipeAdachi, line 23: We also don't have an overview section, where we actually explain what's going on. How whylogs and Flask interact with each other? When is data being logged, and where is the profile going to?
You could use it as reference the __overview__ section of this old blog post: https://towardsdatascience.com/deploy-and-monitor-your-ml-application-with-flask-and-whylabs-4cd1e757c94b
It could be a simpler explanation, but I think we should have it.

FelipeAdachi (commented):
Added some comments.
About the ci errors:
- do we want to run mypy in the examples folder?
- you'll want to add this example to the skip_list in the `test_notebook` code

andyndang:
Task linked: [CU-2p2d800 Flask integration](https://app.clickup.com/t/2p2d800)

python/pyproject.toml
TheMellyBee: Good point!

python/examples/integrations/flask_streaming/flask_with_whylogs.ipynb
TheMellyBee: Deal!

python/examples/integrations/flask_streaming/flask_with_whylogs.ipynb
TheMellyBee: That's a great point. :)

jamie256 (approved):
Looks great!

PR #690, opened 2022-07-13
Change KLL accuracy according to numerical/submetrics type

### LINKED ISSUE #684: Erratic behavior for ks-test drift calculation for large-sized samples.
## Description

In some cases, running __ks-test__ p-value calculation multiple times yields wildly different results for the same data. This seems to occur for large datasets.

For example, when using this [weather forecast dataset](https://storage.yandexcloud.net/yandex-research/shifts/weather/full-dataset.tar), the p-value for the feature `gfs_u_wind` when using `scipy`'s ks-test is __0.45__ between `dev_in` and `eval_in`. This is highly expected, since both datasets come from the same distribution and have 50k and 561k samples, respectively.

However, when profiling both datasets and calculating the p-values, we get erratic results.

For instance, these are the results when calculating whylogs' ks-test p-value 10 times:
```
p_Value>>> 0.0017776882037403947
p_Value>>> 0.06992275291815553
p_Value>>> 0.0012571858877319761
p_Value>>> 0.01856089314030054
p_Value>>> 0.6163121691506832
p_Value>>> 2.7349727221794016e-07
p_Value>>> 0.47810836024385617
p_Value>>> 0.737848853948613
p_Value>>> 0.6092895783720292
p_Value>>> 0.17219371613197665
```

We can see that, sometimes, drift is detected, but at other times, drift is not detected.
The expected behavior should be no drift being detected at all times the code is run.

- Steps to reproduce:

```python
import pandas as pd

# Load each data file as a pandas data frame
df_in = pd.read_csv('data/dev_in.csv')["gfs_u_wind"].to_frame() # dev_in has 129 features and 50000 rows
df_eval_in = pd.read_csv('data/eval_in.csv')["gfs_u_wind"].to_frame() # eval_in has 129 features and 561105 rows

from scipy import stats
wind_in = df_in['gfs_u_wind']
wind_eval = df_eval_in['gfs_u_wind']

stats.ks_2samp(wind_in, wind_eval) # this always gives p-value of 0.4538904896808401

from whylogs.viz import NotebookProfileVisualizer
import whylogs as why

profile_baseline = why.log(df_in).profile().view()
profile_eval = why.log(df_eval_in).profile().view()

visualization = NotebookProfileVisualizer()
visualization.set_profiles(target_profile_view=profile_eval, reference_profile_view=profile_baseline)

visualization.summary_drift_report() # for different runs, we'll see different p-values, as listed above
```

## Related

FelipeAdachi:
## Description

This PR aims to solve https://github.com/whylabs/whylogs/issues/684.

- changed kll_k from 256 to 1024, to avoid erratic behavior of ks-test drift calculation for large-sized samples
- adding customize_kll and kll_k_custom to distinguish kll accuracy between main numerical DistributionMetrics and submetrics (unicode, string_length, custom metrics)
- adding `customize_kll` flag when initializing DistributionMetrics for submetrics in the UnicodeRange Metric
## Changes

## Related

### DIFF
--- a/python/whylogs/core/metrics/metrics.py
+++ b/python/whylogs/core/metrics/metrics.py
@@ -38,7 +38,7 @@ METRIC = TypeVar("METRIC", bound="Metric")
 @dataclass(frozen=True)
 class MetricConfig:
     hll_lg_k: int = 12
-    kll_k: int = 256
+    kll_k: int = 1024
     fi_lg_max_k: int = 10  # 128 entries
     fi_disabled: bool = False
     track_unicode_ranges: bool = False
--- a/python/whylogs/core/metrics/metrics.py
+++ b/python/whylogs/core/metrics/metrics.py
@@ -336,7 +336,7 @@ class DistributionMetric(Metric):
 
     @classmethod
     def zero(cls, config: MetricConfig) -> "DistributionMetric":
-        if config.track_unicode_ranges or config.customize_kll:
+        if config.customize_kll:
             sk = ds.kll_doubles_sketch(k=config.kll_k_custom)
         else:
             sk = ds.kll_doubles_sketch(k=config.kll_k)
--- a/python/whylogs/core/metrics/metrics.py
+++ b/python/whylogs/core/metrics/metrics.py
@@ -38,12 +38,12 @@ METRIC = TypeVar("METRIC", bound="Metric")
 @dataclass(frozen=True)
 class MetricConfig:
     hll_lg_k: int = 12
-    kll_k: int = 1024
+    kll_k: int = 256
     fi_lg_max_k: int = 10  # 128 entries
     fi_disabled: bool = False
     track_unicode_ranges: bool = False
-    customize_kll: bool = False
-    kll_k_custom = 256
+    large_kll_k: bool = True
+    kll_k_large: int = 1024
     unicode_ranges: Dict[str, Tuple[int, int]] = field(
         default_factory=lambda: {
             "emoticon": (0x1F600, 0x1F64F),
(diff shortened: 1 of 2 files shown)


TheMellyBee (approved):
Aww that does explain some of that wonky behavior. Good catch.

python/whylogs/core/metrics/metrics.py
```diff
@@ -38,10 +38,12 @@
 @dataclass(frozen=True)
 class MetricConfig:
     hll_lg_k: int = 12
-    kll_k: int = 256
+    kll_k: int = 1024
     fi_lg_max_k: int = 10  # 128 entries
     fi_disabled: bool = False
     track_unicode_ranges: bool = False
+    customize_kll: bool = False
```
jamie256: The name "customize" here is a little confusing since it is just another alternate configured value, better would be to use something like `kll_k`, `kll_k_large`.

Also what about leaving the existing key and default value as is for the string handling, and create a new key for the larger value, e.g. something like `kll_k_large`  and a bool `large_kll_k`, and then later we pass in the metric config for Distribution metric?

python/whylogs/core/metrics/metrics.py
```diff
@@ -334,7 +336,10 @@ def min(self) -> float:
 
     @classmethod
     def zero(cls, config: MetricConfig) -> "DistributionMetric":
-        sk = ds.kll_doubles_sketch(k=config.kll_k)
+        if config.customize_kll:
+            sk = ds.kll_doubles_sketch(k=config.kll_k_custom)
+        else:
+            sk = ds.kll_doubles_sketch(k=config.kll_k)
```
jamie256: Keeping the names the same for the suggestion but update those if you decide to based on comment above.
```suggestion
        configured_kll_k = config.kll_k_custom if config.customize_kll else config.kll_k
        sk = ds.kll_doubles_sketch(k=configured_kll_k )
```

python/whylogs/core/metrics/metrics.py
FelipeAdachi: Changed in the latest commit!
see if that's better!

python/whylogs/core/metrics/metrics.py
FelipeAdachi: added changes in the last commit

jamie256 (approved):
LGTM! We might change the default to False, and update where this is set explicitly so that new uses of kll get the light version by default, but can also do later.

PR #693, opened 2022-07-14
Update dependency on whylogs-sketching that has kll merge fix

jamie256:
## Description
We fixed a bug in whylogs-sketching where single item distributions failed to merge with multi-item distributions due to preamble mismatch. In this change we depend on versions of whylogs-sketching after that fix to force updates to include the fix when people install whylogs.

also ran poetry update to update poetry.lock file accordingly.


FelipeAdachi (approved):
LGTM!

PR #701, opened 2022-07-19
Add python working directory for python commands in release workflow

jamie256:
## Description

Try to fix the release workflow so that the "make install" executes in the python directory for the python related release steps.


TheMellyBee (approved):
Looks good!

PR #702, opened 2022-07-19
Orders all views columns before sending to pandas

### LINKED ISSUE #648: Profile view generates datafame with no specific order for metrics
## Description

Suppose we are merging two profiles. The merged profile doesn't have necessarily the metrics in the same order as the original profiles, which is very confusing.

Take the following example:
```python
import pandas as pd
data = {
    "animal": ["cat", "hawk", "snake", "cat"],
    "legs": [4, 2, 0, 4],
    "weight": [4.3, 1.8, 1.3, 4.1],
}

df = pd.DataFrame(data)

import whylogs as why

profile_view = why.log(df).profile().view()

print("Profile view #1\n")
print(profile_view.to_pandas().columns)

merged_view = profile_view.merge(profile_view)
print("\nProfile view #2\n")
print(merged_view.to_pandas().columns)
```

which gives us the following output:

```
Profile view #1

Index(['counts/n', 'counts/null', 'types/integral', 'types/fractional',
       'types/boolean', 'types/string', 'types/object', 'cardinality/est',
       'cardinality/upper_1', 'cardinality/lower_1',
       'frequent_items/frequent_strings', 'type', 'distribution/mean',
       'distribution/stddev', 'distribution/n', 'distribution/max',
       'distribution/min', 'distribution/q_01', 'distribution/q_05',
       'distribution/q_10', 'distribution/q_25', 'distribution/median',
       'distribution/q_75', 'distribution/q_90', 'distribution/q_95',
       'distribution/q_99', 'ints/max', 'ints/min'],
      dtype='object')

Profile view #2

Index(['types/integral', 'types/fractional', 'types/boolean', 'types/string',
       'types/object', 'cardinality/est', 'cardinality/upper_1',
       'cardinality/lower_1', 'counts/n', 'counts/null',
       'frequent_items/frequent_strings', 'type', 'distribution/mean',
       'distribution/stddev', 'distribution/n', 'distribution/max',
       'distribution/min', 'distribution/q_01', 'distribution/q_05',
       'distribution/q_10', 'distribution/q_25', 'distribution/median',
       'distribution/q_75', 'distribution/q_90', 'distribution/q_95',
       'distribution/q_99', 'ints/max', 'ints/min'],
      dtype='object')
```

## Suggestions

Every profile should have the same order for metrics and columns.

## Related

Similar problem than https://github.com/whylabs/whylogs/issues/641, and possibly related.

TheMellyBee:
## Description

To keep things easy to use every profile should have the same order for metrics and columns. There is no bug for how they merge, just to make it easy to use and less confusing. 

## Changes

- sorts the dictionary by column name before sending to pandas

## Related

Closes #648

### DIFF
--- a/python/whylogs/core/view/dataset_profile_view.py
+++ b/python/whylogs/core/view/dataset_profile_view.py
@@ -225,6 +225,6 @@ class DatasetProfileView(object):
             sum_dict = col.to_summary_dict(column_metric=column_metric, cfg=cfg)
             sum_dict["column"] = col_name
             sum_dict["type"] = SummaryType.COLUMN
-            all_dicts.append(sum_dict)
+            all_dicts.append(dict(sorted(sum_dict.items())))
         df = pd.DataFrame(all_dicts)
         return df.set_index("column")
--- a/python/whylogs/core/view/dataset_profile_view.py
+++ b/python/whylogs/core/view/dataset_profile_view.py
@@ -226,7 +226,7 @@ class DatasetProfileView(Writable):
 
     def to_pandas(self, column_metric: Optional[str] = None, cfg: Optional[SummaryConfig] = None) -> pd.DataFrame:
         all_dicts = []
-        for col_name, col in self._columns.items():
+        for col_name, col in sorted(self._columns.items()):
             sum_dict = col.to_summary_dict(column_metric=column_metric, cfg=cfg)
             sum_dict["column"] = col_name
             sum_dict["type"] = SummaryType.COLUMN
--- a/python/tests/core/view/test_dataset_profile_view.py
+++ b/python/tests/core/view/test_dataset_profile_view.py
@@ -43,3 +43,43 @@ def test_order_columns() -> None:
     data_keys = d.keys()
     profile_keys = profile_view.get_columns().keys()
     assert data_keys == profile_keys
+
+
+def test_sort_columns_sending_to_pandas() -> None:
+    data = {
+        "animal": ["cat", "hawk", "snake", "cat"],
+        "legs": [4, 2, 0, 4],
+        "weight": [4.3, 1.8, 1.3, 4.1],
+    }
+
+    df = pd.DataFrame(data)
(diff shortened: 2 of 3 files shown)


jamie256 (commented):
Can you add a test that fails without this change? You can take Felipe's repro maybe.

python/whylogs/core/view/dataset_profile_view.py
```diff
@@ -230,6 +230,6 @@ def to_pandas(self, column_metric: Optional[str] = None, cfg: Optional[SummaryCo
             sum_dict = col.to_summary_dict(column_metric=column_metric, cfg=cfg)
             sum_dict["column"] = col_name
             sum_dict["type"] = SummaryType.COLUMN
-            all_dicts.append(sum_dict)
+            all_dicts.append(dict(sorted(sum_dict.items())))
```
jamie256, line 233: Looks like we are calling sort inside a loop for every column's metrics. Does it work to call sort outside the loop? 

I'm wondering if we have guarantees on consistent column ordering already? maybe we get column ordering consistency by moving the `sorted()` outside the column iteration loop?

TheMellyBee: I did try that first, but the outside is a list of dicts which didn't want to sort the inside dicts. Whereas in the loop it sorts each dict once. 

Read through the sorted docs and goggled, but couldn't get the outside one to sort the inner dicts. if you know let me know. (I mean I can do a loop that does it, but at that point I might as well do it here)

jamie256: Ahh, that makes sense, so maybe we need to sort at both locations? I'm not too worried about the performance of sorting inside the loop I was concerned it might not be sufficient. The to_pandas() callstack isn't part of the performance critical code of profiling and is a callstack that I expect to be done less frequently as part of analyzing the profiles after logging is done.

jamie256 (approved):
LGTM!

PR #710, opened 2022-07-21
Add testing python for 3.10 and restrict compatibility to less than 3.11

jamie256:
## Description
posting early to test:

Update test matrix to include python 3.10 and 3.11


TheMellyBee (approved):
Huzzah! Any particular reason for narrowing it down from 4 to 3.11? I don't think we've been testing each as they come out without testing, so this sounds right to me in general, but just checking.

PR #715, opened 2022-07-26
Making constraint helpers publicly available

murilommen:
## Description
Some helper functions were placed under
python/examples/advanced/Constraints_Suite.ipynb.
This commit makes them available through whylogs.core.constraints import


TheMellyBee (approved):
LGTM!

python/tests/core/constraints/test_helpers.py
```diff
@@ -0,0 +1,205 @@
...
+    null_percentage_below_number,
+    null_values_below_number,
+    quantile_between_range,
+    smaller_than_number,
+    stddev_between_range,
+)
+
+
+@pytest.fixture
+def builder(profile_view):
```
jamie256: Can we add a test against a NaN value?

python/whylogs/core/constraints/helpers.py
```diff
@@ -0,0 +1,252 @@
...
+    ----------
+    column_name : str
+        Column the constraint is applied to
+    number : float
+        reference value for applying the constraint
+
+    """
+    constraint = MetricConstraint(
+        name=f"{column_name} greater than number {number}",
+        condition=lambda x: x.min >= number,
```
jamie256: Suggest defining a function for the conditions over using lambdas for readability and ability to add some checking: e.g. a trace/callstack will be easier to read when the function is named and we can check the input is not NaN for example, and then do the comparison.

python/tests/core/constraints/test_helpers.py
```diff
@@ -0,0 +1,205 @@
...
+    assert not constraint.validate()
+    assert constraint.report() == [
+        ("legs 0.9-th quantile value between 0.0 and 4.0 (inclusive)", 1, 0),
+        ("legs 0.5-th quantile value between 0.0 and 1.2 (inclusive)", 0, 1),
+        ("animal 0.5-th quantile value between 0.0 and 5.0 (inclusive)", 0, 1),
+    ]
+
+
+def test_quantile_between_range_fails_on_missing_metrics(builder):
+    pass
```
jamie256: remove?

jamie256 (commented):
Looks close. Wondering if we can come up with a better name for the file than "helpers.py"? Not sure if any of these are better "presets", "conditions", "factories"?

python/examples/advanced/Constraints_Suite.ipynb
```diff
@@ -82,143 +82,23 @@
...
-    "\n",
-    "    constraint = MetricConstraint(\n",
-    "    name=\"{} {}-th quantile value  between {} and {} (inclusive)\".format(column_name, quantile, lower, upper),\n",
-    "    condition = quantile_in_range,\n",
-    "    metric_selector=MetricsSelector(column_name=column_name, metric_name=\"distribution\"),\n",
-    "    )\n",
-    "    return constraint\n",
-    "\n"
+    "from whylogs.core.constraints import ConstraintsBuilder\n",
+    "from whylogs.core.constraints import (\n",
```
FelipeAdachi: shouldn't this be imported from .helpers, instead of directly from constraints?

python/whylogs/core/constraints/helpers.py
```diff
@@ -0,0 +1,252 @@
+from typing import Union
```
FelipeAdachi: I wonder if it would be better to have more namespaces to help distinguish between the constraints?
In the future, we'll also have Dataset level constraints and other types. It will be confusing to have them all under `helpers` (or other name, if we're going to change it)

FelipeAdachi: But even the currently ones could be distinguished by the metric it uses.

python/tests/core/constraints/test_helpers.py
murilommen: you mean having a profile_view column with only NaNs and running all tests against it? or just insert some NaNs in the column itself

python/whylogs/core/constraints/helpers.py
murilommen: right!

python/whylogs/core/constraints/helpers.py
murilommen: what is your suggestion? is 
```
factories/metric_constraints.py 
factories/dataset_constraints.py
...
```
enough?

python/examples/advanced/Constraints_Suite.ipynb
murilommen: i have made it available through `whylogs.core.constraints.__init__.py`, trying to simplify its usage. do you suggest something else?

python/whylogs/core/constraints/helpers.py
murilommen: firstly i had created different namespaces for each metric, but i ended up having to repeat all the imports in the modules and fixtures to create the very similar tests.

python/examples/advanced/Constraints_Suite.ipynb
FelipeAdachi: In the other cells, it's being imported from `.helpers`, so at least we should standardize this.
I personally prefer being imported from `helpers`, to distinguish from the core Constraints classes, like the Builder, but that's up to you!

python/whylogs/core/constraints/helpers.py
FelipeAdachi: Yes, I don't have a clear suggestion yet.
Maybe separating by each metric is too much granularity?

Right now, some come to mind:
- metric_constraints.py
   - multiple_metrics_constraint (this could live with the metrics_constraints, I guess)
- multiple_columns_constraints.py
- multiple_profile_constraints (some constraints need a baseline, like constraints for ks_test and chi_squared. whether this should be, in fact, a constraint, is also a relevant discussion)
- dataset_constraints.py

But since we have only metric_constraints, we could create a namespace for it and make the other ones as we implement them?

python/tests/core/constraints/test_helpers.py
murilommen: something like:
```python
@pytest.fixture
def nan_builder():
    df = pd.DataFrame([np.nan, np.nan], columns=["a"])
    profile_view = why.log(pandas=df).view()
    nan_builder = ConstraintsBuilder(dataset_profile_view=profile_view)
    return nan_builder
 
 def test_greater_than_number_with_nan(nan_builder):
    nan_builder.add_constraint(greater_than_number(column_name="a", number=2.2))
    constraint = nan_builder.build()

    assert constraint.validate() is False
```

murilommen: otherwise the only difference would be to always use the `weight` column, since it contains a NaN ;)

python/whylogs/core/constraints/helpers.py
murilommen: yes, and they should all live either on 
whylogs.core.constraints (the actual classes that define them)
or
whylogs.core.constraints.factories (the factories with helper functions to create a constraints suite)

python/tests/core/constraints/test_helpers.py
jamie256: Right, your nan_builder looks good, as long as there is one NaN value in our test input considering we had seen pitfalls around that scenario. Maybe one bad row, or an empty dataset kind of thing is a realistic scenario if the inputs come in batches over time, we can test that the constraints don't fail when there is a NaN value in there.

The other idea was to profile an empty numeric column so that the DistributionMetric contained NaN for the max/min  so we can show that the constraint condition wasn't surprising in the results. I'm expected we want to skip failing when evaluating and empty distribution metric with respect to a `greater_than_number ` constraint condition?

python/whylogs/core/constraints/helpers.py
FelipeAdachi: I think they should live under `factories` and leave `core.constraints` for the "building blocks".

python/tests/core/constraints/test_helpers.py
murilommen: for the second idea, we can make that optional, as @FelipeAdachi did for `quantile_between_range`. should that be the case for all  metrics then?

jamie256: its normal to have empty batches of data in some window of time if we are considering streaming scenarios or segments, and I don't think of those cases where there is no data as violating a range based constraint. Having an optional parameter is fine to control skipping NaN.

python/whylogs/core/constraints/factories/cardinality_metrics.py
```diff
@@ -0,0 +1,30 @@
...
+    column_name : str
+        Column the constraint is applied to
+    lower : int
+        Lower bound of defined range
+    upper : int
+        Upper bound of the value range
+    """
+
+    def distinct_in_range(metric: Metric):
+        distinct_estimate = metric.hll.value.get_estimate()
```
jamie256, line 22: if the hll metric is empty does this throw?

FelipeAdachi (approved):
LGTM!

python/whylogs/core/constraints/factories/cardinality_metrics.py
murilommen: @jamie256 in the tests i ran, all column types will lead to existing hll metrics. talking to felipe we were not sure whether this was correct or not

murilommen: image
this was from a 10 elements np.float type pandas DataFrame.

PR #716, opened 2022-07-26
WIP Exploratory V1 Java DataType

TheMellyBee:
## Description
Looking at the dataypes of v1 in python it seems they haven't changed much since v0. After looking through the TypedData in the v0 it looks like it's pretty close. 

But please note that I don't think we are looking at the insides of strings in v1. This would need to be taken out. 

As a first look please let me know if I'm missing a v1 data type


java/core/src/main/java/com/whylogs/core/types/TypedDataConverter.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        if (data instanceof Boolean) {
+            return TypedData.booleanValue((boolean) data);
+        }
+
+        return TypedData.unknownValue();
+    }
+
+    // TODO: this get's the insides of the string which I don't believe we are doing in v1
```
TheMellyBee, line 63: Check this one out especially

jamie256 (commented):
The TypedDataConverter was a deprecated part of v0 design, we don't parse strings as other types in v0 anymore, and we don't wrap the protobuf types in our core classes, those message types should be limited to the proto message conversion methods, so I advise against starting with com.whylogs.v0.core.message.InferredType.

If you wanted to look at the typing part of whylogs v1, look at:
See: https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/core/datatypes.py
and then see how DataType is used in the schema and resolvers as a means of defining which metrics are bound to a column:
https://github.com/whylabs/whylogs/blob/e45b5001f01506ee95b5c872ca2afe2ba5a9798d/python/whylogs/core/resolvers.py#L27
https://github.com/whylabs/whylogs/blob/e45b5001f01506ee95b5c872ca2afe2ba5a9798d/python/whylogs/core/schema.py#L146

But the concept of DataType, schema and resolvers might be easier to port after you started working on the metric components and metrics first:
https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/core/metrics/metric_components.py
https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/core/metrics/metrics.py#L91

I suggest you work backwards from what you need to build the concrete DistributionMetric as a first thing to port:
https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/core/metrics/metrics.py#L202

TheMellyBee:
I'm thinking we won't need this do_match() in the Java side, but it's here for now as a place holder. These are fairly empty shells that will get more refined as I go through metrics.

PR #718, opened 2022-07-29
Fix DatasetProfile not resolving with a custom schema

### LINKED ISSUE #717: DatasetProfile is not resolving with a custom schema
## Description

Following through the docstrings on `DatasetProfile`, I have tried to create a similar example with a custom schema to test out resolvers. But when calling `profile.track(df)` it can't resolve for the column names.

- System info
  - Operating system and version:  macOS 12
- Steps to reproduce:
```python
import pandas as pd
import numpy as np
from whylogs.core import DatasetSchema, DatasetProfile


class MySchema(DatasetSchema):
    types = {"col1": str, "col2": np.int32, "col3": str}

schema = MySchema()
prof = DatasetProfile(schema=schema)
df = pd.DataFrame({"col1": ['foo'], "col2": np.array([1], dtype=np.int32), "col3": ['bar']})
prof.track(pandas=df)

```

And the strack trace:

```bash
  File "/Users/murilomendonca/Documents/repos/whylogs/python/.venv/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3553, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-12-b428995ebc7d>", line 1, in <cell line: 1>
    prof.track(pandas=df)
  File "/Users/murilomendonca/Documents/repos/whylogs/python/whylogs/core/dataset_profile.py", line 86, in track
    self._do_track(obj, pandas=pandas, row=row)
  File "/Users/murilomendonca/Documents/repos/whylogs/python/whylogs/core/dataset_profile.py", line 125, in _do_track
    self._columns[k].track_column(pandas[k])
KeyError: 'col1'
```

## Suggestions

I am not entirely sure why this isn't working, but when trying to repro the steps, I've noticed that it does not pass the ["dirty" check](https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/core/dataset_profile.py#L113-L121) on `_do_track` method, and that's probably why it can't fetch the columns from the DatasetSchema object.

>**NOTE**: Calling `why.log(df)` on the same dataframe goes well

richard-rogers:
DatasetProfile grab initial columns from the schema
to keep them in synch.

## Changes

- `DatasetProfile` copies the initial columns from the `DatasetSchema`

## Related

Closes whylabs/whylogs#717

### DIFF
--- a/python/whylogs/core/dataset_profile.py
+++ b/python/whylogs/core/dataset_profile.py
@@ -48,6 +48,8 @@ class DatasetProfile(Writable):
         self._columns: Dict[str, ColumnProfile] = dict()
         self._is_active = False
         self._track_count = 0
+        new_cols = schema.get_col_names()
+        self._initialize_new_columns(new_cols)
 
     @property
     def creation_timestamp(self) -> datetime:
@@ -113,12 +115,7 @@ class DatasetProfile(Writable):
         if dirty:
             schema_col_keys = self._schema.get_col_names()
             new_cols = (col for col in schema_col_keys if col not in self._columns)
-            for col in new_cols:
-                col_schema = self._schema.get(col)
-                if col_schema:
-                    self._columns[col] = ColumnProfile(name=col, schema=col_schema, cache_size=self._schema.cache_size)
-                else:
-                    logger.warning("Encountered a column without schema: %s", col)
+            self._initialize_new_columns(tuple(new_cols))
 
         if pandas is not None:
             for k in pandas.keys():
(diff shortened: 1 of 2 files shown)


python/whylogs/core/dataset_profile.py
```diff
@@ -48,6 +48,8 @@ def __init__(
         self._columns: Dict[str, ColumnProfile] = dict()
         self._is_active = False
         self._track_count = 0
+        new_cols = schema.get_col_names()
+        self._initialize_new_columns(new_cols)
```
murilommen, line 52: shouldn't new_cols be a tuple here too? or is schema.get_col_names doing this already

murilommen: shoot! it returns a tuple, so it's fine

murilommen (approved):

python/whylogs/core/dataset_profile.py
richard-rogers: `get_col_names()` returns a tuple.

PR #719, opened 2022-08-02
Java IntegralMetric and  Components

TheMellyBee:
## Description

Part of the longer effort to port v1 to Java. This iteration is a chunk of code that involves the Metric, Integral Metric and supporting classes along with their tests. 

This allows for Integral Metrics to be created from their metric components, update from a column of mixed data types. 

Biggest arch area to take note of is the Metric, Base Metric and Integral Metric. This allows us to have Metrics in collections, but still maintain the typing of the subclass throughout. It makes it clearer for user and writer of the metrics what their type is and what can be combined with what at anyone time. 

Notes: 
   - no registries, serializers, deserializers, to/from protobuf
   - Lots of TODOs just to not forget section for the next iteration (which is the registry) 
   - Types are very rough and will be iterated on as the registry and resolvers are done <- direct pull from python right now which isn't going to be the final version
  
## Related


java/core/src/main/java/com/whylogs/core/metrics/Metric.java
```diff
@@ -0,0 +1,48 @@
...
+import com.whylogs.core.PreProcessedColumn;
+import com.whylogs.core.SummaryConfig;
+
+import java.util.HashMap;
+
+public abstract class Metric{
+
+    private String namespace;
+
+    // TODO: needs a __post_init__ for the registry
```
richard-rogers: Maybe not--the `__post_init__()` was a bit of a kludge to cope with interaction between `@property` and `@dataclass`. It would be worth trying to do the registration in the base class ctor(s) in the Java version.

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public static String getNameSpace(Metric metric, MetricConfig config){
+        return metric.zero(config).namespace;
```
richard-rogers: This was another Python kludge. You could probably just `return this.namespace`

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public abstract <T extends Metric> T zero(MetricConfig config);
+
+    public <T extends Metric> T add(Class<T> otherMetric){
```
richard-rogers: Another weird one... the above says `add()` returns the same class as `otherMetric`, but what we want to say is that `otherMetric` and the return value are the same class as `this`. I'd probably just use `Metric`, assertions, and comments :)
Or maybe make it a ctor and assert the class of `otherMetric` is the same as `this`'.

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
```diff
@@ -0,0 +1,48 @@
...
+        return this.merge(otherMetric);
+    }
+
+    public <T extends Metric> T merge(Class<T> otherMetric){
```
richard-rogers: See above comment

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private String namespace;
+
+    // TODO: needs a __post_init__ for the registry
+
+    public static String getNameSpace(Metric metric, MetricConfig config){
+        return metric.zero(config).namespace;
+    }
+
+    public abstract <T extends Metric> T zero(MetricConfig config);
```
richard-rogers: I wonder if we could just make this a ctor? I don't think we need this to be generic.

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
```diff
@@ -0,0 +1,48 @@
...
+    public <T extends Metric> T merge(Class<T> otherMetric){
+        // TODO: Metric Components
+        return null;
+    }
+
+    // TODO: protobuf needs MetricMessage
+
+    // TODO: get_component_paths needs MetricComponents
+
+    public abstract <O> HashMap<String, O> toSummaryDict(SummaryConfig config); // TODO: this doesn't make a good api so ...
```
richard-rogers: I think we want `HashMap<String, Object>` ?

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
richard-rogers: Or it needs to be static and generic...

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
```diff
@@ -0,0 +1,48 @@
...
+    public static <T extends Metric> T from_protobuf(MetricMessage message){
+        // Todo: check that it's a Metric dataclass
+
+        HashMap<String, Class<M extends MetricComponent<?>> M> components = new HashMap<>();
+        for k, m in message.components.items():
+            components[k] = MetricComponent.from_protobuf(m);
+
+        // TODO: We will have to figure out how to do this to make sure it's not just a Metric
+        // but the sublcass
+        return null;
```
richard-rogers: ` MethodHandles.lookup().lookupClass()`

java/core/src/main/java/com/whylogs/core/metrics/MetricConfig.java
```diff
@@ -0,0 +1,52 @@
+package com.whylogs.core.metrics;
+
+import lombok.experimental.UtilityClass;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+@UtilityClass
+public class MetricConfig {
+    private final int hll_lg_k = 12;
```
richard-rogers, line 12: Can we override these defaults?

java/core/src/main/java/com/whylogs/core/metrics/components/MaxIntegralComponent.java
```diff
@@ -0,0 +1,19 @@
...
+@Data
+@EqualsAndHashCode(callSuper=false)
+public class MaxIntegralComponent extends IntegralComponent {
+    private final int type_id = 2;
+
+    public MaxIntegralComponent(Integer value) {
+        super(value);
+    }
+
+    // TODO: python has the @_id_aggregator decorator, how do we update?
```
richard-rogers: `class MetricComponentAggregator { public static <T> T aggregate(T a, T b) {raise NotImplementedException} }`
Subclass that to implement the aggregators. `MetricComponent`s can then have an aggregator member. `MetricComponent` ctor can call an aggregator factory to get a default aggregator from the registry, or subclasses can just set it.

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
TheMellyBee: hmm I agree that this won't be the correct situation, but I think I have a different way of doing it that will allow not need the assertions

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
TheMellyBee: I don't think we do either. Will update this when I update the other section of metric

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
TheMellyBee: Yes :) That was me playing with something and forgot to switch it back.

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
TheMellyBee: Ooh thanks!

java/core/src/main/java/com/whylogs/core/metrics/MetricConfig.java
TheMellyBee: The python had this as a frozen class so i thought these are final values, but if this should be changable that's fine to. As we are using this are these values that should stay constant?

java/core/src/main/java/com/whylogs/core/metrics/components/MaxIntegralComponent.java
TheMellyBee: Makes sense. That's next on my list. I'm thinking grabbing the aggregator from the registry might be good for extensibility, but it also may not be needed with what we have

java/core/src/main/java/com/whylogs/core/metrics/MetricConfig.java
richard-rogers: In Python the defaults can be overridden at construction time. They "can't" be changed after construction because they're frozen (I think the dictionary members are still mutable?)

TheMellyBee: Ahh so our use case for this is that we may want this config overridden at construction but then frozen. Got it

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
```diff
@@ -0,0 +1,99 @@
...
+        // This metric does not need the config, but others do
+        HashMap<String, Object> summary = new HashMap<>();
+        summary.put("max", this.maxComponent.getValue());
+        summary.put("min", this.minComponent.getValue());
+        return summary;
+    }
+
+    @Override
+    public IntegralMetric merge(IntegralMetric other) {
+        // QUESTION: Is there ever a case were we want to merge an Integral with say a Distribution Metric?
```
TheMellyBee: @richard-rogers @jamie256  Please review this question: Is there every a case we want to merge and Integral with say a Distribution Metric?

java/core/src/main/java/com/whylogs/core/metrics/components/MaxIntegralComponent.java
```diff
@@ -0,0 +1,41 @@
...
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+import java.util.Collection;
+
+@Getter
+@EqualsAndHashCode(callSuper=false)
+public class MaxIntegralComponent extends IntegralComponent {
+    private static final int TYPE_ID = 2;
```
TheMellyBee, line 11: I'm probably going to move this to an enum next iteration

TheMellyBee:
This chunk is ready for review. :) Please check my notes above for the two places I'd really like an extra look/question.

java/core/src/main/java/com/whylogs/core/metrics/BaseMetric.java
```diff
@@ -0,0 +1,13 @@
...
+import lombok.NonNull;
+
+public abstract class BaseMetric<TSubclass extends BaseMetric> extends Metric {
+
+    public BaseMetric(@NonNull String namespace) {
+        super(namespace);
+    }
+
+    public abstract TSubclass merge(TSubclass other);
+    public abstract TSubclass zero(MetricConfig config);
```
richard-rogers: I think `zero` needs to be static because it's used like a constructor/factory.

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
```diff
@@ -0,0 +1,99 @@
...
+
+        // QUESTION: What happens if they have two different namespaces? Putting this for now
```
richard-rogers: This is correct.

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
```diff
@@ -0,0 +1,99 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+
+@Getter
+@EqualsAndHashCode(callSuper=false)
+public class IntegralMetric extends BaseMetric<IntegralMetric> {
```
richard-rogers: This should be `IntsMetric` to match Python

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
richard-rogers: No, we always require `this.namespace == other.namespace` (namespace is isomorphic to class).

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
```diff
@@ -0,0 +1,99 @@
...
+    public static final String DEFAULT_NAMESPACE = "ints";
```
richard-rogers: Maybe just `NAMESPACE`? "Default" suggests it can be changed.

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
TheMellyBee: Got it :)

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
TheMellyBee: Ahh I went for default since you can pass in a different namespace, but I like the change to just namespace.

richard-rogers: Why are we allowing it to change namespaces? That seems dangerous :)

java/core/src/main/java/com/whylogs/core/metrics/BaseMetric.java
TheMellyBee: Moved to static, but I may look at this more with the factory in the next iteration.

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
TheMellyBee: I'm going to make the formal ask that the python moves to IntegralMetrics as 'IntsMetric' feels very unclear to me. Is it a list of ints, is it just for the int type, and what on earth does it do? Integral isn't a big change but it stays in line with our IntegralMetricComponent, doesn't sound like a list, and (at least coming from Tableau) makes sense in industry why there would be a min/max to it. :) @richard-rogers @jamie256

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
TheMellyBee: I agree depending on how we do namespaces. :) I thought I saw it had a change constructor in python but just double checking it didn't. :P So I'll change that. 

My thought had been if you're doing a namespace to do like metrics but wanted and integral metric that wasn't mergable with another IntegralMetric for some reason. Having the ability to change the namespace might allow that, which I thought might be helpful with custom. That being said that would add a whole list of interesting cases so not allowing that would be simpler and safer.

TheMellyBee: (also it didn't allow it to change after init, just give a different namespace passed in)

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
richard-rogers: I would have preferred "Integer" and "Rational" over "Integral" (we're not doing calculus ;) and "Fractional". But alas, we've shipped "IntsMetric" so I think the best we could do is deprecate it and make `IntegralMetric` an alias for it. `MinMaxMetric` might be better?

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
richard-rogers: You can always make a subclass if you want to change the namespace. I think that emphasizes that you're making a completely different metric if you change the namespace.

java/core/src/main/java/com/whylogs/core/PreProcessedColumn.java
```diff
@@ -0,0 +1,68 @@
+package com.whylogs.core;
+
+import lombok.*;
+
+import java.util.Collection;
+import java.util.List;
+
+@Getter @Setter @EqualsAndHashCode
+public class PreProcessedColumn {
```
richard-rogers: I think it's `PreprocessedColumn` in Python

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
TheMellyBee: That would be better. :) that would be lovely down the road. :)

java/core/src/main/java/com/whylogs/core/PreProcessedColumn.java
TheMellyBee: it's preprocessed_column so yes going to camel cause I thinnk PreProcessed is easier to read and still the same name.

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
TheMellyBee: Totally. This change is already in

java/core/src/main/java/com/whylogs/core/PreProcessedColumn.java
jamie256: Richard is correct on the class name, in python it is PreprocessedColumn: https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/core/preprocessing.py#L69

java/core/src/main/java/com/whylogs/core/metrics/components/MetricComponent.java
```diff
@@ -0,0 +1,57 @@
...
+        return value;
+    }
+
+    public int getTypeId() {
+        return TYPE_ID;
+    }
+
+    public MetricComponent<T> add(MetricComponent<T> other) {
+        // TODO this is where we will use the aggregators
+        return other;
```
TheMellyBee: Change to throw NotImplementedError

TheMellyBee: Merge throughout or add throughout

java/core/src/main/java/com/whylogs/core/metrics/components/MetricComponent.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+
+
+    // TODO to_protobuf
+    // TODO from_protobuf
+    // TODO: add a from_protobuf iwht registries passed in
+    public static <T extends MetricComponent> T from_protobuf(MetricComponentMessage message) {
+        // TODO: check that it's a MetricComponent dataclass
+        return null;
```
TheMellyBee: Change to throw NotImplementedError

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
```diff
@@ -0,0 +1,24 @@
...
+@EqualsAndHashCode
+@Getter @Setter
+@RequiredArgsConstructor
+public abstract class Metric{
+
+    @NonNull
+    private String namespace;
+
+    public abstract HashMap<String, Object> toSummaryDict(SummaryConfig config);
+    public abstract OperationResult columnarUpdate(PreProcessedColumn data);
```
TheMellyBee: Change the name to the correct PreprocessedColumn

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
```diff
@@ -0,0 +1,24 @@
...
+
+@EqualsAndHashCode
+@Getter @Setter
+@RequiredArgsConstructor
+public abstract class Metric{
+
+    @NonNull
+    private String namespace;
+
+    public abstract HashMap<String, Object> toSummaryDict(SummaryConfig config);
```
TheMellyBee, line 18: Can we make this a default summary config. Make an issue in python for same thing.

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
```diff
@@ -0,0 +1,24 @@
...
+    public abstract OperationResult columnarUpdate(PreProcessedColumn data);
+
+    public String getNamespace(MetricConfig config){
```
TheMellyBee: Remove MetricConfig, no parameters

jamie256: No parameter for getNamespace

java/core/src/main/java/com/whylogs/core/metrics/BaseMetric.java
```diff
@@ -0,0 +1,13 @@
+package com.whylogs.core.metrics;
+
+import lombok.NonNull;
+
+public abstract class BaseMetric<TSubclass extends BaseMetric> extends Metric {
```
TheMellyBee: MetricWrapper?

jamie256: I don't understand why we need this class and I'm confused about what problem it is solving?

Can you comment why we need this? 

The naming is confusing since Base* is actually extending Metric, maybe "WrapperMetric" name would be easier to read

TheMellyBee: Add comments to explain the indirection to not have type eraser

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
```diff
@@ -0,0 +1,92 @@
+package com.whylogs.core.metrics;
+
+import com.whylogs.core.PreProcessedColumn;
```
TheMellyBee: Change refactored name

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
jamie256: PreProcessed -> Preprocessed

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
```diff
@@ -0,0 +1,92 @@
...
+        return summary;
+    }
+
+    @Override
+    public IntegralMetric merge(IntegralMetric other) {
+        // QUESTION: Is there ever a case were we want to merge an Integral with say a Distribution Metric?
+
+        // QUESTION: What happens if they have two different namespaces? Putting this for now
+        if (!this.getNamespace().equals(other.getNamespace())) {
+            throw new IllegalArgumentException("Cannot merge IntegralMetrics with different namespaces");
```
jamie256: suggestion: include the namespaces in the error.

java/core/src/main/java/com/whylogs/core/metrics/OperationResult.java
```diff
@@ -0,0 +1,31 @@
...
+        return new OperationResult(0, failures);
+    }
+
+    public static OperationResult failed() {
+        return new OperationResult(0, 1);
+    }
+
+    public OperationResult add(OperationResult other) {
+        int new_successes = this.successes + other.getSuccesses();
+        int new_failures = this.failures + other.getFailures() ;
```
jamie256:
```suggestion
        int new_failures = this.failures + other.getFailures();
```

java/core/src/main/java/com/whylogs/core/types/DataType.java
```diff
@@ -0,0 +1,15 @@
+package com.whylogs.core.types;
```
jamie256: Maybe remove these?

jamie256 (changes requested):
Looks really close, left a few comments and questions.

TheMellyBee (commented):
Updates from live code review

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
jamie256: I agree, but also fine to leave this out of this change.

java/core/src/main/java/com/whylogs/core/metrics/BaseMetric.java
TheMellyBee: This an application of the Curiously Recursive Template Pattern ([CRTP](https://nuah.livejournal.com/328187.html?) - which collections in Java also follow this) which allows us to have f-bound polymorphism so it's down at compile time not runtime. This does give a performance gain, but my practical reason for doing this is for useability. 


- It allows for method chaining like 
  m.zero().setMax(6) 
  
  Now that's a bit of a silly example, but without this the compiler wouldn't know that m is an IntsMetric and couldn't perform this without casting. You'd have to do a check and then go. 

- In the resolver will will have a list of Metrics this allows us to not have type eraser and not depend on passing strings around
- Well just more in general a lot of the passing of type names including custom types will be handled automatically with this as it allows us to use the typing system. The strings work great for python where its not strongly typed, but I recommend us utilizing the typing in Java to our advantage

It may be as we go that this isn't as useful and need it iterate, but I strongly think it worth exploring as we go through the registry and resolver stages to make full use of the language.

andyndang:
Task linked: [CU-2n0ejuw Implement basic metric and metric components](https://app.clickup.com/t/2n0ejuw)

java/core/src/main/java/com/whylogs/core/metrics/BaseMetric.java
TheMellyBee: ^ oh also example the merge loses it's type to Metric if it's abstracted under the Metric class. Having it under the (currently named) BaseMetric is what allows us to do the merge while returning that type. 

As for the change in name - What about just AbstractMetric?

jamie256 (approved):
LGTM!

PR #724, opened 2022-08-03
Datasets Module

FelipeAdachi:
## Description

Datasets Module:
 - implementation
 - tests
 - documentation
 - example

## Changes

## Related


TheMellyBee (approved):
Nice! This looked good to me :)

python/whylogs/datasets/weather.py
```diff
@@ -0,0 +1,229 @@
...
+    For a detailed description, please use the dataset's describe() method or visit whylog's documentation website.
+    """
+
+    baseline_df: pd.DataFrame
+    inference_df: pd.DataFrame
+    inference_interval: str = "1d"
+    number_days: int = 1
+    unit: str = "D"
+    url: str = WeatherConfig.url
+    baseline_timestamp: Union[date, datetime] = date.today()
```
jamie256: Do we need timezone info here?

FelipeAdachi: Yes!
I'll have to do some refactoring, because whylogs expects datetime objects, and the default for this dataset is date, which is naive. I'll just clear hour,seconds, etc to represent dates as datetimes and set to UTC timezone so it is compatible with whylogs' other modules.

FelipeAdachi: Done in the latest commit.
everything is in datetime format now. If the user doesn't pass any timezone info, it is assumed local and then converted to UTC timezone. In the dataframe, dates always contain time zone info.

python/tests/datasets/test_weather.py
```diff
@@ -0,0 +1,120 @@
...
+    def test_existing_version(self):
+        dataset = Weather(version="in_domain")
+        assert dataset
+
+    def test_non_existing_version(self):
+        with pytest.raises(ValueError, match="Version not found in list of available versions."):
+            Weather(version="nonexisting-version")
+
+    def test_supported_interval(self, dataset):
+        dataset.set_parameters("1d")
```
murilommen, line 25: what is the assertion here?

python/whylogs/datasets/utils.py
```diff
@@ -0,0 +1,71 @@
...
+    adjusted_date = _validate_timestamp(current_date) + date_offset
+    return adjusted_date
+
+
+def _parse_interval(interval: str) -> Tuple[int, str]:
+    try:
+        result = re.findall(r"^(\d+)([DdMmHhSs])", interval)[0]
+        return (int(result[0]), result[1].upper())
+    except IndexError:
+        raise ValueError("Could not parse interval!")
```
murilommen: good idea to add `f"Could not parse interval {interval}!"`

python/tests/datasets/test_weather.py
```diff
@@ -0,0 +1,120 @@
...
+from whylogs.datasets import Weather
+from whylogs.datasets.configs import WeatherConfig
+
+dataset_config = WeatherConfig()
+
+
+class TestWeatherDataset(object):
+    @pytest.fixture
+    def dataset(self):
+        return Weather(version="in_domain")
```
murilommen, line 14: what happens if you don't specify the dataset version?

python/examples/datasets/weather.ipynb
```diff
@@ -0,0 +1,1553 @@
...
+      ],
+      "source": [
+        "print(Weather.describe())"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "note: the output was cleared as `describe()` will print a rather lengthy description."
```
murilommen: so what is the correct approach to use this method? where should users refer to its description?

murilommen (commented):
a few comments and questions. overall work looks amazing!

python/whylogs/datasets/utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def _validate_timestamp(timestamp: Union[date, datetime, pd.Timestamp, str]) -> datetime:
+    if isinstance(timestamp, pd.Timestamp):
+        return timestamp.to_pydatetime()
+    if isinstance(timestamp, str):
+        try:
+            return datetime.strptime(timestamp, "%Y-%m-%d %H:%M:%S%z")
+        except ValueError:
+            raise ValueError("Could not parse string as datetime.")
```
richard-rogers: Helpful to give the expected format in the error message?

python/tests/datasets/test_weather.py
FelipeAdachi: default version is "in_domain". The version here is actually not needed.

python/examples/datasets/weather.ipynb
FelipeAdachi: This is the correct method. I cleared the output because it would pollute the notebook, but the user should use this method if he wants to get the full description of the dataset.

Is the note confusing? Maybe I should remove it?

python/tests/datasets/test_weather.py
FelipeAdachi: added it in the last commit

python/whylogs/datasets/utils.py
FelipeAdachi: Thanks!
Added this in the last commit!

python/whylogs/datasets/utils.py
FelipeAdachi: Thanks!
Added this in the last commit!

python/examples/datasets/weather.ipynb
FelipeAdachi: I truncated the message in the last commit.

jamie256 (approved):
LGTM!

PR #725, opened 2022-08-03
Add callback for rolling logger profile writes

### LINKED ISSUE #237: Support callback hook when log rotation happens
## Use case

In the context of Kafka, user wants to be able to trigger a Kafka commit when log rotation happens. This is to ensure that in case of failures, the system can "playback" the messages without losing fidelity.

## API suggestion
This is passed as a configuration in the writer/logger as a Python/Java callback.

## Remaining questions

What should be the parameters for the callback function?

jamie256:
## Description

the rolling logger needs an optional callback so that integrations have a hook for when the rolling logger writes out dataset profile views.

## Changes

- Adds a callback on the TimedRollingLogger passing the Writer, DatasetProfileView  and file_name (as str)
- Add some tests for this callback

## Related

Closes #237

### DIFF
--- a/python/tests/api/logger/test_rolling.py
+++ b/python/tests/api/logger/test_rolling.py
@@ -4,6 +4,7 @@ import time
 from os import listdir
 from os.path import isfile
 from typing import Any
+from unittest.mock import MagicMock
 
 import pandas as pd
 import pytest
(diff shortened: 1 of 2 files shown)


TheMellyBee (approved):
This makes sense and looks good, but the callback name vague to me in a way that lacks clarity. That being said, I'm struggling to think of a better descriptive name so... 🤷 

Otherwise looks good! Thanks for the comments in the test.

PR #731, opened 2022-08-05
Add some missing properties on metrics to cover things in summaries

### LINKED ISSUE #695: Missing properties on Distribution metric
## Description

The metric summaries include keys that are not available as a properties, and sometimes with different names: avg vs mean.

We should add properties to the metrics to allow for easier access to the keys that show up in a summary via to_pandas(), which will also make writing constraint conditions easier.

jamie256:
## Description

Metrics like DistributionMetric contain keys in their summaries such as "median" but these do not exist as properties. We got feedback that these metrics would be easier to use if they consistently had properties to cover these values.

## Changes

- Added properties where there was a property in the summary that was not already there in the metric.
- Where a metric component has the name, we cannot create a property because of name colisions e.g. "mean" (component) vs "avg" property.

### DIFF
--- a/python/whylogs/core/metrics/metrics.py
+++ b/python/whylogs/core/metrics/metrics.py
@@ -4,7 +4,7 @@ import statistics
 import sys
 from abc import ABC, abstractmethod
 from dataclasses import asdict, dataclass, field
-from typing import Any, Dict, List, Tuple, Type, TypeVar, Union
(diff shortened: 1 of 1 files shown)


TheMellyBee (approved):
Looks good, and that makes sense why that would be easier to use :)

PR #739, opened 2022-08-10
Update README.md

jamie256:
Fix broken examples link to python/examples

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -220,7 +220,7 @@ whylogs supports both structured and unstructured data, specifically:
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


FelipeAdachi (commented):
while you're at it, can you remove line 225? Binder is being used in v0 only

TheMellyBee (approved):
LGTM

jamie256:
> while you're at it, can you remove line 225? Binder is being used in v0 only

Lets clean up that section in a follow up, because there are a number of links there all pointing to whylogs-examples v0 content.

PR #745, opened 2022-08-11
Enabling whylogs Reader

murilommen:
## Description

We need an API to enable reading profiles from different locations, the same way we already do with the `whylogs.api.writer` package.

## Changes

- Added `whylogs.api.reader` package
- Included a `reader()` method on whylogs' `__init__.py`

The way to read profiles shall be either with both approaches as described below:

```python
# Option 1
reader = LocalReader()
view_result = reader.read(path="some/path.bin")

# Option 2
view_result = why.reader("s3").option(bucket_name="MY BUCKET", object_name="my/object.bin").read()
```
Both of these approaches shall return a `ViewResultSet` object, which can then be turned into a `DatasetProfile` or `DatasetProfileView`.


TheMellyBee (approved):
Looks good. :) Was easy to read too

PR #747, opened 2022-08-12
Fix subclassing DatasetSchema

richard-rogers:
## Description

Make subclasses pay attention to changed default member values.

## Changes

- Make resolvers, type_mapper, default_config use default_factory so their defaults can be overridden in subclasses

### DIFF
--- a/python/whylogs/core/schema.py
+++ b/python/whylogs/core/schema.py
@@ -39,20 +39,27 @@ class DatasetSchema:
     --------
     >>> import pandas as pd
     >>> import numpy as np
+    >>> from dataclasses import dataclass, field
+    >>> from typing import Any, Dict
     >>> from whylogs.core import DatasetSchema, DatasetProfile
+    >>> from whylogs.core.resolvers import Resolver, StandardResolver
     >>>
+    >>> class MyResolver(StandardResolver):
+    >>>    pass
+    >>>
+    >>> @dataclass
     >>> class MySchema(DatasetSchema):
-    ...    types = {
-    ...        "col1": str,
-    ...        "col2": np.int32,
-    ...        "col3": pd.Categorical(categories=('foo', 'bar'), ordered=True)
-    ...    }
-    ...    cache_size = 12
+    ...     types: Dict[str, Any] = field(default_factory=lambda: {
+    ...         "col1": str,
+    ...         "col2": np.int32,
+    ...         "col3": pd.CategoricalDtype(categories=('foo', 'bar'), ordered=True)
+    ...     })
+    ...     resolvers: Resolver = field(default_factory=MyResolver)
+    ...     cache_size: int = 12
     >>> schema = MySchema()
     >>> prof = DatasetProfile(MySchema())
     >>> df = pd.DataFrame({"col1": ['foo'], "col2": np.array([1], dtype=np.int32), "col3": ['bar']})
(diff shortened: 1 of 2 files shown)


FelipeAdachi (commented):
I tested on my previous example, and it's still getting the resolver from the base class.

For reproducibility, the code snippet is:

```python
from whylogs.core.resolvers import Resolver
from whylogs.core.metrics import StandardMetric
from whylogs.core.schema import ColumnSchema, DatasetSchema
from dataclasses import dataclass, field
from whylogs.core.datatypes import DataType, Fractional, Integral
from typing import Dict, List
from whylogs.core.metrics.metrics import Metric



class MyCustomResolver(Resolver):
    """Resolver that keeps distribution metrics for Fractional and frequent items for Integral, and counters and types metrics for all data types."""

    def resolve(self, name: str, why_type: DataType, column_schema) -> Dict[str, Metric]:
        metrics: List[StandardMetric] = [StandardMetric.counts, StandardMetric.types]
        if isinstance(why_type, Fractional):
            metrics.append(StandardMetric.distribution)
        if isinstance(why_type, Integral):
            metrics.append(StandardMetric.frequent_items)


        result: Dict[str, Metric] = {}
        for m in metrics:
            result[m.name] = m.zero(column_schema.cfg)
        return result


@dataclass
class MyCustomSchema(DatasetSchema):
    resolvers = MyCustomResolver()


schema = MyCustomSchema()
print(schema.resolvers)
```

richard-rogers:
> I tested on my previous example, and it's still getting the resolver from the base class.

You have to use `field` in the subclass. I'll update the comment to indicate that. 

```python
@dataclass
class MyCustomSchema(DatasetSchema):
    resolvers = field(default_factory=MyCustomResolver)
```

python/whylogs/core/schema.py
```diff
@@ -56,9 +56,9 @@ class DatasetSchema:
     """
 
     types: Dict[str, Any] = field(default_factory=dict)
-    default_configs: MetricConfig = MetricConfig()
-    type_mapper: TypeMapper = StandardTypeMapper()
-    resolvers: Resolver = StandardResolver()
+    default_configs: MetricConfig = field(default_factory=MetricConfig)
```
TheMellyBee, line 66: oof, this doesn't seem very readable especially as we make more custom ones like @FelipeAdachi example. Could this be iterated on?

TheMellyBee (changes requested):
Personally this doesn't seem very readable and easy to miss like Felipe did in even after the update. I think we could make this easy for users somehow

richard-rogers:
> Personally this doesn't seem very readable and easy to miss like Felipe did in even after the update. I think we could make this easy for users somehow

I did a lot of experimentation, and this was the only way I found that would override default values of class types. We could recommend passing the overriding values through the constructor, but then users would have to remember to do it everywhere rather than class authors remembering to do it once. Or make it not a `@dataclass` and see if that behaves more sensibly.

TheMellyBee:
What is the benefit of it being a dataclass here?

richard-rogers:
> What is the benefit of it being a dataclass here?

Originally it was not, but some of the code strongly suggested it was intended to be. So a few PRs ago I added the `@dataclass` decorator. I think we'd end up with essentially the same issue if we remove it though. Maybe split it into separate classes for config and behavior?

PR #750, opened 2022-08-13
DatasetSchema no longer a @dataclass

richard-rogers:
`DatasetSchema::copy()` should work for subclasses, but hopefully it won't need to be subclassed.

## Description

The 2 problems were

1. Users had to use `field(default_factory=...)` to override the defaults when subclassing `DatasetSchema`
2. `DatasetSchema::copy()` always returned `DatasetSchema` rather than the dynamic type of the copied schema

To fix 1, we're essentially saying "don't subclass `DatasetSchema`, just pass your `types` and `resolvers` into the constructor." If one did wish to subclass `DatasetSchema`, one would have to override `__init__()` to change the defaults, which is probably worse than using `field()`. But my impression is that users mostly only want to subclass `Resolver`, and passing that into the `DatasetSchema` constructor is probably better than subclassing `DatasetSchema` just to use a different `Resolver`.

Problem 2 is fixed by fixing it :) just in case users still subclass `DatasetSchema`.

## Related


python/whylogs/core/schema.py
```diff
@@ -40,30 +39,41 @@ class DatasetSchema:
...
     ...        "col1": str,
     ...        "col2": np.int32,
-    ...        "col3": pd.Categorical(categories=('foo', 'bar'), ordered=True)
-    ...    }
-    ...    cache_size = 12
-    >>> schema = MySchema()
-    >>> prof = DatasetProfile(MySchema())
+    ...        "col3": pd.CategoricalDtype(categories=('foo', 'bar'), ordered=True)
+    ...    },
+    ...    resolvers=MyResolver()
```
TheMellyBee, line 54: Now the user will need to get their own resolver correct? How did this fix the problem you were encountering before?

python/whylogs/core/schema.py
```diff
@@ -76,15 +86,18 @@ def __post_init__(self) -> None:
...
-                )
+        for col, tpe in self.types.items():
+            self._columns[col] = ColumnSchema(
+                dtype=tpe, resolver=self.resolvers, type_mapper=self.type_mapper, cfg=self.default_configs
+            )
 
     def copy(self) -> "DatasetSchema":
         """Returns a new instance of the same underlying schema"""
-        copy = DatasetSchema(**self.__dict__)
+        key_dict = vars(self).copy()
```
TheMellyBee, line 97: Personally I think this is more clear :)

TheMellyBee (commented):
Just some questions around the resolver.

python/whylogs/core/schema.py
TheMellyBee: The new changes to be more clear.

python/whylogs/core/schema.py
richard-rogers: The 2 problems were
1. Users had to use `field(default_factory=...)` to override the defaults when subclassing `DatasetSchema`
2. `DatasetSchema::copy()` always returned `DatasetSchema` rather than the dynamic type of the copied schema

To fix 1, we're essentially saying "don't subclass `DatasetSchema`, just pass your `types` and `resolvers` into the constructor." If one did wish to subclass `DatasetSchema`, one would have to override `__init__()` to change the defaults, which is probably worse than using `field()`. But my impression is that users mostly only want to subclass `Resolver`, and passing that into the `DatasetSchema` constructor is probably better than subclassing `DatasetSchema` just to use a different `Resolver`.

Problem 2 is fixed by fixing it :) just in case users still subclass `DatasetSchema`.

python/whylogs/core/schema.py
richard-rogers: We'll have to update all the examples, test code, and documentation to stop subclassing `DatasetSchema`.

python/whylogs/core/schema.py
```diff
@@ -76,15 +86,18 @@ def __post_init__(self) -> None:
...
+        key_dict.pop("_columns")
+        keys = key_dict.keys()
+        args = {k: self.__dict__[k] for k in keys if k not in self.types}
```
richard-rogers: I don't understand why the `self.types` items end up in `self.__dict__`

python/whylogs/core/schema.py
TheMellyBee: That makes sense :) I agree.

python/whylogs/core/schema.py
TheMellyBee: ooh how fun

python/whylogs/core/schema.py
TheMellyBee: when you're doing a copy why not use import deepcopy or import copy? I believe those module are accessible for use

TheMellyBee: Also are you saying you needed to fix this because they ended up in self._dict__ or they are somehow still getting in there?

richard-rogers: `deepcopy` might be better...

I don't understand how the items in `self.types` end up in `self.__dict__`. Maybe I don't fully understand `vars()`. Line 99 makes sure the `self.types` items aren't passed as arguments to the `DatasetSchema` constructor, which would complain about them because they're not members...

FelipeAdachi (approved):
LGTM!
I don't have any specific suggestions. Passing resolvers, etc. into constructor feels cleaner to me, so I think we can update the examples later to show this.

I was trying to think of cases where subclassing would be required, but couldn't think of none. If there are edge cases in the future where it is needed, we should just make sure to show an example of how to do it properly.

python/whylogs/core/schema.py
murilommen: i think `vars()` is just a builtin method to access an object's attributes that are stored on `__dict__` and it makes the code more readable :)

python/whylogs/core/schema.py
```diff
@@ -40,30 +39,41 @@ class DatasetSchema:
...
-    schema_based_automerge: bool = False
-
-    def __post_init__(self) -> None:
-        self._columns = {}
+    def __init__(
+        self,
+        types: Optional[Dict[str, Any]] = None,
+        default_configs: Optional[MetricConfig] = None,
+        type_mapper: Optional[TypeMapper] = None,
+        resolvers: Optional[Resolver] = None,
```
murilommen, line 67: maybe name it on singular if it takes in only one Resolver type object

python/whylogs/core/schema.py
```diff
@@ -76,15 +86,18 @@ def __post_init__(self) -> None:
                 LARGE_CACHE_SIZE_LIMIT,
             )
 
-        if self.types:
-            for col, tpe in self.types.items():
-                self._columns[col] = ColumnSchema(
-                    dtype=tpe, resolver=self.resolvers, type_mapper=self.type_mapper, cfg=self.default_configs
-                )
+        for col, tpe in self.types.items():
```
murilommen: what is tpe?

murilommen: it'd be better to have it more explicit, as `data_type` or something

python/whylogs/core/schema.py
FelipeAdachi: same thought on default_configs?

richard-rogers: I agree in principle, but I don't want to break backward compatibility by changing it :(

python/whylogs/core/schema.py
richard-rogers: That I knew :)  The same thing happened when I used `self.__dict__` instead of `vars()`.  I just don't understand how/where the `self.types` items get into `self.__dict__`.

jamie256 (approved):
Looks fine.

PR #752, opened 2022-08-16
Update push-release to not block on autobump version

jamie256:
-- lets start by not blocking on the pre-commit auto edits, to get the release automation unblocked.
-- update dependency cache key

The previous failures occur because a pre-commit hook if formatting or editing a file and then fails the autobump step. by allowing a dirty (files changed) locally we won't block the publish step.


FelipeAdachi (approved):
LGTM!

PR #753, opened 2022-08-16
WhyLabsWriter's log_reference_profile

FelipeAdachi:
- implementation
- tests
- example

## Description

This repository needs 

## Changes

## Related

### DIFF
--- a/python/examples/integrations/writers/Writing_Reference_Profiles_to_WhyLabs.ipynb
+++ b/python/examples/integrations/writers/Writing_Reference_Profiles_to_WhyLabs.ipynb
@@ -79,25 +79,41 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Enter your WhyLabs Org ID\n",
+      "Enter your WhyLabs Dataset ID\n",
+      "Enter your WhyLabs API key\n",
+      "Using API Key ID:  ygG04qE3gQ\n"
+     ]
+    }
+   ],
    "source": [
     "import getpass\n",
     "import os\n",
     "\n",
+    "os.environ[\"WHYLABS_API_ENDPOINT\"] = \"https://songbird.development.whylabsdev.com\"\n",
+    "\n",
     "# set your org-id here - should be something like \"org-xxxx\"\n",
     "print(\"Enter your WhyLabs Org ID\") \n",
-    "os.environ[\"WHYLABS_DEFAULT_ORG_ID\"] = input()\n",
+    "# os.environ[\"WHYLABS_DEFAULT_ORG_ID\"] = input()\n",
+    "os.environ[\"WHYLABS_DEFAULT_ORG_ID\"] = \"org-0\"\n",
     "\n",
     "# set your datased_id (or model_id) here - should be something like \"model-xxxx\"\n",
     "print(\"Enter your WhyLabs Dataset ID\")\n",
-    "os.environ[\"WHYLABS_DEFAULT_DATASET_ID\"] = input()\n",
-    "\n",
+    "# os.environ[\"WHYLABS_DEFAULT_DATASET_ID\"] = input()\n",
+    "os.environ[\"WHYLABS_DEFAULT_DATASET_ID\"] = \"model-2110\"\n",
     "\n",
     "# set your API key here\n",
     "print(\"Enter your WhyLabs API key\")\n",
-    "os.environ[\"WHYLABS_API_KEY\"] = getpass.getpass()\n",
+    "# os.environ[\"WHYLABS_API_KEY\"] = getpass.getpass()\n",
+    "os.environ[\"WHYLABS_API_KEY\"] = \"ygG04qE3gQ.7loXmAkuMQ8yyjwDhH47tdqpAowlIieGoc2527pPL52PDCI9EKzw1\"\n",
+    "\n",
     "print(\"Using API Key ID: \", os.environ[\"WHYLABS_API_KEY\"][0:10])"
    ]
   },
@@ -112,7 +128,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [
     {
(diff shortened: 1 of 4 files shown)


python/whylogs/api/writer/whylabs.py
```diff
@@ -66,20 +67,32 @@ def __init__(self, org_id: Optional[str] = None, api_key: Optional[str] = None,
...
 
     def option(
-        self, org_id: Optional[str] = None, dataset_id: Optional[str] = None, api_key: Optional[str] = None
-    ) -> None:
+        self,
+        org_id: Optional[str] = None,
+        dataset_id: Optional[str] = None,
+        api_key: Optional[str] = None,
+        reference: Optional[bool] = None,
+        reference_alias: Optional[str] = None,
```
jamie256: How about dropping the bool parameter and just inferring this from the name given?
```suggestion
        reference_profile_name: Optional[str] = None,
```

python/whylogs/api/writer/whylabs.py
```diff
@@ -66,20 +67,32 @@ def __init__(self, org_id: Optional[str] = None, api_key: Optional[str] = None,
...
         if dataset_id is not None:
             self._dataset_id = dataset_id
         if org_id is not None:
             self._org_id = org_id
         if api_key is not None:
             self._api_key = api_key
+        if reference is not None:
+            self._reference = reference
+        if reference_alias is not None:
+            self._reference_profile_alias = reference_alias
```
jamie256: from above:
```suggestion
        if reference_profile_name is not None:
            self._reference_profile_name = reference_alias
```

python/examples/integrations/writers/Writing_Reference_Profiles_to_WhyLabs.ipynb
```diff
@@ -0,0 +1,422 @@
...
+   "source": [
+    "## ✍️ The WhyLabs Writer"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now, you can simply create a WhyLabsWriter object and use it to send your profiles.\n",
+    "The process is very similar to sending a regular profile. The only difference is you should the `option` method in order to pass `reference` as True and `reference_alias` as the alias for your Reference Profile:"
```
jamie256: Suggest following rephrasing, also probably good to use the same language in our platform e.g. "static" and while the log_reference api uses "alias" that term might not be helpful in this context or the new option parameter, maybe just call it something like reference profile name?
```suggestion
    "The process is very similar to sending a regular profile. The only difference is you should use the `option` method in order to specify `reference_profile_name` to indicate and name this profile as a static Reference Profile:"
```

jamie256 (commented):
Some minor rephrasing and renaming suggestions. Looks good!

python/whylogs/api/writer/whylabs.py
FelipeAdachi: That's a good idea!

Just one thing: the WhyLabs API allows you to create a reference profile without a name. In this case, the name is the timestamp of when it is created.

Inferring from the name would prevent the user of logging a reference profile without a name, even though the API allows it. Do we want that?
@jamie256

python/whylogs/api/writer/whylabs.py
```diff
@@ -66,20 +67,32 @@ def __init__(self, org_id: Optional[str] = None, api_key: Optional[str] = None,
...
             raise BadConfigError("Bad WhyLabsWriter config: interval must be greater or equal to five minutes")
 
     def option(
-        self, org_id: Optional[str] = None, dataset_id: Optional[str] = None, api_key: Optional[str] = None
-    ) -> None:
+        self,
+        org_id: Optional[str] = None,
+        dataset_id: Optional[str] = None,
+        api_key: Optional[str] = None,
+        reference: Optional[bool] = None,
```
richard-rogers: If we keep this, make it `reference: bool = False` instead of `Optional`

python/whylogs/api/writer/whylabs.py
richard-rogers: Use `""` for an unnamed reference profile, and `None` for "not a reference profile?"

python/whylogs/api/writer/whylabs.py
```diff
@@ -132,22 +155,40 @@ def _upload_whylabs(
...
-                        f"{self.whylabs_api_endpoint} with API token ID: {api_key_id}"
-                    )
-                return http_response
-        except requests.RequestException as e:
-            logger.info(
-                f"Failed to upload {self._org_id}/{self._dataset_id}/{dataset_timestamp} to "
-                + f"{self.whylabs_api_endpoint} with API token ID: {api_key_id}. Error occurred: {e}"
-            )
+        if not reference:
+            try:
```
richard-rogers: I don't see a difference between the `if` and `else`branches?

python/whylogs/api/writer/whylabs.py
```diff
@@ -132,22 +155,40 @@ def _upload_whylabs(
...
+                with open(profile_path, "rb") as f:
+                    http_response = requests.put(upload_url, data=f.read())
+                    if http_response.status_code == 200:
+                        logger.info(
+                            f"Done uploading {self._org_id}/{self._dataset_id}/{dataset_timestamp} to "
+                            f"{self.whylabs_api_endpoint} with API token ID: {api_key_id}"
```
richard-rogers: Hopefully the API token is not secret :)

python/whylogs/api/writer/whylabs.py
FelipeAdachi: ah, sure. I did try this before but didn't realize it was just a tweak to make it work.
changed this in the last commmit.

python/whylogs/api/writer/whylabs.py
FelipeAdachi: changed in the last commit

python/examples/integrations/writers/Writing_Reference_Profiles_to_WhyLabs.ipynb
FelipeAdachi: agreed, changed it.

python/whylogs/api/writer/whylabs.py
FelipeAdachi: haha, yes. must be an old code I forgot to update.
fixed in the last commit

python/whylogs/api/writer/whylabs.py
FelipeAdachi: I got scared for a bit, but if you see `api_key_id`, it's already only the first 10 characters of the complete key!
So I think it's good.

jamie256 (approved):
LGTM!

PR #755, opened 2022-08-17
pyproject.toml dependency updates, and poetry.lock sync

jamie256:
## Description
poetry.lock updates.
force mistune update by putting it as an explicit dev dependency (was only in via nbconvert from sphinx for docs extras).


TheMellyBee (approved):
LGTM! Love to see how many became optional and/or moved to dev. As long as the build comes back green you're good

PR #757, opened 2022-08-17
Aggregator for Metric Components

TheMellyBee:
## Description
Adds a registry for the aggregators. Adds just the sum aggregator to get started. 

I changed it halfway through from requiring the IAggregator which extends BiFunction to anything that extends BiFunction. This is because I realized post lambdas are using the BiOperator which is a subclass of BiFunction. I'm not thrilled with the API though as it returns the BiFunction not the aggregator itself. It's fine since all the aggregators are required to have an apply() so it works the same. Take a peek and give your thoughts.

## Changes

- 6e9339e101b0987657c39a660cb33a038142cb58 Adds a global registry for aggregators on components
- 46327f50a8515c48045f6e261ec58186b1a2b5d2 Changes the registry to be class that extends BiFunction

## Related
Building out Java v1

### DIFF
--- a/java/core/src/main/java/com/whylogs/core/metrics/aggregators/AggregatorRegistry.java
+++ b/java/core/src/main/java/com/whylogs/core/metrics/aggregators/AggregatorRegistry.java
@@ -1,0 +1,33 @@
+package com.whylogs.core.metrics.aggregators;
+
+import com.whylogs.core.metrics.components.MetricComponent;
+import lombok.Data;
+
+import java.util.HashMap;
+
+@Data
+public class AggregatorRegistry {
+    private HashMap<String, IAggregator> namedAggregators;
+    private HashMap<Integer, IAggregator> idAggregators;
+
+    public <T extends MetricComponent> void register(T component, IAggregator aggregator) {
+        idAggregators.put(component.getTypeId(), aggregator);
+        namedAggregators.put(component.getTypeName(), aggregator);
+    }
+
+    public void register(int typeId, IAggregator aggregator) {
+        idAggregators.put(typeId, aggregator);
+    }
+
+    public IAggregator get(MetricComponent component) {
+        return idAggregators.get(component.getTypeId());
+    }
+
+    public IAggregator get(int typeId) {
+        return idAggregators.get(typeId);
+    }
+
+    public IAggregator get(String typeName) {
+        return namedAggregators.get(typeName);
+    }
+}
--- a/java/core/src/main/java/com/whylogs/core/metrics/aggregators/AggregatorRegistry.java
+++ b/java/core/src/main/java/com/whylogs/core/metrics/aggregators/AggregatorRegistry.java
@@ -1,52 +1,49 @@
 package com.whylogs.core.metrics.aggregators;
 
 import com.whylogs.core.metrics.components.MetricComponent;
-import lombok.Data;
-import lombok.Synchronized;
-
 import java.util.HashMap;
 import java.util.function.BiFunction;
-
 
 // Registry for aggregators. You can register a component or a type-id to an aggregator.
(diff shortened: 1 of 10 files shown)


TheMellyBee:
@jamie256 when we talked about using annotations, we can have a custom annotation for a method to be added to this registry. Let me know if you want that added in this or a later.

java/core/src/main/java/com/whylogs/core/metrics/aggregators/AggregatorRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+    // Instance holder to avoid double checking anti-pattern for singlton
+    private static final class InstanceHolder {
+        static final AggregatorRegistry instance = new AggregatorRegistry();
+    }
+
+    public static AggregatorRegistry getInstance() {
+        return InstanceHolder.instance;
+    }
+
+    public <T extends MetricComponent, A extends BiFunction> void register(T component,  A aggregator) {
```
richard-rogers: Should `A` extend `IAggrevator` ?

java/core/src/main/java/com/whylogs/core/metrics/aggregators/IAggregator.java
```diff
@@ -0,0 +1,9 @@
+package com.whylogs.core.metrics.aggregators;
+
+import com.whylogs.core.errors.UnsupportedError;
+
+import java.util.function.BiFunction;
+
+public interface IAggregator<T, U, R> extends BiFunction<T, U, R> {
```
richard-rogers: Is it not always the case that `T == U == R` ?

TheMellyBee: I guess I was leaving it open in case a custom one wanted to be able to merge two different types. Sheep and cows to get the herd type of thing. I can change it if we would rather.

java/core/src/main/java/com/whylogs/core/metrics/aggregators/AggregatorRegistry.java
TheMellyBee: I went back and forth. To do a lambda I think it needs to be able to be BiFunction

java/core/src/main/java/com/whylogs/core/metrics/aggregators/IAggregator.java
richard-rogers: I don't think we (should) support merging components outside the context of metrics. I don't think we (should) support heterogeneous metric merges either :)  So I vote `T == U == R`.

java/core/src/main/java/com/whylogs/core/metrics/aggregators/AggregatorRegistry.java
richard-rogers: Ah, I didn't grok that when I read the description...  It looks like we can drop `IAggregator` now?

jamie256 (approved):
Agree with Richard's comment on restricting merges to same type.

java/core/src/main/java/com/whylogs/core/metrics/aggregators/AggregatorRegistry.java
TheMellyBee: Ok, we could do that. There is the benefit of when we want to extend from aggregators as classes they get the merge methods. So yes it could be removed, but at the same time since IAggregtor extends BiFunction it's fine, and I think makes life a little nicer to read when making customs, *but* no not needed. Thoughts?

java/core/src/main/java/com/whylogs/core/metrics/aggregators/IAggregator.java
TheMellyBee: Done

jamie256 (approved):
LGTM!

PR #760, opened 2022-08-22
Remove the java release's python integration now that we have pyspark

jamie256:
## Description

The java/python dependencies aren't used here and we should remove them over updating and maintaining them.

dependabot scans detected potential issues in using older versions of spark referenced in these unused poetry.lock file here:
* https://github.com/advisories/GHSA-9rr6-jpg7-9jg6
* https://github.com/advisories/GHSA-4x9r-j582-cgr8


FelipeAdachi (approved):
LGTM!

PR #761, opened 2022-08-22
Condition Validators

FelipeAdachi:
## Description

**Condition Validators** use the same conditions as **Condition Count Metrics**, but with a different goal. Validators will evaluate a set of conditions for a given column and, if the condition fails (is False), will trigger one or more actions.

Condition Validators evaluate individual values, just as the Condition Count Metrics. In the future, it could make sense to think of Validators for Metrics as they are being logged (similar to Metric Constraints, but in a streaming scenario?).

The motivation is to have a way of validating your data in a real-time/streaming scenario, where immediate actions need to be taken, like sending an alert or hitting an emergency stop button. No metrics are logged with Condition Validators, only evaluations and triggered actions.

## Changes

## Related


python/whylogs/core/validators/condition_validator.py
```diff
@@ -0,0 +1,52 @@
...
+from whylogs.core.validators.validator import Validator
+
+
+@dataclass
+class ConditionValidator(Validator):
+    def __init__(self, name: str, conditions: Dict[str, Condition], actions=List[Callable]) -> None:
+        self.conditions = conditions
+        self.name = name
+        self.actions = actions
+        self.total: IntegralComponent = IntegralComponent(0)
```
richard-rogers: Why not plain integers?

richard-rogers: Actually, I would pull `total_evaluations` and `total_failures` out of `conditions` and make them separate dictionaries. I'm not sure if we need the totals for the validator (rather than just the individual conditions)?

python/whylogs/core/validators/validator.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class Validator(ABC):
+    name: str
+    conditions: Dict[str, Condition]
+    actions: List[Callable]
+
+    @abstractmethod
+    def columnar_validate(self, data: PreprocessedColumn):
+        raise NotImplementedError
```
richard-rogers: I might add `to_summary_dict()` to the `Validator` interface.

python/whylogs/core/validators/condition_validator.py
FelipeAdachi: Yes, you're right, I carried `IntegralComponent` from the Condition Count implementation, but it's not needed here. I'll remove it.

I agree that maybe we don't need `total_failures`. If only `total_evaluations` is left, do we still need to separate it into another dictionary?

My idea is to make it like:
```
{
"total_evaluations": 75,
"condition#1": 2,
"condition#2": 1,
.
.
.
}
```

python/whylogs/core/validators/validator.py
FelipeAdachi: Done!

python/whylogs/core/schema.py
```diff
@@ -89,7 +92,11 @@ def __init__(
 
         for col, data_type in self.types.items():
             self._columns[col] = ColumnSchema(
-                dtype=data_type, resolver=self.resolvers, type_mapper=self.type_mapper, cfg=self.default_configs
+                dtype=data_type,
+                resolver=self.resolvers,
+                validators=self.validators,
```
jamie256: can you put validators last?

python/whylogs/core/schema.py
```diff
@@ -164,6 +173,13 @@ class ColumnSchema:
     cfg: MetricConfig = MetricConfig()
     type_mapper: TypeMapper = StandardTypeMapper()
     resolver: Resolver = StandardResolver()
+    validators: Dict[str, Validator] = field(default_factory=dict)
 
     def get_metrics(self, name: str) -> Dict[str, Metric]:
         return self.resolver.resolve(name=name, why_type=self.type_mapper(self.dtype), column_schema=self)
+
+    def get_validators(self, name: str) -> Set[Validator]:
```
jamie256: Two things:
1) rename this to something singular like: get_validator()
2) If self.validators is a dictionary of Validator values, then this should return Optional[Validator] right?

python/whylogs/core/schema.py
```diff
@@ -164,6 +173,13 @@ class ColumnSchema:
...
+        try:
+            return self.validators[name]
+        except KeyError:
+            return set()
```
jamie256: Avoid using exception handling for missing keys, better to use something like:
```suggestion
        if self.validators:
            return self.validators.get(name)
```

python/whylogs/core/validators/condition_validator.py
```diff
@@ -0,0 +1,51 @@
...
+@dataclass
+class ConditionValidator(Validator):
+    def __init__(
+        self, name: str, conditions: Dict[str, Condition], actions=List[Callable[[str, str, Any], None]]
+    ) -> None:
+        self.conditions = conditions
+        self.name = name
+        self.actions = actions
+        self.total: int = 0
+        self.failures: Dict[str, int] = dict()
```
jamie256: With dataclass you don't have to write most of that init method:
```suggestion
        name: str
        conditions: Dict[str, Condition]
        actions=List[Callable[[str, str, Any], None]]
        total: int = 0
        failures: Dict[str, int] = field(default_factory=dict)
```
Not sure you need the condition key checks?

python/whylogs/core/validators/validator.py
```diff
@@ -0,0 +1,19 @@
+from abc import ABC, abstractmethod
+from typing import Any, Callable, Dict, List
+
+from whylogs.core.metrics.condition_count_metric import Condition
+from whylogs.core.preprocessing import PreprocessedColumn
+
+
+class Validator(ABC):
```
jamie256, line 8: This is ok, but do we need a base class? Is that intended to be an extensibility point?

python/whylogs/core/validators/condition_validator.py
```diff
@@ -0,0 +1,51 @@
+from dataclasses import dataclass
+from itertools import chain
+from typing import Any, Callable, Dict, List
+
+from whylogs.core.metrics.condition_count_metric import Condition
```
jamie256, line 5: Why use Condition? The name sounds fitting but it doesn't appear to simplify the code, seems like you just need a function that returns bool?

python/whylogs/core/validators/condition_validator.py
```diff
@@ -0,0 +1,51 @@
...
+            count += 1
+            for cond_name, condition in self.conditions.items():
+                try:
+                    if not condition.relation(x):
+                        self.failures[cond_name] += 1
+                        count_failures += 1
+                        for action in self.actions:
+                            action(self.name, cond_name, x)
+                except:  # noqa
+                    pass
```
jamie256: it is surprising to have an exception be "passing" the condition.

What if we remove the try/except block?

or if we keep it, make the except case count as a failure by default.

jamie256 (commented):
Left some comments, summarizing discussion offline.

python/whylogs/core/validators/condition_validator.py
FelipeAdachi: I changed it as per my last comment. resolving this comment, if needed we can reopen it.

python/whylogs/core/schema.py
FelipeAdachi: done in the latest commits

python/whylogs/core/schema.py
FelipeAdachi: yeah that was wrong, but it's actually a list of Validators, so I figured it would be best to stick with get_validators() in the plural, and change the return to be `List[Optional[Validator]]`.

python/whylogs/core/schema.py
FelipeAdachi: done in the latest commits

python/whylogs/core/validators/validator.py
FelipeAdachi: Yes, I thought we could extend from Base Validator class if we have Validators operating on Metrics. But if the Conditions themselves are changed to work with metrics, this use case might not be valid anymore.

But we could have other unforeseen validators in the future.

What do you think? Do we want to remove that base class, or keep it?

python/whylogs/core/validators/condition_validator.py
FelipeAdachi: Having predefined conditions feel like the right way, but I don't have a good argument for that. Maybe to simplify serialization in the future?
@richard-rogers , any thoughts on this?

python/whylogs/core/validators/condition_validator.py
FelipeAdachi: I agree, removed it in the latest commit

python/tests/core/validators/test_condition_validator.py
```diff
@@ -0,0 +1,121 @@
...
+
+def test_condition_validator_dataframe(credit_card_validator, transcriptions):
+
+    df = pd.DataFrame(data=transcriptions, columns=["transcriptions"])
+    validators = {"transcriptions": [credit_card_validator]}
+
+    schema = DatasetSchema(validators=validators)
+    profile = why.log(df, schema=schema).profile()
+
+    profile._columns["transcriptions"]._column_validators[0].total == 4
```
richard-rogers: were these meant to be `assert`s?

python/whylogs/core/validators/condition_validator.py
richard-rogers: `Condition` should be split-up/moved out of `condition_count_metric.py` since the goal is to share the predicate infrastructure between the condition count metric and validators. I've made a start at this in the binary condition PR #781

python/whylogs/core/validators/condition_validator.py
```diff
@@ -0,0 +1,40 @@
...
+from typing import Any, Callable, Dict, List
+
+from whylogs.core.metrics.condition_count_metric import Condition
+from whylogs.core.preprocessing import PreprocessedColumn
+from whylogs.core.validators.validator import Validator
+
+
+@dataclass
+class ConditionValidator(Validator):
+    conditions: Dict[str, Condition]
```
richard-rogers, line 12: I think this will need a `field(default_factory=dict)`

python/whylogs/core/validators/condition_validator.py
FelipeAdachi: I think @jamie256's question was about the need to create conditions with Relations, when a lambda expression could cover all of our cases? or maybe any function that returns a bool.

richard-rogers: Originally, `Condition` carried a little metadata in addition to the `Callable` implementing the predicate.  We should definitely refactor to share between validators and metrics. #781 is a start, but could probably use more refactoring still.

python/tests/core/validators/test_condition_validator.py
FelipeAdachi: yes, they were, thanks!

fixed.

python/whylogs/core/validators/condition_validator.py
jamie256: If you are adding more functionality to Condition then it seems better justified. At present the benefits of `Condition `weren't clear over using a lambda or predicate function and requiring a class for `Condition` increases complexity in how users have to construct validators so its seemed to detract from usability.

python/whylogs/core/validators/validator.py
jamie256: I was looking to understand the motivation, thanks for explanation.

jamie256 (approved):
Looks good, we can address some of the follow ups outside this PR.

PR #762, opened 2022-08-22
Add Serialization and Deerialization Registry

TheMellyBee:
## Description
Adding in the metric serialization and deserialization registries with the number serializers. This change utilizes the typing system, but doesn't have annotation. Check out commit f9d14e770a4d53c9a350c9ab82aeff5b8f7bb263 if we want to do annotation in the future, but the reflection is going to be slower and annotation adds some interesting design questions. Will implement in another PR if we believe the user benefit makes sense. 

## Changes

- Last change adds a lot for the linter. Please look at b2a5e8d08dcba192ca6d6c04ef544c19f9d5c846 if you want the complete change without the linter. (which is still changing a bunch do to different PRs in flight) 
- f9d14e770a4d53c9a350c9ab82aeff5b8f7bb263 start of annotation WIP for future thoughts


java/core/src/test/java/com/whylogs/core/metrics/serializers/TestSerializer.java
```diff
@@ -0,0 +1,38 @@
...
+import org.testng.annotations.Test;
+
+@Test
+public class TestSerializer {
+
+  @Test
+  public void test_serializationRegistry() {
+    SerializerRegistry registry = SerializerRegistry.getInstance();
+    Assert.assertNotNull(registry);
+    Assert.assertEquals(registry, SerializerRegistry.getInstance());
```
jamie256: minor:
```suggestion
    Assert.assertSame(registry, SerializerRegistry.getInstance());
```

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
```diff
@@ -19,8 +19,8 @@ public class IntegralMetric extends AbstractMetric<IntegralMetric> {
 
   public IntegralMetric() {
     super(IntegralMetric.NAMESPACE);
-    this.maxComponent = new MaxIntegralComponent();
-    this.minComponent = new MinIntegralComponent();
+    this.maxComponent = new MaxIntegralComponent(Integer.MIN_VALUE);
+    this.minComponent = new MinIntegralComponent(Integer.MAX_VALUE);
```
jamie256, line 23: Is this related to serialization / deserialization?

jamie256 (commented):
Reading your description: I don't think you need to do the annotation portion. 

Looks like you are only plumbing through numeric type serialization first (so other types to come later).

some minor comments/questions.

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
TheMellyBee: This was due to a bug that I found while testing. I hadn't correctly done the IntegralMetric default and the CI wasn't up at that point so the test fail got missed. This was the fix.

jamie256 (approved):
LGTM!

PR #768, opened 2022-08-24
Update Documentation - Getting Started

FelipeAdachi:
- combine both Getting_Started.ipynb and getting_started.md
- remove getting_started.md and point Getting Started section to example notebook
- Updated Getting_Started.ipynb with recent examples

## Description

This repository needs 

## Changes

## Related


python/examples/basic/Getting_Started.ipynb
```diff
@@ -438,9 +449,17 @@
...
+    "    - [Writing Profiles](https://whylogs.readthedocs.io/en/stable/examples/integrations/Writing_Profiles.html) - See different ways and locations to output your profiles\n",
+    "    - [Flask](https://whylogs.readthedocs.io/en/stable/examples/integrations/flask_streaming/flask_with_whylogs.html) - See how you can create a Flask app with whylogs and WhyLabs integration\n",
+    "    - [Feature Stores](https://whylogs.readthedocs.io/en/stable/examples/integrations/Feature_Stores_and_whylogs.html) - Learn how to log features from your Feature Store with feast and whylogs\n",
+    "    - [BigQuery](https://whylogs.readthedocs.io/en/stable/examples/integrations/BigQuery_Example.html) - Profile data queried from a Google BigQuery table\n",
+    "    - [MLflow](https://whylogs.readthedocs.io/en/stable/examples/integrations/Mlflow_Logging.html) - Log your whylogs profiles to an MLflow environment\n",
+    "\n",
+    "Or go to the [examples page](https://whylogs.readthedocs.io/en/stable/examples.html) for the complete list of examples!"
```
jamie256: Why are we leading users away from our github repo when they are exploring the example notebooks? The html versions here are less useful than the ipynb files  since you can't run them.

Also some of these links are broken: https://whylogs.readthedocs.io/en/stable/examples/integrations/Writing_Profiles.html

FelipeAdachi: If users are seeing this on the documentation, then we'd be leading them away from the docs page. Maybe that's ok?

Another issue is that notebooks are not well rendered on github (a lot of plots and graphs go missing when viewing them on github), which doesn't happen on readthedocs. Referencing directly to the repo would also point to the "latest" examples, instead of the stable ones, but maybe it's not an issue.

About the fact that you can't run them, I think the collab button takes care of it? you can run it from docs page or gh repo on collab.

jamie256: You are right, the visuals are better in the html version, and the colab button works there, good points.

Maybe it is fine. I find switching between the two locations a bit confusing. I would prefer as a user that the example notebooks in github would link to other example notebooks in github (if logically equivalent content is in more than one location). If that isn't easy to do then maybe this is an acceptable trade off to link into the read the docs here.

But we at least need to fix the links so they don't give us 400 status responses.

jamie256 (approved):
LGTM, just need to resolve some conflicts and good to go!

python/examples/basic/Getting_Started.ipynb
FelipeAdachi: Fixed the links!

Just pasting from slack for reference:

I actually agree with you on it being confusing switching between places. I don't have an easy way to do that for the moment, so I'll keep the PR as is, and try to think of a solution to change that later!

PR #772, opened 2022-08-24
WIP Automating Broken Link Check in ipynb

### LINKED ISSUE #770: Need a way to check for broken links in notebooks
## Description

We have link checking for markdown, but we still missed a broken link in an example notebook. We need a way to check for those in our CI:
e.g. https://github.com/whylabs/whylogs/issues/769
https://whylogs.readthedocs.io/en/stable/examples/integrations/Writing_Profiles.html

TheMellyBee:
## Description
This is playing with the action 

## Related
closes #770 


TheMellyBee:
Sadly there is an external problem with the link checker. It grabs the entire line at times which has " and other characters such as >.

There is a work around by 1. Ignoring or 2. adding a space between the end of the link and the ending ". Now, this unfortunately is most of the links so 1 would make it not very useful for automation. 2 does work, but is funky. The good news is it still works as a link just fine, and if we ever had to join the string which is rare in .ipynb a strip fixes it. This would take a while to add a space to the end of every string, and it would be difficult for the team to remember.

For now, we are closing the PR to watch the external bug and revisit.

PR #779, opened 2022-08-26
StandardResolver and Supporting classes

TheMellyBee:
## Description
Adds the Schemas, a DataType, and Resolver along with all tests. Again this is focused on the ints to be able to focus the whole way through. 


java/core/src/main/java/com/whylogs/core/metrics/StandardMetric.java
```diff
@@ -0,0 +1,52 @@
+package com.whylogs.core.metrics;
+
+public enum StandardMetric {
+  /*
```
richard-rogers: Remove comment

java/core/src/main/java/com/whylogs/core/resolvers/StandardResolver.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class StandardResolver extends Resolver {
+  public StandardResolver() {
+    super();
+  }
+
+  // TODO: the rest of the metrics need implmeented
+  // TODO: does this Metric loose it's typing?
+  @Override
+  public HashMap<String, ? extends Metric> resolve(ColumnSchema schema) {
```
TheMellyBee: Note: this doesn't have the column_name passed as it is never used in any python resolver

java/core/src/main/java/com/whylogs/core/metrics/StandardMetric.java
TheMellyBee: This are here so we know what types we need to come back to implement in the next pass (next week)

java/core/src/main/java/com/whylogs/core/DataTypes.java
```diff
@@ -0,0 +1,68 @@
+package com.whylogs.core;
+
+import java.lang.reflect.Type;
+import java.util.HashSet;
+
+public enum DataTypes {
+  Numerical {
```
TheMellyBee, line 7: The DataTypes Enum works as a type mapper given we are using the type system, but this could be changed a bit to have classes that function similarly if we believe users will need more customization.

java/core/src/main/java/com/whylogs/core/resolvers/StandardResolver.java
jamie256: For Dataschema override where user specifies type overrides for a column, not used in standard resolvers, so fine to leave off at first if you want, but its part of the python interface for that reason.

java/core/src/main/java/com/whylogs/core/schemas/DatasetSchema.java
```diff
@@ -0,0 +1,79 @@
...
+  public boolean resolve(HashMap<String, ?> data) {
+    for (String key : data.keySet()) {
+      if (this.columns.containsKey(key)) {
+        continue;
+      }
+
+      this.columns.put(
+          key, new ColumnSchema(data.get(key).getClass(), this.defaultConfig, this.resolver));
+    }
+    return true;
```
jamie256: return true only if you created a new ColumnSchema is the logic in python, but I'm not sure we will use the "dirty" return value in java or not.

java/core/src/main/java/com/whylogs/core/resolvers/Resolver.java
```diff
@@ -0,0 +1,10 @@
+package com.whylogs.core.resolvers;
+
+import com.whylogs.core.metrics.Metric;
+import com.whylogs.core.schemas.ColumnSchema;
+import java.util.HashMap;
+
+public abstract class Resolver {
+
+  public abstract <T extends Metric> HashMap<String, T> resolve(ColumnSchema schema);
```
jamie256: suggest HashMap<String, Metric> return

java/core/src/main/java/com/whylogs/core/resolvers/StandardResolver.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    if (schema.getConfig().isFi_disabled()) {
+      // standardMetrics.remove(StandardMetric.frequent_items);
+    }
+
+    HashMap<String, Metric> result = new HashMap<>();
+
+    for (StandardMetric metric : standardMetrics) {
+      result.put(metric.name(), StandardMetric.getMetric(metric, schema.getConfig()));
```
jamie256: chatted offline, can probably use the metric.zero directly, consider using HashMap or some bidirectional mapping of metric name to metric types to support this.

java/core/src/test/java/com/whylogs/core/metrics/TestStandardMetricFactory.java
```diff
@@ -0,0 +1,19 @@
+package com.whylogs.core.metrics;
+
+import java.util.ArrayList;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+@Test
+public class TestStandardMetricFactory {
```
jamie256: name?

jamie256 (commented):
Remove some of the commented out lines, and left some minor comments below.

java/core/src/main/java/com/whylogs/core/resolvers/StandardResolver.java
TheMellyBee: Thank you :)

java/core/src/main/java/com/whylogs/core/schemas/DatasetSchema.java
TheMellyBee: When would we use the dirty? Is this something we should keep in or just remove?

TheMellyBee:
Sorry!  I was on the wrong branch so ended up merging this into itself?! Weird.

java/core/src/main/java/com/whylogs/core/resolvers/Resolver.java
```diff
@@ -0,0 +1,10 @@
+package com.whylogs.core.resolvers;
+
+import com.whylogs.core.metrics.Metric;
+import com.whylogs.core.schemas.ColumnSchema;
+import java.util.HashMap;
+
+public abstract class Resolver {
+
+  public abstract HashMap<String, Metric> resolve(ColumnSchema schema);
```
richard-rogers, line 9: I still vote this should take the column name, but we can add it later :)

java/core/src/main/java/com/whylogs/core/resolvers/StandardResolver.java
```diff
@@ -0,0 +1,35 @@
...
+    } else if (DataTypes.Fractional.includes(schema.getType())) {
+      throw new NotImplementedException("Fractional metrics not implemented");
+    } else if (DataTypes.String.includes(schema.getType())) {
+      if (schema.getConfig().isTrack_unicode_ranges()) {
+        throw new NotImplementedException("String & Unicode metrics not implemented");
+      }
+    }
+
+    if (schema.getConfig().isFi_disabled()) {
+      throw new NotImplementedException("Frequent Items metrics not implemented");
```
richard-rogers, line 31: Surely it's lack of implementation is not a problem if it's disabled?

java/core/src/main/java/com/whylogs/core/schemas/DatasetSchema.java
```diff
@@ -0,0 +1,80 @@
...
+    if (cache_size < 0) {
+      // TODO: log warning
+      this.cache_size = 0;
+    }
+
+    if (cache_size > LARGE_CACHE_SIZE_LIMIT) {
+      // TODO: log warning
+    }
+
+    // Type name will be used as the column name
```
richard-rogers, line 41: I think the keys are the column names.

java/core/src/main/java/com/whylogs/core/schemas/DatasetSchema.java
```diff
@@ -0,0 +1,80 @@
...
+      for (String typeName : this.types.keySet()) {
+        this.columns.put(
+            typeName,
+            new ColumnSchema(this.types.get(typeName), this.defaultConfig, this.resolver));
+      }
+    }
+  }
+
+  public DatasetSchema copy() {
+    DatasetSchema copy = new DatasetSchema();
```
richard-rogers, line 52: This will have the same problem the original Python code did---it always returns `DatasetSchema` instead of the dynamic type. I think we decided to discourage subclassing `DatasetSchema`, but we still might want to fix this.

java/core/src/main/java/com/whylogs/core/resolvers/StandardResolver.java
TheMellyBee: Oh this is just as a temp as I'm making that metric later this week :)

jamie256 (approved):
LGTM! check with Richard but looks good.

java/core/src/main/java/com/whylogs/core/schemas/ColumnSchema.java
```diff
@@ -0,0 +1,24 @@
...
+
+import com.whylogs.core.metrics.Metric;
+import com.whylogs.core.metrics.MetricConfig;
+import com.whylogs.core.resolvers.Resolver;
+import java.lang.reflect.Type;
+import java.util.HashMap;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
```
richard-rogers, line 11: I take it the `@Data` decorator makes the getters for the private members available?

java/core/src/main/java/com/whylogs/core/schemas/DatasetSchema.java
```diff
@@ -0,0 +1,80 @@
...
+      this.cache_size = 0;
+    }
+
+    if (cache_size > LARGE_CACHE_SIZE_LIMIT) {
+      // TODO: log warning
+    }
+
+    // Type name will be used as the column name
+    if (!this.types.isEmpty()) {
+      for (String typeName : this.types.keySet()) {
```
richard-rogers, line 43: s/typeName/columnName/

java/core/src/main/java/com/whylogs/core/schemas/ColumnSchema.java
TheMellyBee: Yup! :)

java/core/src/main/java/com/whylogs/core/schemas/DatasetSchema.java
TheMellyBee: Sorry, I don't know how to translate your comment?

PR #783, opened 2022-08-31
move link check into separate action

jamie256:
## Description

run link check only on manual dispatch to reduce noise in the ci checks on PRs and merge.

We had seen noise in failure to connect to links in our docs that were external (and live) and the failures were frequent and intermittent.

## Changes

remove the link check from whylogs-ci and allow this tool to be run manually. Consider follow up to run this as part of release process or at some other cadence less frequent than PR checks.


FelipeAdachi (approved):
LGTM!

PR #788, opened 2022-09-02
Update whylogs-sketching version and update poetry.lock

jamie256:
## Description

With the fix of the kll serialization size mismatch error in whylogs-sketching, updating the required version to include the fix, and update poetry.lock.


FelipeAdachi (approved):
LGTM!

Issue #784, opened 2022-08-31
release workflow hits error creating automated pull request after bump-version


jamie256:
## Description

the release workflow appears to be making automated changes to:
python/examples/integrations/flask_streaming/.dockerignore

and then failing to create the automated PR commits:
https://github.com/whylabs/whylogs/runs/8123396886?check_suite_focus=true#step:7:512

jamie256:
still encountered:
https://github.com/whylabs/whylogs/runs/8215090060?check_suite_focus=true#step:7:510

maybe we should try ignoring this file or creating the file on demand (Felipe's idea).

PR #789, opened 2022-09-06
Update version of create-pull-request

jamie256:
## Description

There have been recent fixes to the action, and they look similar to the error we see in #784 lets try latest version.


TheMellyBee (approved):
LGTM

PR #792, opened 2022-09-09
Initial segments support for v1.1

jamie256:
## Description

* Add support for Segment, partitions, filtered segments, custom mapping
* Limit scope to single partitions, column mapping for v1, send files as v0 format initially,


python/whylogs/api/logger/logger.py
```diff
@@ -65,6 +69,10 @@ def log(
             # TODO: check for shell environment and emit more verbose error string to let user know how to correct.
             raise LoggingError("log() was called without passing in any input!")
 
+        # If segments are defined use segment_processing to return a SegmentedResultSet
+        if self._schema and self._schema.segments:
+            return _log_partitions(self._schema, obj, pandas, row)
```
TheMellyBee: Why not call this segmentation-processing instead to make it clearer then? (Based off the comment) 

Or even log_segment?

python/examples/advanced/Segments.ipynb
```diff
@@ -0,0 +1,148 @@
...
+    "d = {\n",
+    "    \"col1\": [i for i in range(input_rows)],\n",
+    "    \"col2\": [i * i * 1.1 for i in range(input_rows)],\n",
+    "    segment_column: [f\"x{str(i%number_of_segments)}\" for i in range(input_rows)],\n",
+    "}\n",
+    "\n",
+    "df = pd.DataFrame(data=d)\n",
+    "print(df.describe())\n",
+    "segmentation_partition = segment_on_column(\"col3\")\n",
+    "segmentation_partition.filter = SegmentFilter(filter_function=lambda df: df.col1 > 49)\n",
```
FelipeAdachi, line 67: segmentation_partition is a dict, so this line breaks. should access the ['col3'] key, perhaps?

jamie256: Yeah good catch I have local changes that include uploading segments to whylabs, let me clean that up and update this example as the refactor updates are in the PR here.

FelipeAdachi (commented):
Segmenting on an Integer column gives me the following error:

`TypeError: 'int' object is not iterable`

Int features are often categorical. should we allow segmenting on it, or should we warn the user to transform it to string before segmenting or something like that?

jamie256:
> Segmenting on an Integer column gives me the following error:
> 
> `TypeError: 'int' object is not iterable`
> 
> Int features are often categorical. should we allow segmenting on it, or should we warn the user to transform it to string before segmenting or something like that?

It was a regression, should be fixed, and I added a test to cover the integer column segmentation case

FelipeAdachi (commented):
Is it possible to segment based on inequalities, such as `Age>18` ?

I understand you can filter it with `filter`, but that would require the data to be segmented on another column, right?

What if I want two segments, one for Age>18 and the other Age<=18?

jamie256:
> Is it possible to segment based on inequalities, such as `Age>18` ?
> 
> I understand you can filter it with `filter`, but that would require the data to be segmented on another column, right?
> 
> What if I want two segments, one for Age>18 and the other Age<=18?

Thats what the custom mapper.map function is for but I don't have examples/tests yet for that scenario, will do in follow up.

python/whylogs/core/schema.py
```diff
@@ -107,6 +110,7 @@ def copy(self) -> "DatasetSchema":
         args = {k: deepcopy(self.__dict__[k]) for k in keys if k not in self.types}
         copy = self.__class__(**args)
         copy._columns = deepcopy(self._columns)
+        copy.segments = self.segments.copy()
```
TheMellyBee, line 113: Does this need to be a deep copy like Dict and Columns?

python/whylogs/core/view/dataset_profile_view.py
```diff
@@ -168,7 +169,9 @@ def read(cls, path: str) -> "DatasetProfileView":
 
             dataset_segment_header = read_delimited_protobuf(f, DatasetSegmentHeader)
             if dataset_segment_header.has_segments:
-                raise DeserializationError("File contains segments. This is not supported yet")
+                logger.warning(
+                    "File contains segments. Only first profile will be deserialized into this DatasetProfileView"
+                )
 
             # TODO: handle segment data
```
TheMellyBee: Is this TODO still accurate?

python/whylogs/core/view/segmented_dataset_profile_view.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def get_default_path(self) -> str:
+        return f"profile_{self._profile_view.creation_timestamp}_{self._segment.parent_id}_{'_'.join(self._segment.key)}.bin"
+
+    def _write_as_v0_message(self, path: Optional[str] = None, **kwargs: Any) -> None:
+        message_v0 = v1_to_dataset_profile_message_v0(self.profile_view, self.segment, self.partition)
+        path = path or self.get_default_path()
+        with open(path, "w+b") as out_f:
+            write_delimited_protobuf(out_f, message_v0)
+            # out_f.write(message_v0.SerializeToString())
```
TheMellyBee: Can this comment be removed?

python/whylogs/core/view/segmented_dataset_profile_view.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def _write_v1(self, path: Optional[str] = None, **kwargs: Any) -> None:
```
TheMellyBee, line 79: This code looks the same as what's in DatasetProfileView, could this get refactored to pull the method out allowing for less duplication?

TheMellyBee: I do see the bit of segments at the bottom. Doesn't need to be in this release, but I'd recommend us extracting some of these as private methods and using them in series instead of this long somewhat complicated method that has a lot of reuse. 

I do see this doesn't inherit from DatasetProfileView. Maybe it should and those can be extracted up, or having an utility for potions of it for both of them

python/whylogs/core/view/segmented_dataset_profile_view.py
```diff
@@ -0,0 +1,184 @@
...
+            logger.warn(f"constructed DatasetProperties for segmented profile file: {properties}")
+
+            dataset_header = DatasetProfileHeader(
+                column_offsets=column_chunk_offsets,
+                properties=properties,
+                length=total_len,
+                indexed_metric_paths=metric_index_to_name,
+            )
+
+            # TODO: multi segment file format requires multiple offset calculations.
```
TheMellyBee, line 148: Is this TODO: still valid? Are we just oding single file segments right now?

python/whylogs/core/view/segmented_dataset_profile_view.py
```diff
@@ -0,0 +1,184 @@
...
+                    f"Writing segmented profile file: whylogs file and segment headers wrote {out_f.tell()} bytes"
+                )
+                write_delimited_protobuf(out_f, dataset_header)
+                logger.debug(
+                    f"Writing segmented profile file: and with dataset header wrote a total of {out_f.tell()} bytes before writing the chunks."
+                )
+
+                f.seek(0)
+                while f.tell() < total_len:
+                    buffer = f.read(1024)
```
TheMellyBee: This is true for here and DatasetProfileView, let's make this not a magic number for easier change.

python/whylogs/migration/converters.py
```diff
@@ -21,8 +22,83 @@
...
+
+PARTITION_ID = "segp_id"
+PARTITION_NAME = "segp_name"
+PARTITION_HAS_FILTER = "segp_filter"
+SEGMENT_ON_COLUMN = "segp_col"
+SEGMENT_ON_COLUMNS = "segp_cols"
+_TAG_PREFIX = "whylogs.tag."
+
+
+def _generate_segment_tags_metadata(
```
TheMellyBee, line 65: Gosh, what a signature. Although true, and the return is fairly readable.. if we're returning 3 related but different dictionaries and lists, why not just go ahead and make it a DataClass?

python/whylogs/migration/converters.py
```diff
@@ -84,19 +180,131 @@ def _extract_type_counts_metric(msg: ColumnMessageV0) -> TypeCountersMetric:
...
+
+
+def _extract_frequent_items_sketch_message_v0(col_prof: ColumnProfileView) -> FrequentItemsSketchMessageV0:
+    frequent_items_metric: FrequentItemsMetric = col_prof.get_metric(FrequentItemsMetric.get_namespace())
+    frequent_items_v0 = FrequentItemsSketchMessageV0(sketch=EMPTY_FI_SKETCH)
+    if frequent_items_metric is None:
+        return frequent_items_v0
+
+    frequent_items_v0.sketch = frequent_items_metric.frequent_strings.value.serialize()
+    frequent_items_v0.lg_max_k = 12
```
TheMellyBee: Why 12?

python/whylogs/migration/converters.py
```diff
@@ -84,19 +180,131 @@ def _extract_type_counts_metric(msg: ColumnMessageV0) -> TypeCountersMetric:
...
+
+def _extract_hll_sketch_message_v0(col_prof) -> HllSketchMessageV0:
+    cardinality_metric: CardinalityMetric = col_prof.get_metric(CardinalityMetric.get_namespace())
+    hll_v0 = HllSketchMessageV0(sketch=EMPTY_HLL)
+    if cardinality_metric is None:
+        return hll_v0
+    hll_sketch = cardinality_metric.hll.value
+    if not hll_sketch.is_empty():
+        hll_v0.sketch = hll_sketch.serialize_compact()
+        hll_v0.lg_k = 12
```
TheMellyBee: Same as above, why 12?

python/examples/advanced/Segments.ipynb
```diff
@@ -0,0 +1,139 @@
+{
```
TheMellyBee, line 1: This looks like it's just the test code without any example or anything. Is this a temporary situation? What's the benefit between putting this in vs having Felipe or I make an example on Monday that is easy to follow?

TheMellyBee (changes requested):
A handful of questions really. Over all looking good.

python/whylogs/core/view/segmented_dataset_profile_view.py
jamie256: correct just single segment files for v1 initial

python/whylogs/migration/converters.py
jamie256: it was the v0 default, I can create constants for these

python/whylogs/migration/converters.py
jamie256: its just so we have consistent values to what v0 used.

python/examples/advanced/Segments.ipynb
jamie256: I can leave it out, or we can update it, just thought showcasing how to run it would help, I actually used this to test upload to whylabs etc, but left those out of the PR.

python/whylogs/migration/converters.py
jamie256: I don't like the signature either, I factored this into a method to reuse the logic, we can refactor better later, but deferring the follow up here if acceptable?

python/whylogs/core/view/segmented_dataset_profile_view.py
jamie256: Agreed, might defer to keep the scope limited to not fixing DatasetProfileView in this PR?

python/whylogs/core/view/segmented_dataset_profile_view.py
jamie256: Fair point.

This is just a container, since I didn't want to modify dataset profile, or create circular references, I just wanted a way to bind the segments and the profile for later serialization. There is some duplication in the interest of reducing risk (so I don't have to modify the core unsegmented scenario too much in DatasetProfile), so the new behavior is isolated to this class.

We can factor out the common parts of the file format so there is less duplication, or proceed with this and iterate?

python/whylogs/core/view/dataset_profile_view.py
jamie256: Not really, thanks!

python/whylogs/core/schema.py
jamie256: If we wanted to support special casing segmentation in subclasses of DatasetSchema, but I don't think we want to support that scenario.

python/whylogs/api/logger/logger.py
jamie256: yeah those are better names, will rename this top level entry point into segmentation processing

python/whylogs/core/view/segmented_dataset_profile_view.py
jamie256: done.

python/whylogs/core/view/segmented_dataset_profile_view.py
jamie256: Updated comment.

python/whylogs/core/view/segmented_dataset_profile_view.py
jamie256: Created a constant for this file, but leaving DatasetProfileView out of this change for now.

python/whylogs/migration/converters.py
jamie256: constants created for all default values.

python/whylogs/core/view/segmented_dataset_profile_view.py
TheMellyBee: I think we're on to proceed and iterate. Maybe we can make an issue for an easy code refactor when someone has bandwidth? :)

python/whylogs/migration/converters.py
TheMellyBee: Yeah I think it's fine to defer to something a little cleaner later

TheMellyBee (approved):

PR #795, opened 2022-09-09
Hardcode en/latest into Announcement URL.  Otherwise is coming up 404.

jghoman:
## Description

Currently the announcement [link](https://whylogs.readthedocs.io/migration/basics.html) on the live site (click on the "whylogs v1 has been launched! Make sure you checkout the [the migration guide](https://whylogs.readthedocs.io/migration/basics.html) to ensure a smooth transition" link) is 404ing because there is no `/en/latest` in the url, which is prepended by readthedocs.  This patch hardcodes the `/en/latest`, which isn't technically correct (since it really should go to the version of the code that you're reading), but is the best we can probably do since this is raw html included in the announcement text and not a Sphinx-link that can be massaged.

Screen Shot 2022-09-09 at 3 17 27 PM

### DIFF
--- a/python/docs/conf.py
+++ b/python/docs/conf.py
@@ -102,7 +102,7 @@ myst_enable_extensions = [
 
 html_theme_options: Dict[str, Any] = {
     "announcement": """<em>whylogs v1</em> has been launched! Make sure you checkout the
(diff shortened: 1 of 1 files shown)


jamie256 (approved):
Thanks Jacob!

PR #796, opened 2022-09-12
Model Performance Metrics port into v1.1 from v0

jamie256:
## Description

Initial port of model performance metrics into v1
* This uses the same protobuf message format from v0 for model performance metrics
* small simplifications, and rephrasing.
* model performance metrics can be added to DatasetProfiles (including segmented profiles), or created as a standalone ModelPerformanceMetrics object.


python/whylogs/core/dataset_profile.py
```diff
@@ -129,6 +147,92 @@ def _do_track(
 
         raise NotImplementedError
 
+    def track_classification_metrics(
```
andyndang: this API is extremely error prone so I'm not sure we should even expose it here as a public API. Probably should add underscore to it at least

python/whylogs/api/logger/__init__.py
```diff
@@ -20,6 +23,94 @@ def log(
     return TransientLogger(schema=schema).log(obj, pandas=pandas, row=row)
 
 
+def log_classification_metrics(
```
andyndang, line 26: Any example of this from end user's perspective?

python/whylogs/core/dataset_profile.py
jamie256: Done.

python/whylogs/api/logger/__init__.py
jamie256: Updated this a bit, but here's two different ways to log model performance metrics with this interface:
1) standalone:
```
import random
import whylogs as why

# some fake example classification data taken from 5 labels
label_set = ['x0', 'x1', 'x2', 'x3', 'x4']
number_of_predictions = 100
example_predictions = [label_set[random.randint(0,4)] for i in range(number_of_predictions)]
# simulate targets being correct more than 85% of the time, then random guess
example_targets = [example_predictions[i] if random.random() > 0.15 else label_set[random.randint(0,4)] for i in range(number_of_predictions)]

# profile result set containing model performance metrics (in an empty DatasetProfile)
model_perf_results = why.log_classification_metrics(targets=example_targets, predictions=example_predictions)

# Because this is a ResultSet, you can write it in the usual ways, e.g.
model_perf_results.writer("whylabs").write()
```

2) Add model performance metrics to an existing DatasetProfile:
```
from whylogs.core.model_performance_metrics import ModelPerformanceMetrics

model_perf = ModelPerformanceMetrics()
model_perf.compute_confusion_matrix(
    predictions=example_predictions,
    targets = example_targets)

first_segment_profile.add_model_performance_metric(model_perf)
```

This is very similar to v0 calling pattern with some minor simplifications.

python/whylogs/core/model_performance_metrics/confusion_matrix.py
```diff
@@ -0,0 +1,244 @@
...
+    def merge(self, other_cm):
+        """
+        Merge two seperate confusion matrix which may or may not overlap in labels.
+
+        Args:
+              other_cm (Optional[ConfusionMatrix]): confusion_matrix to merge with self
+        Returns:
+              ConfusionMatrix: merged confusion_matrix
+        """
+        # TODO: always return new objects
```
TheMellyBee, line 120: Is this still a todo? Looks like you are returning everywhere, does one just need adjustment?

TheMellyBee (changes requested):
One question on a todo, otherwise everything looks good and aligned with our other metrics.

python/whylogs/core/model_performance_metrics/confusion_matrix.py
jamie256: its an older todo, the case where the current labels are empty or where the other_cm is None we don't create a new object.

TheMellyBee (approved):
Thanks!! Looks good.

bernease:
The major thing that I see missing is the ROC curves (using the optional scores). We can add this on as a fast follow, but I wonder if the absence of ROC curves may be causing the front end issues (if still present). In v0, we always had both ROC and confusion matrix for classification tasks.

python/whylogs/api/logger/__init__.py
```diff
@@ -20,6 +23,92 @@ def log(
...
+    targets,
+    predictions,
+    scores=None,
+    target_field=None,
+    prediction_field=None,
+    score_field=None,
```
bernease: These parameters no longer make sense with this API. In v0, we're logging performance metrics on a logger that already has profile(s) attached. So then pointing to a column makes sense. Here, with why.log_classification_metrics(), we don't look for a profile, so these parameters have no purpose.

I'm voting to delete, but the other option is the Matplotlib/seaborn-like API where you can pass in a dataframe plus these parameters.

python/whylogs/api/logger/__init__.py
```diff
@@ -20,6 +23,92 @@ def log(
...
+        passed
+    target_field : str, optional
+        Description
+    prediction_field : str, optional
+        Description
+    score_field : str, optional
+        Description
+    target_field : str, optional
+    prediction_field : str, optional
+    score_field : str, optional
```
bernease: Delete with parameters above. Also duplicates.

python/whylogs/api/logger/__init__.py
```diff
@@ -20,6 +23,92 @@ def log(
...
+    profile = DatasetProfile()
+    profile.add_model_performance_metric(model_performance_metrics)
+    return ProfileResultSet(profile)
+
+
+def log_regression_metrics(
+    targets,
+    predictions,
+    target_field=None,
+    prediction_field=None,
```
bernease: Same as described above, but for regression.

python/whylogs/api/logger/__init__.py
```diff
@@ -20,6 +23,92 @@ def log(
...
+        passed
+    target_field : str, optional
+        Description
+    prediction_field : str, optional
+        Description
+    score_field : str, optional
+        Description
+    target_field : str, optional
+    prediction_field : str, optional
+    score_field : str, optional
```
bernease: Same as above, delete.

python/whylogs/core/dataset_profile.py
```diff
@@ -82,6 +88,18 @@ def add_metric(self, col_name: str, metric: Metric) -> None:
             raise ValueError(f"{col_name} is not a column in the dataset profile")
         self._columns[col_name].add_metric(metric)
 
+    def add_dataset_metric(self, name: str, metric: Metric) -> None:
+        self._metrics[name] = metric
+
+    def add_model_performance_metric(self, metric: ModelPerformanceMetrics) -> None:
```
bernease: Super minor nit, can be skipped: I don't quite understand why this is singular and the `get` method for model_performance_metrics is pluralized. There doesn't seem to be a way to have multiple. Should we always call it `model_performance_metrics`? Or simply `model_performance`?

python/whylogs/api/logger/__init__.py
jamie256: Done! very nice to simplify, agreed, don't need to pull in everything from previous api.

python/whylogs/api/logger/__init__.py
jamie256: done.

python/whylogs/api/logger/__init__.py
jamie256: done.

python/whylogs/api/logger/__init__.py
jamie256: done

python/whylogs/core/dataset_profile.py
jamie256: The v0 class is largely inherited here which was pluralized, so I'd say it should be pluralized consistently.

The v1 future API to add more flexible metrics with `add_metric` I think it ok.

jamie256: fixed.

jamie256:
> The major thing that I see missing is the ROC curves (using the optional scores). We can add this on as a fast follow, but I wonder if the absence of ROC curves may be causing the front end issues (if still present). In v0, we always had both ROC and confusion matrix for classification tasks.

The optional scores default to 1s the same way as they were optional in v0 confusion matric calculations. We still get ROC calculations in the platform.

andyndang (approved):
So far so good! Will probably dig into the API a bit further after this

PR #803, opened 2022-09-12
Fix whylabs writer uncompound on segmented profiles

jamie256:
## Description

We had a merge bug where the _uncompund_dataset_profile method assumed the Writable file is a DatasetProfile/View which is not the case with segments, updated to handle segmented profiles.


TheMellyBee (approved):
LGTM

PR #810, opened 2022-09-13
Writer Examples - fixing environment variables and deprecated parameters

FelipeAdachi:
## Description

This repository needs 

## Changes

## Related


python/examples/integrations/writers/Writing_to_WhyLabs.ipynb
```diff
@@ -89,7 +89,7 @@
       "Enter your WhyLabs Org ID\n",
       "Enter your WhyLabs Dataset ID\n",
       "Enter your WhyLabs API key\n",
-      "Using API Key ID:  \n"
+      "Using API Key ID:  ygG04qE3gQ\n"
```
TheMellyBee, line 92: Is this ok to get hard coded in? Is this something that will change?

python/examples/integrations/writers/Writing_to_WhyLabs.ipynb
```diff
@@ -344,25 +344,14 @@
...
-      "text/plain": [
-       "<Response [200]>"
-      ]
-     },
-     "execution_count": 4,
-     "metadata": {},
-     "output_type": "execute_result"
-    }
-   ],
+   "outputs": [],
```
TheMellyBee, line 349: I'm curious why this had an output before but not now.

TheMellyBee (commented):
Have some questions for you, but otherwise looking fine.

python/examples/integrations/writers/Writing_to_WhyLabs.ipynb
FelipeAdachi: That's ok because we're only printing the first 10 characters of the key, and not the complete thing!

python/examples/integrations/writers/Writing_to_WhyLabs.ipynb
FelipeAdachi: That's actually a change on how the `writer` works: it used to return the call's response, like [200], but it returns `None` now.

I actually feel like it would be nice to have a response to let the user know it went ok. I'll open a discussion for it and maybe open an issue for it.

TheMellyBee: Sounds great

TheMellyBee (approved):
As soon as we get it green, I'd say this is good to go.

PR #811, opened 2022-09-13
Simplify calling pattern for top level performance metrics

### LINKED ISSUE #807: ConfusionMatrix.add hits errors when parameters are pandas.Series
## Description

This is an older bug, but noticed that if you pass in a pandas series the confusion matrix checks for instance of list, and if not (e.g. pandas series or scalar values) the implementation wraps in a list, which then doesn't work with the series passed in.

jamie256:
## Description

After discussing the top level API some, we're going to try to simplify the calling pattern around existing pandas dataframes containing the prediction and targets as columns.

You can still call the non-pandas version by creating a ModelPerformanceMetrics directly and adding the predictions/targets to it.

Bonus here is you also get a dataset profile so uploading the result set won't hit the pitfall where users upload perf metrics with no profiles.

Fixes #807

### DIFF
--- a/python/tests/core/test_model_performance_metrics.py
+++ b/python/tests/core/test_model_performance_metrics.py
@@ -1,5 +1,8 @@
 from logging import getLogger
 
+import pandas as pd
+
+from whylogs import log_classification_metrics, log_regression_metrics
 from whylogs.core.model_performance_metrics import ModelPerformanceMetrics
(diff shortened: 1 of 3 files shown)


FelipeAdachi (approved):
LGTM!

andyndang (approved):
LGTM!

PR #823, opened 2022-09-15
Have Writer's Return a Response

### LINKED ISSUE #820: WhyLabsWriter write() returns None
## Description

The WhyLabsWriter write() used to return the response from the API call, but that changed on the latest updates. Now it returns None.

The method should let the user know whether the write went through.

- System info
  - Operating system and version: 
  - Browser and version: 
- Steps to reproduce:
  - Step
  - Next step

## Related

TheMellyBee:
## Description
WhyLabsWriter returned the response in v0 but currently returns None. Changed the return for write to be a Response or a Boolean as many of the other writers do not return a response., but thought it would be nice for the same reason. 

## Related

Closes  #820

### DIFF
--- a/python/whylogs/api/writer/writer.py
+++ b/python/whylogs/api/writer/writer.py
@@ -2,7 +2,7 @@ import os
 from abc import ABC, abstractmethod
 from typing import Any, Optional, TypeVar, Union
 
-import requests
+import requests  # type: ignore
 
 T = TypeVar("T", bound="Writer")
 
--- a/python/whylogs/api/writer/writer.py
+++ b/python/whylogs/api/writer/writer.py
@@ -1,6 +1,6 @@
 import os
 from abc import ABC, abstractmethod
-from typing import Any, Optional, TypeVar
+from typing import Any, Optional, Tuple, TypeVar
 
 T = TypeVar("T", bound="Writer")
 
@@ -20,7 +20,7 @@ class Writable(ABC):
         pass
 
     @abstractmethod
-    def write(self, path: Optional[str] = None, **kwargs: Any) -> Optional[int]:
+    def write(self, path: Optional[str] = None, **kwargs: Any) -> Tuple[bool, str]:
         pass
 
 
@@ -38,7 +38,7 @@ class Writer(ABC):
         file: Writable,
         dest: Optional[str] = None,
         **kwargs: Any,
-    ) -> Optional[int]:
+    ) -> Tuple[bool, str]:
         pass
 
     @abstractmethod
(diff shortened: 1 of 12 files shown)


python/whylogs/api/writer/writer.py
```diff
@@ -20,7 +22,7 @@ def get_default_path(self) -> str:
         pass
 
     @abstractmethod
-    def write(self, path: Optional[str] = None, **kwargs: Any) -> None:
+    def write(self, path: Optional[str] = None, **kwargs: Any) -> requests.Response:
```
jamie256: Two issues:
1. We can't reference an optional dependency's type ('requests' package is only used in whylogs[whylabs]) in our base abstract classes used across all Writers/Writables
2. I would expect the write() method would have the same return type across Writer and Writable

python/whylogs/api/writer/writer.py
```diff
@@ -1,6 +1,8 @@
 import os
 from abc import ABC, abstractmethod
-from typing import Any, Optional, TypeVar
+from typing import Any, Optional, TypeVar, Union
+
+import requests  # type: ignore
```
jamie256: we can't import this package as a top level import in Writers, it is not a dependency of whylogs and shouldn't be. 'requests' should be specific to the [whylabs] extra and any reference needs to guard against cases where users don't have it (see PIL or pandas import patterns)

jamie256 (changes requested):
Also check the CI failures.

python/whylogs/api/writer/writer.py
TheMellyBee: Got it. So only in the whylabs one then. :) I can change this to an any type for the abstract return which will be able to keep this from being here

python/whylogs/api/writer/writer.py
TheMellyBee: 1. Got it. 
2. That's fair, that makes sense. Hmm ok, well then. One option is that we return Any for all writables. That's a little odd but also kinda understandable. We could return an int for all writes which would be more understandable with the error codes. 

Or a wrapper around the write just for whylabs. I don't like this. 

Or lastly, not add the response which would force use of the private method _do_upload instead. Also do not like. 

My favorite would be the return any or int. Thoughts?

jamie256: `Any` works, or maybe something like `Tuple[bool, str]` if this is just to give success/fail status + some debug / status information. The string part could be: filepath, url, in success case, or error or http status code in the failure case?

TheMellyBee: Didn't see this before latest commit. I had moved it to an Optional[int] where most don't return anything. I can iterate and move it to that. I thought the direct response of the status code would be nice, but I agree that has more information to give

jamie256 (approved):
Looks good! just need to resolve your conflicts

PR #830, opened 2022-09-16
Support configurable exif tags for image logging

richard-rogers:
## Description

Adds `ImageMetricConfig <: MetricConfig` that allows you to specify exif tags to track.


python/tests/extras/test_image_metric.py
```diff
@@ -39,27 +44,45 @@ def image_loader(path: str = None) -> ImageType:
         return img
 
 
+IMAGE_METRIC_CONFIG = ImageMetricConfig(
+    exif_tags={
```
TheMellyBee, line 48: what is an exif tag?

TheMellyBee (approved):
Looks good, just a question that doesn't need to hold up the review.

python/tests/extras/test_image_metric.py
richard-rogers: Exchangeable Image File Format metadata, available in a variety of image formats.

python/whylogs/extras/image_metric.py
```diff
@@ -135,13 +144,15 @@ def zero(cls, config: MetricConfig) -> "ImageMetric":
...
                 submetrics[tag] = FrequentItemsMetric.zero(config)
+        submetrics.update(config.exif_tags)
 
         return ImageMetric(submetrics)
 
 
 def log_image(
     images: Union[ImageType, List[ImageType], Dict[str, ImageType]],
     default_column_prefix: str = "image",
+    config: ImageMetricConfig = ImageMetricConfig(),
```
jamie256, line 155: If we instantiate a default value here, it will be a singleton instance, not sure that is what you want?

python/whylogs/extras/image_metric.py
```diff
@@ -125,6 +131,9 @@ def columnar_update(self, view: PreprocessedColumn) -> OperationResult:
 
     @classmethod
     def zero(cls, config: MetricConfig) -> "ImageMetric":
+        if not isinstance(config, ImageMetricConfig):
```
jamie256, line 134: Can we have a default of None here and instantiate an ImageMetricConfig if not passed in? I think it makes calling zero a lot easier, for example in a reduce call.

python/whylogs/extras/image_metric.py
```diff
@@ -101,6 +102,11 @@ def image_based_metadata(img):
     }
 
 
+@dataclass(frozen=True)
+class ImageMetricConfig(MetricConfig):
+    exif_tags: Dict[str, Metric] = field(default_factory=dict)
```
jamie256, line 107: How does this interact with `get_pil_image_metadata` above? I would expect that this limits the keys we inspect but its not doing that.

I think the way this new ImageMetricConfig behaves is surprising with respect to how its named:
`get_pil_image_metadata` already will log any exif tags found in the images passed in, so I'm confused about what value this configuration provides to users?

Your notebook example shows that your config forces there to be empty metrics for the configured names when the exif tags are empty. Are we not logging those tags if present in the input images?

jamie256 (changes requested):
I'm not sure I understand the scenario this is targeting. Can you show what happens with input images that already contain exif tags (what happens without your changes and what does this change provide)?

jamie256:
> I'm not sure I understand the scenario this is targeting. Can you show what happens with input images that already contain exif tags (what happens without your changes and what does this change provide)?

Thanks for the offline explanation, what I was missing is that there is a limitation in the current compound metrics in that they resolve the set of submetrics once when initialized, hence the dummy image log in zero(). In v0 the exif tags were logged as columns which can be dynamically added to a dataset profile after logging has started.

We should see if we can support dynamically adding submetrics to compound metrics after initialization so that we don't need to configure the set of submetrics before logging. This will take some rework of compound metrics but should simplify what a caller has to configure before logging images.

richard-rogers:
Solved this in a different PR

PR #832, opened 2022-09-16
Update getting started with WhyLabs notebook to v1

jamie256:
## Description

This is an example notebook that was in WhyLabs integrations page, and wanted to move this into whylogs repo.


FelipeAdachi (approved):
looks good!

just two things:

- colab link not pointing to the example
- needs to be skipped as there's user input

jamie256:
> just two things:
> 
> * colab link not pointing to the example
> * needs to be skipped as there's user input

done.

PR #837, opened 2022-09-16
Uncomment pip install so notebook examples mostly just work

### LINKED ISSUE #835: We shouldn't have the pip install commands commented out in our examples
## Description

If a user just runs our notebooks they get errors because the pip install cell is commented out by default with instructions to run if they need. But running pip install when whylogs is already present is fine, and if they don't read the instructions things look broken.

Lets make things not broken in first impression scenarios.

### LINKED ISSUE #838: our read the docs build installs latest poetry which breaks
## Description

Where we install latest versions of things in our actions it leads to breaks unrelated to the changes being tested which can cause confusion.

install specific version of tools in setup scripts and unblock building read the docs. 

jamie256:
## Description

Notebook examples need to work when users run all cells.

## Changes

Uncomment out all `pip install whylogs` commands.

## Related

Fixes #835 also fixes #838

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -7,8 +7,8 @@ build:
   jobs:
     pre_create_environment:
       - asdf plugin add poetry
-      - asdf install poetry latest
-      - asdf global poetry latest
+      - asdf install poetry 1.2.0
+      - asdf global poetry 1.2.0
       - cd python && poetry export -E docs --without-hashes -f requirements.txt --output requirements.txt
       - cd python/docs && make pre-build
 sphinx:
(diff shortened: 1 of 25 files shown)


andyndang (approved):
LGTM!

FelipeAdachi:
I have one concern with this - if we install whylogs on the test environment, we wouldn't be able to check for bugs for the latest commits, only for the stable version. Neiter new features that won't be present when you install it.

Unless the pip install wouldn't override the environment's current whylogs, but I'm not sure what is the behavior in this case.

Any thoughts on how to address this?

TheMellyBee:
This also means that the analytics for these will be done as a download for whylogs every time test notebook runs.

PR #845, opened 2022-09-19
typing-extensions dependency entry update

jamie256:
## Description

typing-extensions dependency entry update


TheMellyBee (approved):
Thanks!

Issue #846, opened 2022-09-19
Segment tags are not being set correctly when there are multiple columns


christinedraper:
I uploaded some segments based on two columns, and the tags are incorrect for the second column. I'm getting `key1=val1&key2=val1`  where val1 is the value of key1 and not key2.  Think this is where it's going wrong:

https://github.com/whylabs/whylogs/blob/59d015927057d914675f632b6a8f4a940c5f8093/python/whylogs/migration/converters.py#L79-L81

PR #848, opened 2022-09-19
Fix multi-column segment key translation

jamie256:
Multi-column segment keys were mistranslated to v0 profiles because we incorrectly duplicated the first columns value in the SegmentTag in subsequent values.

Addresses #846 

* added some tests,
* more logging
* fix catching RuntimeError in deserialization of kll double/float in the translator to make round trip serialization tests work across whylogs versions v0 and v1.


TheMellyBee (approved):
Lgtm!

PR #852, opened 2022-09-20
Adding Profile Store functionality

murilommen:
## Description

This repository needs a Profile store wrapper, that will manage `get` and `write` for structured timestamp-based queries


python/examples/advanced/Profile_Store/LocalStore/generate_data.py
```diff
@@ -0,0 +1,21 @@
...
+
+import pandas as pd
+
+import whylogs as why
+from whylogs.api.store.profile_store import ProfileStore
+
+logger = why.logger(mode="rolling", interval=10, when="S")
+logger.append_store(store=ProfileStore(base_name="test_constraints"))
+
+
```
naddeoa: What happens here

```python
logger = why.logger(mode="rolling", interval=10, when="S")
store = ProfileStore(base_name="test_constraints")
logger.append_store(store=store)

logger.log(some_data)
store.get(config=DateConfig(some_datas_timestamp))
```

I assume it only exists after the logger rolls over. Can people trigger flushes?

python/examples/advanced/Profile_Store/LocalStore/constraints_example.py
```diff
@@ -0,0 +1,23 @@
...
+
+from whylogs.api.store.date_config import DateConfig
+from whylogs.api.store.profile_store import ProfileStore
+from whylogs.core.constraints import ConstraintsBuilder
+from whylogs.core.constraints.factories import greater_than_number
+
+# 1. Configure Store
+store = ProfileStore(base_name="test_constraints")
+now_config = DateConfig(start_date=datetime.now())
+reference_config = DateConfig(start_date=datetime.now() - timedelta(days=7), end_date=datetime.now())
```
naddeoa: Rather than a config, it might make more sense to call this something like a `StoreQuery`. It isn't really configuring anything, its just a way to ask for profiles right? Then you would have `DateQuery`, `ProfileIdQuery`, etc.

python/examples/advanced/Profile_Store/LocalStore/constraints_example.py
```diff
@@ -0,0 +1,23 @@
...
+
+# 2. Get the profile
+profile_view = store.get(date_config=now_config)
+reference_profile = store.get(date_config=reference_config)
```
naddeoa, line 15: Is it appropriate to use polymorphism here in python? The `date_config` would just be `config` (or `query` if my other comment) and people would just pass in whatever type of `StoreQuery` they want to use.

python/whylogs/api/logger/rolling.py
```diff
@@ -126,6 +129,10 @@ def __init__(
     def check_writer(self, writer: Writer) -> None:
         writer.check_interval(self.interval)
 
+    def append_store(self, store: ProfileStore) -> None:
+        if store is not None:
+            self.store = store
+
```
naddeoa: This looks more like a `set_store` than an `append_store`. That would imply we maintain a list of stores. I think its actually an ok idea to have a list of stores but it doesn't look like that's in yet. What was the intent?

python/whylogs/api/logger/rolling.py
```diff
@@ -153,6 +160,23 @@ def _do_rollover(self) -> None:
...
+            if dst_now != dst_then:
+                if dst_now:
+                    addend = 3600
+                else:
+                    addend = -3600
+                time_tuple = time.localtime(self._current_batch_timestamp + addend)
+        return time_tuple
+
     def _flush(self, profile: DatasetProfile) -> None:
         if profile is None:
```
naddeoa, line 174: oh I guess there is a flush, but it's private. Any reason we shouldn't just expose this and let people know that its ok to call?

python/whylogs/api/store/profile_store.py
```diff
@@ -1,51 +1,26 @@
...
-        segment: Optional[Segment] = None,
-        schema: Optional[DatasetSchema] = None,
-    ) -> None:
+    def get(self, date_config: DateConfig) -> DatasetProfileView:
         pass
 
     @abstractmethod
-    def query(self, *args: Any, **kwargs: Any) -> List[DatasetProfile]:
+    def write(self, profile: Optional[DatasetProfileView]) -> None:
         pass
```
naddeoa, line 16: I guess you were thinking about `query` at one point. Maybe it would make more sense to have `get` be key based and then allow for a separate `query` api that lets you query based on date ranges or other properties? Might even make sense to make it 

```python
def query(self, query: List[StoreQuery]) -> List[DatasetProfile]
```

Then the api would allow for people to specify multiple criteria, which would allow them to do `or` with multiple date ranges.

python/whylogs/api/store/local.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class LocalStore(ProfileStore):
+    """
+    The LocalStore is the implementation of the base :class:ProfileStore
+    that will manage reading and writing profiles on the local file system.
+
+    In order to instantiate the object, you will need to define a base_name,
+    which is related to the name of the model or dataset you're profiling.
+    To properly use the LocalStore to generate files, you should append it
```
naddeoa: Can we add how the `base_name` is used? Is it just something to use in file path names?

python/whylogs/api/store/local.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+
+    def __init__(self, base_name: str):
+        self.base_name = base_name
+        self._default_path = os.path.join(os.getcwd(), DEFAULT_DIR)
+        logger.debug(f"Default ProfileStore path set to {self._default_path}")
+        self._writer = LocalWriter(base_dir=self._default_path)
+
+        if not os.path.isdir(os.path.join(self._default_path, self.base_name)):
+            os.makedirs(os.path.join(self._default_path, self.base_name))
```
naddeoa: might just move this to the write method so it always happens before writes and it doesn't happen if nothing is ever written. I can imagine doing some local tests on a loop or something, manually deleting the profile dir to test something, then having to restart the app because it can't remake the dir outside of startup.

python/whylogs/api/store/local.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def _read_profile(self, date_config: DateConfig) -> Optional[DatasetProfileView]:
+        dates_list = self.get_dates_list(date_config=date_config)
+        files_list = []
+        base_directory = os.path.join(self._default_path, self.base_name)
+        for date in dates_list:
+            files_list.extend(glob(f"{base_directory}/profile_{date}*.bin"))
+        if len(files_list) > 0:
+            logger.debug(f"Profiles found! Number of profiles is {len(files_list)}")
+            profiles_list = [why.read(file).view() for file in files_list]
```
naddeoa: Probably ok for now but I assume we should be returning some sort of wrapper around the files rather than just reading them all at once. There could be a lot of them, or a small number of large ones.

naddeoa: I guess this comment applies more for the API that just returns views, Noticed that line below does the merge. A similar comment applies to the merge though: we can read them one at a time and merge them right away to avoid memory issues, which would just mean updating `x.merge` to `read && x.merge`

python/whylogs/api/store/local.py
```diff
@@ -0,0 +1,103 @@
...
+        self._writer = LocalWriter(base_dir=self._default_path)
+
+        if not os.path.isdir(os.path.join(self._default_path, self.base_name)):
+            os.makedirs(os.path.join(self._default_path, self.base_name))
+
+    @staticmethod
+    def _get_profile_timestamp() -> str:
+        now = datetime.utcnow()
+        return f"profile_{now.date()}_{now.hour}:{now.minute}:{now.second}_{random_string()}.bin"
+
```
naddeoa: More like `_get_profile_filename` right? Might be a python question here, why bother making private methods static rather than just making this a function in the module above?

python/whylogs/api/store/local.py
```diff
@@ -0,0 +1,103 @@
...
+            logger.debug(f"Could not find any profiles at {base_directory}")
+            return None
+
+    def get(self, date_config: DateConfig) -> DatasetProfileView:
+        logger.debug("Fetching profiles with specified DateConfig...")
+        profile = self._read_profile(date_config=date_config)
+        return profile
+
+    def write(self, profile: Optional[DatasetProfileView]) -> None:
+        profile_dest = os.path.join(self._default_path, self.base_name, self._get_profile_timestamp())
```
naddeoa: Why `Optional`?

python/examples/advanced/Profile_Store/LocalStore/generate_data.py
murilommen: i assume that it should only flush based on the config of the rolling logger. 
why would you want to "instantly" flush instead of rolling every X seconds?

python/whylogs/api/store/profile_store.py
murilommen: actually the `query` method was from Jamie's stubs 😄 
so if I have 2 profiles written today, using `base_name-{timestamp}-{random_string}`, how can I refer to it in order to get it? since we discussed that `id` would only be the `base_name` at the moment

python/whylogs/api/logger/rolling.py
murilommen: i think i commented above, maybe let's leave this discussion just there

python/whylogs/api/store/local.py
murilommen: my bad :( haha

python/whylogs/api/store/local.py
murilommen: renamed it! 
about the private vs. function, i rather keep this method inside the class because I only want local store to use this, not other stuff. and since it doesn't rely on any of the object's attrs, i made it static.

python/examples/advanced/Profile_Store/LocalStore/constraints_example.py
murilommen: yes! makes total sense

python/whylogs/api/logger/rolling.py
murilommen: yeah, the intent was to enable the same behavior of the writers, to write for every store on the store_list. will make sure to fix that!

python/examples/advanced/Profile_Store/LocalStore/generate_data.py
murilommen: and users could always use the regular LocalStore api, as in:
```python
store = LocalStore(base_name="test_constraints")
store.write(profile=some_profile)

store.get(config=DateConfig(some_datas_timestamp))
```

murilommen: i'm here assuming `logger.log()` and `store.get()` will happen on different processes. 1) capture 2) compare

python/examples/advanced/Profile_Store/LocalStore/constraints_example.py
murilommen: only thing it is configuring is the `start_date` `end_date` logic, but I agree. only thing is that I'm thinking of another possible Query - `id` at the moment would basically take in the profile's name, but it doesn't share anything in common with the `DateQuery`, right? 
my point being: what defines a base `StoreQuery`?

murilommen: to answer myself, I think I got somewhere:
```python
query_date = DateQuery(start, end)
query_id = ProfileIdQuery(id)
# ---
store.get(any_query: StoreQuery)
```
internally, `store.get` will always call `StoreQuery.query()`, which is the abstract method. sounds good?

python/whylogs/api/store/local.py
murilommen: adding in a footer note on `base_name`. check if it explains better what it's meant to be

python/whylogs/api/store/local.py
murilommen: as easy as changing it to parenthesis. sometimes i really love Python :D

python/whylogs/api/store/query.py
```diff
@@ -0,0 +1,18 @@
+from dataclasses import dataclass, field
+from datetime import datetime
+from typing import Optional
+
+
+@dataclass
+class DateQuery:
+    start_date: datetime
+    end_date: Optional[datetime] = field(default=None)
+
```
naddeoa: does no end_date mean implicitly end_date=now, or does it turn into a single date picker?

python/whylogs/api/store/local_store.py
```diff
@@ -0,0 +1,111 @@
...
+from whylogs.api.writer.local import LocalWriter
+from whylogs.core import DatasetProfileView
+
+DEFAULT_DIR = "profile_store/"
+logger = logging.getLogger(__name__)
+
+# TODO how to prevent users from creating a huge list?
+# should the profile store be aware of its min/max dates?
+
+
```
naddeoa: I think we can get around this by avoiding loading everything into memory at once if the concern is just not letting the host go oom. To do that, we'd have to have make it possible for the store to use that yield iterator in the reduce.

python/whylogs/api/store/profile_store.py
```diff
@@ -1,51 +1,28 @@
...
-    def write(
-        self,
-        data: Writable,
-        model_id: Optional[str] = None,
-        dataset_timestamp: Optional[datetime] = None,
-        segment: Optional[Segment] = None,
-        schema: Optional[DatasetSchema] = None,
-    ) -> None:
+    def get(self, query: DateQuery) -> DatasetProfileView:
         pass
```
naddeoa, line 16: Just occurred to me, but I guess we don't have to worry too much about having a common base query type, we can just use `Union[DateQuery, ProfileIdQuery, etc]`

naddeoa: That lets us handle each one differently internally without shoehorning them all into returning some common type

python/whylogs/api/store/profile_store.py
```diff
@@ -1,51 +1,28 @@
...
+
+    @staticmethod
+    def _get_date_range(query: DateQuery):
+        for n in range(int((query.end_date - query.start_date).days) + 1):
+            yield query.start_date + timedelta(n)
+
+    # TODO return datetimes instead of strings
+    def _get_dates(self, query: DateQuery) -> List[str]:
+        dates = []
+        for single_date in self._get_date_range(query=query):
```
naddeoa: Not sure if we're waiting to do this or not?

python/whylogs/api/store/query.py
murilommen: it goes on the post_init method. if no end_date, `end_date == start_date`
also note there that if `end_date < start_date`, then i switch them

python/whylogs/api/store/local_store.py
murilommen: it's using it already :D

murilommen: the only catch is that we are still building the `n_days` list in memory

python/whylogs/api/store/profile_store.py
murilommen: sounds good! i will resolve this for now and when we have a more mature idea on a `ProfileIdQuery` class we will add an `Union` type there ;)

python/whylogs/api/store/profile_store.py
murilommen: yes, we should! my mistake to leave this behind, but luckily you caught it

python/examples/advanced/Profile_Store/LocalStore/LocalStore_with_Constraints.ipynb
```diff
@@ -0,0 +1,1042 @@
...
+    "\n",
+    "store = LocalStore()\n",
+    "store.list()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "And since we have an empty list returned, it means that we haven't used the profile store in this location so far. But we can already check that a new directory called `profile_store` was created on our working directory:"
```
FelipeAdachi, line 42: minor thing: there's an element in the list that should be empty: `base_model_name`

murilommen: good catch! will re-run things to clear it up :)

python/whylogs/api/store/profile_store.py
jamie256: we can add query later, starting with this tighter scope looks reasonable.

python/whylogs/api/store/utils/random_strings.py
```diff
@@ -0,0 +1,7 @@
+import random
+import string
+
+
+def random_string(length: int = 7):
+    letters = string.ascii_lowercase
```
jamie256: do the filenames need to be readable and short?

I'm concerned this could generate inappropriate/offensive filenames, and if this is not for human readability, then I'd prefer something like a uuid string here instead.

jamie256 (commented):
Can we change the random file name generation a bit?

python/whylogs/api/store/utils/random_strings.py
murilommen: done!

jamie256 (approved):
LGTM!

Issue #851, opened 2022-09-20
post release create pull request action fails


jamie256:
## Description

see https://github.com/whylabs/whylogs/actions/runs/3092683669/jobs/5004282226#step:7:511

jamie256:
still occuring: https://github.com/whylabs/whylogs/actions/runs/3138237949/jobs/5097439689#step:7:510

jamie256:
python/examples/integrations/flask_streaming/.dockerignore

PR #853, opened 2022-09-20
Move .dockerignore to root

jamie256:
## Description

Previous mitigation fix for #851 was not in the root of the repo, moving there.


TheMellyBee (approved):
LGTM!

PR #855, opened 2022-09-21
update poetry to have requests as part of whylabs

### LINKED ISSUE #804: requests package not described as dependency for whylogs[whylabs]
## Description

if you don't have the requests package and are using the whylabs package you will hit errors when making calls to whylabs using the client. The bug is missing dependency in whylabs-client package, but we should mitigate this in whylogs.

TheMellyBee:
Closes #804 

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -53,7 +53,7 @@ Pillow = {version = "^9.2.0", optional = true}
 
 # Fugue related dependencies
 fugue = {version = "^0.7.2", optional = true}
-requests = "^2.28.1"
+requests = "^2.27"
 
 [tool.poetry.dev-dependencies]
 2to3 = "^1.0"
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -53,6 +53,7 @@ Pillow = {version = "^9.2.0", optional = true}
 
 # Fugue related dependencies
 fugue = {version = "^0.7.2", optional = true}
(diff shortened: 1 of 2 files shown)


python/pyproject.toml
```diff
@@ -53,6 +53,7 @@ Pillow = {version = "^9.2.0", optional = true}
 
 # Fugue related dependencies
 fugue = {version = "^0.7.2", optional = true}
+requests = "^2.28.1"
```
jamie256: this is forcing people to update their version of requests to a build that was released on June 29, 2022 or later, do we need to do be that strict? If a user has an older version of requests already, this entry says whylogs won't work unless they update, but that is probably not true.

I'd suggest change the version number here to allow at least as old as requests 2.27.0 since that still supported python 3.6 and has python 3.10 support, but probably we could support very old versions as we aren't using advanced features of this dependency:
https://requests.readthedocs.io/en/latest/community/updates/#id4

jamie256 (changes requested):

python/pyproject.toml
TheMellyBee: That's helpful thank you! I just did poetry add requests and totally missed it put the newest which makes sense. I'll change this and run an update. :)

jamie256: Thanks!

jamie256 (approved):
Looks good!

PR #854, opened 2022-09-20
Update whylabs-client version for feature importance

jamie256:
* Add updated dependency on whylabs-client 0.4.0.dev0 
* regenerate poetry.lock file

## Description

Tested uploading whylogs profiles. This change allows whylogs to integrate with new whylabs-client API for feature importance upload.

- [x] I have reviewed the [Guidelines for Contributing](CONTRIBUTING.md) and the [Code of Conduct](CODE_OF_CONDUCT.md).

PR #857, opened 2022-09-21
feature importance writers API

FelipeAdachi:
## Description

- add PUT feature weights to WhyLabs Writer's API
- add GET feature weights to WhyLabs Writer's API

Notes:
this should be merged after https://github.com/whylabs/whylogs/pull/854
## Changes

## Related


python/whylogs/core/feature_weights.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class FeatureWeightResponse(TypedDict):
+    segmentWeights: list
+    metadata: Dict
+
+
+class FeatureWeightDict(TypedDict):
+    segment: Optional[str]
+    weights: Dict[str, float]
```
jamie256: In a core class, I don't think we should have separate classes for response vs the conceptual object. What about having a single class be named `FeatureWeights`?

python/whylogs/core/feature_weights.py
```diff
@@ -0,0 +1,76 @@
...
+from whylogs.api.writer.writer import Optional, Writable, Writer
+
+
+class FeatureWeightResponse(TypedDict):
+    segmentWeights: list
+    metadata: Dict
+
+
+class FeatureWeightDict(TypedDict):
+    segment: Optional[str]
```
jamie256: Can you make this an `Optional[Segment]`? The segment key is a Tuple of strings, so the single string will make this hard to support multi-column segments.

python/whylogs/core/feature_weights.py
```diff
@@ -0,0 +1,76 @@
...
+    weights: Dict[str, float]
+
+
+class FeatureWeight(Writable):
+    def __init__(self, weights: Dict[str, float], segment: Optional[str] = None):
```
jamie256: segment should be of type `Optional[Segment]`

python/whylogs/core/feature_weights.py
```diff
@@ -0,0 +1,76 @@
...
+    segmentWeights: list
+    metadata: Dict
+
+
+class FeatureWeightDict(TypedDict):
+    segment: Optional[str]
+    weights: Dict[str, float]
+
+
+class FeatureWeight(Writable):
```
jamie256: suggest pluralize this name since it contains a collection of weights? Do we need the other classes above if you have this one?

python/whylogs/core/feature_weights.py
```diff
@@ -0,0 +1,76 @@
...
+        pass
+
+    def to_json(self) -> str:
+        return json.dumps({"segment": self.segment, "weights": self.weights})
+
+    def to_dict(self) -> FeatureWeightDict:
+        return {"segment": self.segment, "weights": self.weights}
+
+
+class FeatureWeightWriter(object):
```
jamie256: Did you intend to inherit from `Writer`?

python/whylogs/api/writer/whylabs.py
```diff
@@ -237,8 +271,47 @@ def _do_upload(
...
+
+        Returns
+        -------
+        FeatureWeightResponse
+            Response of the GET request, with segmentWeights and metadata.
+        """
+        self._validate_org_and_dataset()
+        self._validate_api_key()
+
+        result = self._get_column_weights()
```
jamie256, line 291: I think it would be ok to translate the response type back into a `FeatureWeights` here. Better to encapsulate the whylabs-client integration and the use of the auto generated service api types so that when the generated classes change we only have to update the code in this file.

python/whylogs/api/writer/whylabs.py
```diff
@@ -174,8 +177,43 @@ def option(  # type: ignore
...
+        -------
+        Tuple[bool, str]
+            Tuple with a boolean (1-success, 0-fail) and string with the request's status code.
+        """
+        self._feature_weights = file.to_dict()
+        if kwargs.get("dataset_id") is not None:
+            self._dataset_id = kwargs.get("dataset_id")
+        return self._do_upload_feature_weights()
+
+    def get_feature_weights(self, **kwargs: Any) -> FeatureWeightResponse:
```
jamie256: suggest we return a `FeatureWeights` here.

Also we might reconsider the name of this class now that its not only writing (outside this PR).

python/whylogs/core/feature_weights.py
FelipeAdachi: so we're assuming that the songbird API will change? currently it's a string.

python/whylogs/core/feature_weights.py
FelipeAdachi: The TypedDict ones actually are mostly there to help with annotation. I could simply remove them and change the annotation.

python/whylogs/api/writer/whylabs.py
FelipeAdachi: Ok, we should probably tweak the code to include version info, then? I think that's an important information for the user to get.

python/whylogs/api/writer/whylabs.py
FelipeAdachi: We might need another class, because the GET request will actually return a list of  `FeatureWeights`, along with metadata information for the whole list.

Maybe making another class, like `FeatureWeightsSet` or something?

python/whylogs/core/feature_weights.py
jamie256: Right, and if you get more than one string for the Segment key, then we'd have to figure out how to translate that to the single string value, but better to keep our API consistent in the whylogs side and do the translations right before upload if needed.

python/whylogs/core/feature_weights.py
jamie256: Remove or prefix with underscore if they are internal helpers

python/whylogs/api/writer/whylabs.py
jamie256: You could include a `metadata: Dict` here, its also interesting to be able to set this when writing but maybe that's not supported yet.

python/whylogs/api/writer/whylabs.py
jamie256: Is it because of segments? Why is the GET returning more than one `FeatureWeights`?

python/whylogs/core/feature_weights.py
FelipeAdachi: Got it. So I'll make it an Optional[Segment], but raise an error if it's actually passing a Segment for now, since that wouldn't work with Songbird at the moment (and Segments is currently not supported by the platform)

python/whylogs/api/writer/whylabs.py
FelipeAdachi: yes, because of segments.

FelipeAdachi: Maybe the easiest thing to do for now is to return FeatureWeights by retrieving the first element of the list and, once segmented feature weights are supported, revisit this. What do you think @jamie256 ?

python/whylogs/api/writer/whylabs.py
FelipeAdachi: added metadata: Dict to FeatureWeights

python/whylogs/api/writer/whylabs.py
FelipeAdachi: With the last commit this is being done: get_feature_weights is retrieving the first element of the list. If we stick to it, we should revisit this once segmented feature weights is supported.

jamie256 (approved):
Looks great!

python/whylogs/api/writer/whylabs.py
jamie256: Sounds good, its probably just returning a dictionary of these keyed by segment then.

PR #859, opened 2022-09-21
Speed up confusion matrix calculation and add test

jamie256:
## Description

The initial integration of confusion matrix called PreprocessedColumn.apply on each entry which results in poor performance as that method is optimized for arrays.

The change here is to prebatch the confusion matrix entries, this works particular well with binary classification (or any confusion matrix with relative small amount of entries). The speed up on 1000 predictions is about 100x.


FelipeAdachi (approved):
LGTM!

bernease (approved):
LGTM as well.

PR #860, opened 2022-09-21
Support adding submetrics to an existing compound metric

richard-rogers:
- Generalize compound metric `merge()` to support merging metrics with non-empty symmetric set difference of submetrics
- Dynamically add EXIF tag submetrics as they're discovered in `ImageMetric`

### DIFF
--- a/python/poetry.lock
+++ b/python/poetry.lock
@@ -99,7 +99,7 @@ tests_no_zope = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy 
 
 [[package]]
 name = "autoflake"
-version = "1.6.0"
+version = "1.6.1"
 description = "Removes unused imports and unused variables"
 category = "dev"
 optional = false
@@ -184,14 +184,14 @@ dev = ["Sphinx (==4.3.2)", "black (==22.3.0)", "build (==0.8.0)", "flake8 (==4.0
 
 [[package]]
 name = "boto3"
-version = "1.24.77"
+version = "1.24.78"
 description = "The AWS SDK for Python"
 category = "main"
 optional = false
 python-versions = ">= 3.7"
 
 [package.dependencies]
-botocore = ">=1.27.77,<1.28.0"
+botocore = ">=1.27.78,<1.28.0"
 jmespath = ">=0.7.1,<2.0.0"
 s3transfer = ">=0.6.0,<0.7.0"
 
@@ -200,7 +200,7 @@ crt = ["botocore[crt] (>=1.21.0,<2.0a0)"]
 
 [[package]]
 name = "botocore"
-version = "1.27.77"
+version = "1.27.78"
 description = "Low-level, data-driven core of boto 3."
 category = "main"
 optional = false
@@ -1149,7 +1149,7 @@ webpdf = ["pyppeteer (>=1,<1.1)"]
 
 [[package]]
 name = "nbformat"
-version = "5.5.0"
+version = "5.6.0"
 description = "The Jupyter Notebook format"
 category = "main"
 optional = false
@@ -1158,7 +1158,7 @@ python-versions = ">=3.7"
 [package.dependencies]
 fastjsonschema = "*"
 jsonschema = ">=2.6"
-jupyter_core = "*"
+jupyter-core = "*"
 traitlets = ">=5.1"
 
 [package.extras]
@@ -1680,7 +1680,7 @@ python-versions = ">=3.6"
 
 [[package]]
 name = "pyzmq"
-version = "24.0.0"
+version = "24.0.1"
 description = "Python bindings for 0MQ"
 category = "main"
 optional = false
@@ -2128,7 +2128,7 @@ widechars = ["wcwidth"]
 
 [[package]]
 name = "tenacity"
-version = "8.0.1"
+version = "8.1.0"
 description = "Retry code until it succeeds"
 category = "dev"
 optional = false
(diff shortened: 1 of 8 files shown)


TheMellyBee:
Just put the tag in so I can see it easier on the PR screen. Let me know when you're ready for a review.

TheMellyBee (commented):
Looking good so far. I'll check back in as it becomes ready.

python/whylogs/extras/image_metric.py
```diff
@@ -101,11 +102,32 @@ def image_based_metadata(img):
     }
 
 
+@dataclass(frozen=True)
+class ImageMetricConfig(MetricConfig):
+    allowed_exif_tags: Set[str] = field(default_factory=set)
+    forbidden_exif_tags: Set[str] = field(default_factory=set)
+
+
+def _trackable_exif_tag(exif_tag: str, allowed_tags: Set[str], forbidden_tags: Set[str]) -> bool:
```
TheMellyBee: Doesn't this need to be tabbed in to be under the class? the ones before are tabbed in so I assume this was supposed to be apart of the class not on its own?

python/whylogs/extras/image_metric.py
```diff
@@ -118,15 +140,28 @@ def columnar_update(self, view: PreprocessedColumn) -> OperationResult:
...
 
     @classmethod
     def zero(cls, config: MetricConfig) -> "ImageMetric":
+        if not isinstance(config, ImageMetricConfig):
+            logger.warning("ImageMetric.zero() needs an ImageMetricConfig")
+            config = ImageMetricConfig()
+
         dummy_image = new_image("HSV", (1, 1))
-        metadata = get_pil_image_metadata(dummy_image)
+        metadata = get_pil_image_metadata(dummy_image)  # any EXIF tags will be discovered as images are logged
```
TheMellyBee, line 163: Wait read this again, why are you putting a dummy image through?

python/whylogs/extras/image_metric.py
richard-rogers: Stylistically I liked this better than `self._trackable_exif_tag(tag_name)`. If it starts growing a lot more arguments that are members of the `ImageMetric` I'd probably change my mind.

richard-rogers: I might be persuaded that `self._wants_to_track(tag_name)` reads nicely...

python/whylogs/extras/image_metric.py
richard-rogers: It gives a nice SPOT for the image stats we track for all images (versus the data-dependent EXIF tags we discover as images are logged).

python/poetry.lock
```diff
@@ -99,7 +99,7 @@ tests_no_zope = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy
 
 [[package]]
 name = "autoflake"
-version = "1.6.0"
+version = "1.6.1"
```
jamie256, line 102: Why is a poetry.lock update in this PR? Do you need this? There are no updates to the install time dependencies in pyproject.toml

python/whylogs/extras/image_metric.py
```diff
@@ -101,32 +102,65 @@ def image_based_metadata(img):
     }
 
 
+@dataclass(frozen=True)
+class ImageMetricConfig(MetricConfig):
+    allowed_exif_tags: Set[str] = field(default_factory=set)
+    forbidden_exif_tags: Set[str] = field(default_factory=set)
```
jamie256, line 108: This capability might be interesting, but I don't know of anyone requesting this feature and it complicates the calling pattern. I thought you only had explored using ImageMetricConfig as a work-around to having no support for logging exif tags like we did in v0, and not having ability to copy metrics in a merge. Why do we need ImageMetricConfig?

python/whylogs/extras/image_metric.py
jamie256: If you can add to the set of submetrics as you log actual images, I don't think you need the dummy image any more, right?

python/whylogs/extras/image_metric.py
```diff
@@ -101,32 +102,65 @@ def image_based_metadata(img):
...
         count = 0
         for image in list(chain.from_iterable(view.raw_iterator())):
             if isinstance(image, ImageType):
                 metadata = get_pil_image_metadata(image)
                 for attr in metadata.keys():
+                    if attr not in submetric_names:  # EXIF tag discovery
+                        if self._wants_to_track(attr):
+                            # TODO: add a resolver to ImageMetricConfig
+                            if isinstance(metadata[attr], int) or isinstance(metadata[attr], float):
+                                self.submetrics[attr] = DistributionMetric.zero(MetricConfig())
```
jamie256, line 146: also TODO for removing MetricConfig here if its not needed (or just remove it now).

python/whylogs/extras/image_metric.py
```diff
@@ -101,32 +102,65 @@ def image_based_metadata(img):
...
+                                self.submetrics[attr] = FrequentItemsMetric.zero(MetricConfig())
+
                     if attr in submetric_names:
                         self.submetrics[attr].columnar_update(PreprocessedColumn.apply([metadata[attr]]))
+
                 count += 1
         return OperationResult.ok(count)
 
     @classmethod
     def zero(cls, config: MetricConfig) -> "ImageMetric":
```
jamie256, line 157: why is `config` required here?

python/whylogs/extras/image_metric.py
```diff
@@ -101,32 +102,65 @@ def image_based_metadata(img):
...
+        return exif_tag in self._allowed_exif_tags
+
     def columnar_update(self, view: PreprocessedColumn) -> OperationResult:
         submetric_names = self.submetrics.keys()
         count = 0
         for image in list(chain.from_iterable(view.raw_iterator())):
             if isinstance(image, ImageType):
                 metadata = get_pil_image_metadata(image)
                 for attr in metadata.keys():
+                    if attr not in submetric_names:  # EXIF tag discovery
```
jamie256, line 142: This is the key new behavior we need: discover and track features when found in images.

python/whylogs/extras/image_metric.py
```diff
@@ -101,32 +102,65 @@ def image_based_metadata(img):
...
+                        if self._wants_to_track(attr):
+                            # TODO: add a resolver to ImageMetricConfig
```
jamie256, line 144: Right, basically we have to resolve the set of metrics here similarly to how the metrics are resolved when a new column is logged.

jamie256 (changes requested):
It would help to have some updates example an notebook with these kinds of changes for context and to help test.

I'm not sure we need (or I am missing context on the motivation) some parts of this PR's changes: ImageMetricConfig? poetry.lock? Can you explain or focus the changes? Maybe ImageMetricConfig can be in this PR and ignored by callers since its optional.

The change that enables users to log images and have newly discovered submetrics included without having to specify them ahead of logging is the key new thing here. This will support the logging pattern we had in v0 for images.

Mostly questions and some minor cleanup, otherwise approach looks good.

richard-rogers:
Merged via another PR with additional work on `ImageMetric`

PR #863, opened 2022-09-22
MetricConfig was supposed to be an optional argument to Metric::zero()

richard-rogers:
This PR makes the config an optional argument to `zero()` and tries to cope reasonably with its absence.


jamie256 (approved):
LGTM!

FelipeAdachi (approved):
LGTM!

PR #867, opened 2022-09-22
Implements Profiles & Views in Java

TheMellyBee:
## Description
Brings in the DatasetProfile/View and ColumnProfile/View. Also creates the StandardResolver and small changes to Metric to allow for needed methods. 


java/core/src/main/java/com/whylogs/core/DatasetProfile.java
```diff
@@ -0,0 +1,141 @@
...
+import java.time.ZonedDateTime;
+import java.util.*;
+import lombok.Getter;
+import lombok.ToString;
+
+// TODO: extend WRITABLE interface
+@Getter
+@ToString
+public class DatasetProfile {
+  // TODO: Time zone is all mixed up. Fix
```
TheMellyBee: this is mostly fixed, but will totally fix in next iteration

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
```diff
@@ -0,0 +1,141 @@
...
+  public static int _LARGE_CACHE_SIZE_LIMIT = 1024 * 100;
+
+  private DatasetSchema schema;
+  // QUESTION: Do we need zones here? Do we just use UTC?
```
TheMellyBee: @jamie256 checking in to see do we need any zones?

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
```diff
@@ -0,0 +1,141 @@
...
+
+  public void addMetric(String colName, Metric metric) {
+    if (!this.columns.containsKey(colName)) {
+      throw new InputMismatchException("Column name not found in schema");
+    }
+    this.columns.get(colName).addMetric(metric);
+  }
+
+  /*
+  TODO: I don't beleive we need this in Java? (with the T Object)
```
TheMellyBee: @jamie256 I'm not sure we would need this in Java with the T Object as we would just be tracking Objects if it's unknown?  

What are your thoughts?

java/core/src/main/java/com/whylogs/core/schemas/ColumnSchema.java
```diff
@@ -0,0 +1,24 @@
...
+import com.whylogs.core.resolvers.Resolver;
+import java.lang.reflect.Type;
+import java.util.HashMap;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class ColumnSchema {
+  // Thoughts: we could have this ColumnSchema<T> instead of having it as a member
```
TheMellyBee, line 14: Here is a thought around passing in the types. Not super important right now as it will become more clear what's easier as I'm testing out the api.

TheMellyBee (commented):
Some questions or points of interest for @jamie256 or @richard-rogers

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
```diff
@@ -0,0 +1,141 @@
...
+  public <T> void track(HashMap<String, T> row) {
+    try {
+      this.isActive = true;
+      this.trackCount += 1;
+      this.doTrack(row);
+    } finally {
+      this.isActive = false;
+    }
+  }
+
```
naddeoa: What's the point of `isActive`? It doesn't look like `track` is thread safe and `isActive` isn't being used towards that end. Is `track` supposed to be thread safe?

java/core/src/main/java/com/whylogs/core/views/ColumnProfileView.java
```diff
@@ -0,0 +1,129 @@
...
+    HashSet<String> allMetricNames = new HashSet<>();
+    allMetricNames.addAll(this.metrics.keySet());
+    allMetricNames.addAll(otherView.metrics.keySet());
+
+    HashMap<String, Metric> mergedMetrics = new HashMap<>();
+    for (String metricName : allMetricNames) {
+      Metric thisMetric = this.metrics.get(metricName);
+      Metric otherMetric = otherView.metrics.get(metricName);
+
+      Metric result = thisMetric;
```
naddeoa: My java is rusty, but I think you can get away without needing to make copies of the things inside of the views with streams

```java
        List<Integer> l1 = new ArrayList<>();
        List<Integer> l2 = new ArrayList<>();

        Stream.concat(l1.stream(), l2.stream()).forEach( i ->{

        });
```

I also assume that the views are all readonly .

java/core/src/main/java/com/whylogs/core/views/DatasetProfileView.java
```diff
@@ -0,0 +1,59 @@
...
+
+  public DatasetProfileView merge(DatasetProfileView otherView) {
+    if (otherView == null) {
+      return this;
+    }
+
+    HashMap<String, ColumnProfileView> mergedColumns = new HashMap<>();
+    HashSet<String> allNames = new HashSet<>();
+    allNames.addAll(this.columns.keySet());
+    allNames.addAll(otherView.columns.keySet());
```
naddeoa, line 37: Similar comment here. Looks like this kind of things happens multiple times per merge. Can probably get pretty noticeable on bigger data sets.

java/core/src/main/java/com/whylogs/core/views/DatasetProfileView.java
```diff
@@ -0,0 +1,59 @@
...
+import lombok.Getter;
+import lombok.ToString;
+
+// TODO: extend writable when we do Protobuf
+@AllArgsConstructor
+@Getter
+@ToString
+public class DatasetProfileView {
+  private HashMap<String, ColumnProfileView> columns;
+  private Date datasetTimestamp;
```
naddeoa: I assume we don't want to take on any dependencies that we don't absolutely need (like guava). Since these things are supposed to be immutable, did we consider at least using things like `Collections.unmodifiableMap()` to make sure we aren't shooting ourselves in the foot by violating that anywhere?

java/core/src/main/java/com/whylogs/core/ColumnProfile.java
```diff
@@ -0,0 +1,80 @@
...
+import lombok.Getter;
+import lombok.ToString;
+
+@Getter
+@ToString
+public class ColumnProfile<T> {
+  // Required
+  private String name;
+  private ColumnSchema schema;
+  private int cachedSize;
```
richard-rogers: s/cachedSize/cacheSize/

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
richard-rogers: Yeah I don't think we need `T` here.

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
```diff
@@ -0,0 +1,141 @@
...
+      try{
+          this.isActive = true;
+          this.trackCount += 1;
+          this.doTrack(obj);
+      } finally {
+          this.isActive = false;
+      }
+  }*/
+
+  public <T> void track(HashMap<String, T> row) {
```
richard-rogers: I think the `T` here is bad as it implies more homogeneity than we want. Probably just `HashMap<String, Object>`

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
```diff
@@ -0,0 +1,141 @@
...
+    try {
+      this.isActive = true;
+      this.trackCount += 1;
+      this.doTrack(row);
+    } finally {
+      this.isActive = false;
+    }
+  }
+
+  private <T> void doTrack(HashMap<String, T> row) {
```
richard-rogers: s/T/Object/

java/core/src/main/java/com/whylogs/core/SingleFieldProjector.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.HashMap;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+@Getter
+public class SingleFieldProjector {
+  private final String columnName;
+
+  public <T> T apply(HashMap<String, T> row) {
```
richard-rogers: s/T/Object/

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
```diff
@@ -62,7 +63,14 @@ public OperationResult columnarUpdate(PreprocessedColumn data) {
     return OperationResult.ok(successes);
   }
 
-  // @Override // TODO: move this to a factory so it can be accessed for building
+  @Override
```
richard-rogers, line 66: On line 30 I think the components should be copied rather than keeping references to them.

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
```diff
@@ -62,7 +63,14 @@ public OperationResult columnarUpdate(PreprocessedColumn data) {
...
+  public HashMap<String, MetricComponent> getComponents() {
+    HashMap<String, MetricComponent> components = new HashMap<>();
+    components.put(this.maxComponent.getTypeName(), this.maxComponent);
+    components.put(this.minComponent.getTypeName(), this.minComponent);
+    return components;
+  }
+
```
richard-rogers, line 73: Is there a way to make the config argument to `zero()` optional?

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
```diff
@@ -19,6 +20,20 @@ public abstract class Metric {
 
   public abstract OperationResult columnarUpdate(PreprocessedColumn data);
 
+  public abstract HashMap<String, MetricComponent> getComponents();
+
+  public Metric merge(Metric other) {
+    Metric merged = this;
```
richard-rogers: This frightens me :)

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
```diff
@@ -19,6 +20,20 @@ public abstract class Metric {
...
+    if (!this.namespace.equals(other.namespace)) {
+      throw new IllegalArgumentException("Cannot merge metrics with different namespaces");
+    }
+
+    if (this instanceof IntegralMetric) {
```
richard-rogers: This frightens me even more :(

java/core/src/main/java/com/whylogs/core/views/ColumnProfileView.java
```diff
@@ -0,0 +1,129 @@
...
+import lombok.Getter;
+
+@Getter
+public class ColumnProfileView {
+  private HashMap<String, Metric> metrics;
+  private int successes = 0;
+  private int failures = 0;
+
+  public ColumnProfileView(HashMap<String, Metric> metrics) {
+    this.metrics = metrics;
```
richard-rogers, line 17: This should probably deep copy?

java/core/src/main/java/com/whylogs/core/views/DatasetProfileView.java
```diff
@@ -0,0 +1,59 @@
...
+
+  public HashMap<String, ColumnProfileView> getColumns(Optional<ArrayList<String>> colNames) {
+    if (colNames.isPresent()) {
+      HashMap<String, ColumnProfileView> result = new HashMap<>();
+      for (String colName : colNames.get()) {
+        result.put(colName, this.columns.get(colName));
+      }
+      return result;
+    } else {
+      return this.columns;
```
richard-rogers: Should this deep copy?

java/core/src/main/java/com/whylogs/core/ColumnProfile.java
```diff
@@ -0,0 +1,80 @@
...
+import com.whylogs.core.schemas.ColumnSchema;
+import com.whylogs.core.views.ColumnProfileView;
+import java.util.ArrayList;
+import java.util.HashMap;
+import lombok.Getter;
+import lombok.ToString;
+
+@Getter
+@ToString
+public class ColumnProfile<T> {
```
richard-rogers: How does this work for columns of mixed types?

java/core/src/test/java/com/whylogs/core/TestColumnProfile.java
```diff
@@ -0,0 +1,144 @@
...
+    ColumnProfile<Integer> profile = getDefaultColumnProfile();
+    Assert.assertEquals(profile.getName(), columnName);
+    Assert.assertEquals(profile.getSchema().getType(), Integer.class);
+    Assert.assertEquals(profile.getCachedSize(), CACHE_SIZE);
+  }
+
+  @Test
+  public void testAddMetric() {
+    ColumnProfile<Integer> profile = getDefaultColumnProfile();
+    profile.addMetric(IntegralMetric.zero(new MetricConfig()));
```
richard-rogers: The column already has an `IntegralMetric`. I think the `addMetric()` behavior in that case is to replace the old metric instance & log a warning. It looks like you test that with `metric2` below, so maybe skip the `addMetric()` here?

java/core/src/test/java/com/whylogs/core/metrics/TestMetric.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.ArrayList;
+import org.junit.Assert;
+import org.testng.annotations.Test;
+
+@Test
+public class TestMetric {
+
+  @Test
+  public void testMetrics() {
+    ArrayList<Metric> metrics = new ArrayList<>();
```
richard-rogers, line 12: It's not clear to me what this is testing

java/core/src/main/java/com/whylogs/core/views/ColumnProfileView.java
TheMellyBee: They are readonly! Thanks for this tip!

java/core/src/main/java/com/whylogs/core/views/DatasetProfileView.java
TheMellyBee: That's a great idea in this place. I did know there was an an unmodifiableMap. I'll look into implementing that here and in the ColumnProfileView

java/core/src/main/java/com/whylogs/core/ColumnProfile.java
TheMellyBee: Sorry, I don't know what you're wanting here? Could you clarify your comment?

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
TheMellyBee: Wouldn't a row always be of the same type? Using T here allows that type not to be lost. If they want their typing to be object for a data then they can set it to that, but I wanted to let the user be able to implement the data they want to give.

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
TheMellyBee: Same as the above.

java/core/src/main/java/com/whylogs/core/SingleFieldProjector.java
TheMellyBee: Same as above

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
TheMellyBee: Yes I can make a second method that allows that. I'll add it in.

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
TheMellyBee: I can do a copy here.

java/core/src/main/java/com/whylogs/core/views/DatasetProfileView.java
TheMellyBee: It could be. That probably would be good so they can't accidentally change the columns, or going with the suggestion above of using an unmodifiableMap could solve it to. What are your thoughts?

java/core/src/main/java/com/whylogs/core/ColumnProfile.java
TheMellyBee: You'd need to pick the top type so for int and float it would be Number. This is going to be true of the data they have in the row as well. My point was to keep it matched up with how they would be using it in Java.

java/core/src/main/java/com/whylogs/core/ColumnProfile.java
TheMellyBee: Is this trying to say to change it to cachSize? This comment isn't very clear to me.

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
TheMellyBee: Wait is this just for one column at a time (I thought that this from the name row in python) or is this meant to be the full table? This changes my opinion greatly

java/core/src/main/java/com/whylogs/core/ColumnProfile.java
richard-rogers: Sorry, I sometimes use sed notation. Yes, I'm saying change `cachedSize` to `cacheSize`.

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
TheMellyBee: This is a pull over from the python, I don't believe this is thread safe. I'll sync with @jamie256 on this. It's only being used here.

java/core/src/main/java/com/whylogs/core/views/ColumnProfileView.java
TheMellyBee: Does the stream combine all before going to the foreach? Aka if one comes in A B C D and the other comes in D C B F Will it correctly be able see these as a set so what we end up going through is (A B C D F) in whatever order?

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
TheMellyBee: Let's chat about this. I don't like it either, but we need some way of knowing if this is an IntegralMetric or not. We could use the namespace. Sadly there are a couple places we use this where we loose the type info even with the CRTP.

java/core/src/main/java/com/whylogs/core/views/ColumnProfileView.java
TheMellyBee: There is no deep copy in java outside use the deserialziation and serialization method although Apache Commons Lang has one according to this [article](https://www.baeldung.com/java-copy-hashmap). I have changed it over to `this.metrics.pullAll(metrics)`. This does a shallow copy

java/core/src/test/java/com/whylogs/core/metrics/TestMetric.java
TheMellyBee: This was just double checking that this is working as expected in a Collection of Metrics with the merge currently having that odd way of making sure it's an IntegralMetric and it won't fail

java/core/src/main/java/com/whylogs/core/views/DatasetProfileView.java
TheMellyBee: Updated this to use the unmodifiable Map on return.

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
TheMellyBee: Re looked at this and I agree. I had misunderstood this parameter. Changed

java/core/src/main/java/com/whylogs/core/views/ColumnProfileView.java
TheMellyBee: I believe the concat just runs through stream 1 and then stream2. I particularly wanted the set here so I'm going to keep it.

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
TheMellyBee: Don't need a copy here.

java/core/src/main/java/com/whylogs/core/metrics/Metric.java
TheMellyBee: @naddeoa could you take a look at this section of the code? I can walk you through it, but basically we are using CRTP to be able to have a collection of Metrics but not loose the typing. Sadly there is a section we use merge on a Collection of Metrics ... which means we have type erasure. All this method is doing is figuring out the type and sending to the correct subclass merge. I'm all ears for a better way to do this.

TheMellyBee: Pretty sure namespace could be done instead of is instance to speed it up, but otherwise it would be the same

TheMellyBee:
Discovered I missed a test, which led to a bug! Huzzah! :D Fixing that then this will be open for review again

java/core/src/main/java/com/whylogs/core/ColumnProfile.java
```diff
@@ -0,0 +1,82 @@
...
+  // TODO: this only gets one not every part of the row. Should projector actually do it multiple?
+  public void track(HashMap<String, Object> row) {
+    T value = this.projector.apply(row);
+    this.cachedValues.add(value);
+
+    if (this.cachedValues.size() >= this.cacheSize) {
+      this.flush();
+    }
+  }
+
```
richard-rogers, line 63: Do we have flush on destruction, or do we leave cached values orphaned?

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
```diff
@@ -0,0 +1,129 @@
...
+      for (String colName : schemaColumnNames) {
+        if (!this.columns.containsKey(colName)) {
+          newColumnNames.add(colName);
+        }
+      }
+      this.initializeNewColumns(newColumnNames);
+    }
+
+    for (String col : row.keySet()) {
+      ArrayList<Object> values = new ArrayList<>();
```
richard-rogers: Maybe pull `values` up to a top-level member to avoid the alloc/construction/destruction in the loop?

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
```diff
@@ -62,11 +63,22 @@ public OperationResult columnarUpdate(PreprocessedColumn data) {
...
   public static IntegralMetric zero(MetricConfig config) {
     return new IntegralMetric();
   }
 
+  public static IntegralMetric zero() {
+    return new IntegralMetric();
```
richard-rogers: maybe `return IntegralMetric.zero(new MetricConfig())` ?

java/core/src/main/java/com/whylogs/core/views/ColumnProfileView.java
richard-rogers: We might need to roll our own deep copy.

java/core/src/test/java/com/whylogs/core/metrics/TestMetric.java
```diff
@@ -0,0 +1,23 @@
...
+    metrics.add(IntegralMetric.zero(new MetricConfig()));
+    metrics.add(IntegralMetric.zero(new MetricConfig()));
+
+    for (Metric metric : metrics) {
+      Assert.assertTrue(metric instanceof IntegralMetric);
+      metric.merge(new IntegralMetric());
```
richard-rogers, line 17: Did you want to assert the `Metric` returned by the merge is an `IntegralMetric`?

java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
TheMellyBee: kk

java/core/src/test/java/com/whylogs/core/metrics/TestMetric.java
TheMellyBee: Both, I was sanity checking that this metric  should have been an IntegralMetric as I was testing out Metric's as typing gets lost in the collection.

java/core/src/main/java/com/whylogs/core/ColumnProfile.java
TheMellyBee: Good question. I was looking into this, I guess Java doesn't have a deconstructor and finalize has been deprecated. I went ahead and implemented AutoClosable. That should do it.

java/core/src/main/java/com/whylogs/core/views/ColumnProfileView.java
TheMellyBee: kk I'm ok with that. :) So when talking through deep copy do we want each metric to also be copied into a new metric?

java/core/src/main/java/com/whylogs/core/ColumnProfile.java
```diff
@@ -0,0 +1,91 @@
...
+    this.cachedValues = new ArrayList<>();
+  }
+
+  public void addMetric(Metric<?> metric) {
+    if (this.metrics.containsKey(metric.getNamespace())) {
+      // TODO: Add logger with warning about replacement
+    }
+
+    this.metrics.put(metric.getNamespace(), metric);
+    // TODO: Wouldn't this implement a success count here?
```
richard-rogers, line 51: It looks like `successCount` is counting merges, so this shouldn't affect it. We should double check that column merges cope with metrics that aren't present in both mergees.

java/core/src/main/java/com/whylogs/core/ColumnProfile.java
richard-rogers: I thought through this more, and I don't think it's a problem to orphan values... Any operation on the view that cares flushes the cache. If you have cached values but don't do a flushing operation on the profile the orphaned values don't matter.

java/core/src/main/java/com/whylogs/core/ColumnProfile.java
```diff
@@ -0,0 +1,91 @@
...
+
+    if (this.cachedValues.size() >= this.cacheSize) {
+      this.flush();
+    }
+  }
+
+  public void flush() {
+    // TODO: Logger was initially here, but only for when it was forced, think it through
+    ArrayList<T> oldCache = this.cachedValues;
+    this.cachedValues = new ArrayList<>();
```
richard-rogers, line 67: How expensive is allocation/littering in Java? Would it be worthwhile to maintain an index into the cache so we can re-use it? (ok as a follow-up PR if we decide to do it).

TheMellyBee: Good point. Yes let's go ahead just cause that is a little better, but let's do it in another PR

java/core/src/main/java/com/whylogs/core/ColumnProfile.java
TheMellyBee: Similar to the others. I'd like to go ahead and do that as it wasn't the original plan for this PR and iterate. :)

richard-rogers (approved):

PR #870, opened 2022-09-23
Better whylabs support for logged images

richard-rogers:
This depends on #860 

It adds a resolver to the `ImageMetric` analogous to the `DatasetSchema`'s to create the standard metrics (counts, types, cardinality, etc.) that the visualizer & whylabs seem to require. The logical submetric name is prefixed with a metric type code so that the related metrics can be aggregated into some `ColumnProfileView` when the `ImageMetric` is "uncompounded" to send it to whylabs.

### DIFF
--- a/python/poetry.lock
+++ b/python/poetry.lock
@@ -99,7 +99,7 @@ tests_no_zope = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy 
 
 [[package]]
 name = "autoflake"
-version = "1.6.0"
+version = "1.6.1"
 description = "Removes unused imports and unused variables"
 category = "dev"
 optional = false
@@ -184,14 +184,14 @@ dev = ["Sphinx (==4.3.2)", "black (==22.3.0)", "build (==0.8.0)", "flake8 (==4.0
 
 [[package]]
 name = "boto3"
-version = "1.24.77"
+version = "1.24.78"
 description = "The AWS SDK for Python"
 category = "main"
 optional = false
 python-versions = ">= 3.7"
 
 [package.dependencies]
-botocore = ">=1.27.77,<1.28.0"
+botocore = ">=1.27.78,<1.28.0"
 jmespath = ">=0.7.1,<2.0.0"
 s3transfer = ">=0.6.0,<0.7.0"
 
@@ -200,7 +200,7 @@ crt = ["botocore[crt] (>=1.21.0,<2.0a0)"]
 
 [[package]]
 name = "botocore"
-version = "1.27.77"
+version = "1.27.78"
 description = "Low-level, data-driven core of boto 3."
 category = "main"
 optional = false
@@ -1149,7 +1149,7 @@ webpdf = ["pyppeteer (>=1,<1.1)"]
 
 [[package]]
 name = "nbformat"
-version = "5.5.0"
+version = "5.6.0"
 description = "The Jupyter Notebook format"
 category = "main"
 optional = false
@@ -1158,7 +1158,7 @@ python-versions = ">=3.7"
 [package.dependencies]
 fastjsonschema = "*"
 jsonschema = ">=2.6"
-jupyter_core = "*"
+jupyter-core = "*"
 traitlets = ">=5.1"
 
 [package.extras]
@@ -1680,7 +1680,7 @@ python-versions = ">=3.6"
 
 [[package]]
 name = "pyzmq"
-version = "24.0.0"
+version = "24.0.1"
 description = "Python bindings for 0MQ"
 category = "main"
 optional = false
@@ -2128,7 +2128,7 @@ widechars = ["wcwidth"]
 
 [[package]]
 name = "tenacity"
-version = "8.0.1"
+version = "8.1.0"
 description = "Retry code until it succeeds"
 category = "dev"
 optional = false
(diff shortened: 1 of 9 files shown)


python/poetry.lock
```diff
@@ -99,7 +99,7 @@ tests_no_zope = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy
 
 [[package]]
 name = "autoflake"
-version = "1.6.0"
+version = "1.6.1"
```
jamie256: Why are you updating the poetry.lock file in this PR?

python/whylogs/extras/image_metric.py
```diff
@@ -101,47 +105,138 @@ def image_based_metadata(img):
...
+    @abstractmethod
+    def prefixes(self) -> Set[str]:
+        raise NotImplementedError
+
+    @abstractmethod
+    def resolve(self, name: str, why_type: DataType, fi_disabled: bool = False) -> Dict[str, Metric]:
+        raise NotImplementedError
+
+
+class StandardResolver(ImageResolver):
```
jamie256: The name `StandardResolver` is surprisingly generic, maybe something like StandardImageMetricResolver while awkward is easier to read and understand?

python/whylogs/extras/image_metric.py
```diff
@@ -101,47 +105,138 @@ def image_based_metadata(img):
...
+            metrics[f"fi_{name}"] = StandardMetric.frequent_items.zero(MetricConfig())
+
+        return metrics
+
+
+@dataclass(frozen=True)
+class ImageMetricConfig(MetricConfig):
+    allowed_exif_tags: Set[str] = field(default_factory=set)
+    forbidden_exif_tags: Set[str] = field(default_factory=set)
+    resolver: Resolver = field(default_factory=StandardResolver)
```
jamie256: I feel like the class hierarchy is broken (but connected by name) here: a `StandardResolver(ImageResolver)` where the base class is `ImageResolver(ABC)` is getting assigned to a field of type `whylogs.core.resolvers.Resolver`.

Maybe having a StandardImageResolver(Resolver) would reduce the hierarchy by 1 level and simplify as you can define the `prefixes` there if needed.

python/whylogs/extras/image_metric.py
```diff
@@ -101,47 +105,138 @@ def image_based_metadata(img):
...
-    def zero(cls, config: MetricConfig) -> "ImageMetric":
+    def zero(cls, config: Optional[MetricConfig] = None) -> "ImageMetric":
+        config = config or ImageMetricConfig()
+        if not isinstance(config, ImageMetricConfig):
+            logger.error("ImageMetric.zero() needs an ImageMetricConfig")
+            config = ImageMetricConfig()
+
         dummy_image = new_image("HSV", (1, 1))
-        metadata = get_pil_image_metadata(dummy_image)
+        metadata = get_pil_image_metadata(dummy_image)  # any EXIF tags will be discovered as images are logged
```
jamie256: probably don't need the dummy image path if you can add to the set of things logged.

python/whylogs/extras/image_metric.py
```diff
@@ -101,47 +105,138 @@ def image_based_metadata(img):
...
+            config.type_mapper,
+            config.fi_disabled,
+        )
 
 
 def log_image(
     images: Union[ImageType, List[ImageType], Dict[str, ImageType]],
     default_column_prefix: str = "image",
+    schema: Optional[DatasetSchema] = None,
+    default_configs: Optional[ImageMetricConfig] = None,
```
jamie256: Adding `schema ` here as optional makes sense, but default_configs as a parameter is confusing, why doesn't the `schema.default_configs` suffice?

python/whylogs/api/writer/whylabs.py
```diff
@@ -59,7 +59,25 @@ def _uncompounded_column_name(column_name: str, metric_name: str, submetric_name
     return f"{column_name}.{metric_name}.{submetric_name}"
 
 
+def _ugly_hack(col_name: str, metric_name: str, metric: CompoundMetric) -> Dict[str, ColumnProfileView]:
```
jamie256: better would be to name this for what it is doing, and why it is needed: e.g. back_compat? image metrics backport translation? with a TODO for simplifying it if this is not needed long term.

jamie256 (changes requested):
Left some questions, looks like we have something that should work for WhyLabs upload of image profiles? Can you describe what you verified in the description, or update the notebook example?

python/poetry.lock
richard-rogers: I had to update my poetry version, and some version bumps came in from mainline in the course of working on this PR.

python/whylogs/extras/image_metric.py
richard-rogers: Oops, that's a bug :(  The `ImageMetricConfig::resolver` should be of type `ImageResolver`.  I looked at having `ImageResolver <: Resolver`, but the arguments to `resolve()` want to be different.

python/whylogs/extras/image_metric.py
richard-rogers: That would be tricky with the allow/forbid sets. Also, SPOT for the image stats :)

python/whylogs/extras/image_metric.py
richard-rogers: If don't pass a `schema`, you can specify the default config to use in the constructed schema. I would guess the more common case is that you're happy with the default `schema` but want to fiddle with the `ImageMetricConfig`.

python/whylogs/api/writer/whylabs.py
richard-rogers: I think writing this function convinced me we need to design something else :(

python/whylogs/extras/image_metric.py
andyndang: That's just very confusing in terms of API though. Removing API is hard so I'd err on less for now

richard-rogers: For usability, I'd prefer to drop the `schema` argument before dropping the `default_configs`.

python/poetry.lock
jamie256: but you should only have to update poetry.lock if its broken or if you change the set of dependencies in pyproject.toml file?

You didn't update the dependencies in this change, so you should not be needing to update the lock file.

Generally I think we can update the poetry.lock periodically if it hasn't been updated in a while but that is probably better as a standalone maintenance PR.

python/whylogs/extras/image_metric.py
jamie256: Can we call get_pil_image_metadata() on the first image logged and do the metric resolution then?

python/whylogs/extras/image_metric.py
jamie256: schema is consistent with the other log* APIs, I'd argue for consistency in how we extend things, even if its a little awkward to instantiate a schema just so you can throw a config in there.

python/poetry.lock
richard-rogers: It's in sync with mainline now.

python/whylogs/extras/image_metric.py
richard-rogers: I restructured the code to eliminate the need for the dummy image.

python/whylogs/extras/image_metric.py
richard-rogers: Outvoted :)

jamie256 (approved):
LGTM!

PR #878, opened 2022-09-27
Fix ImageMetric::merge() bug

richard-rogers:
- Copy over lhs metric config
- Union tracked metadata


FelipeAdachi (approved):
LGTM!
Thanks, Richard!

PR #886, opened 2022-09-28
Update image logging example notebook

richard-rogers:
Update image logging example notebook.
Also removes the ability to configure the `ImageMetric` type mapper and schema.


FelipeAdachi (commented):
I see some "pip install" cells in the example without "!" or "%".

It seems to work in Collab, but what is the right way to do it? I've been using "%".

jamie256 (approved):
Looks good.

Do you think we should prefix the exif tags extracted with "exif" e.g. exif_Artist?

I noticed that the namespace convention for the extract features is image/[metric_namespace]_[extracted_feature]/[metric_summary_key]:
```
image/card_Artist/est': 1.0
```

but because the "est" applies to the "card" part of the name not the "Artist" part, I'd expect going from general to specific in the path they would be adjacent something like this:
```
image/Artist_card/est': 1.0
```
or maybe:
```
image/exif_Artist_card/est': 1.0
```

richard-rogers:
> I noticed that the namespace convention for the extract features is image/[metric_namespace]_[extracted_feature]/[metric_summary_key]:
> 
> ```
> image/card_Artist/est': 1.0
> ```

#894 changes this to `image/[extracted_feature]:[metric_namespace]/[component_key]`. This groups all the metrics tracked for the extracted feature together.

PR #893, opened 2022-09-30
Fixed standard aggregators/serializers/deserializers 

### LINKED ISSUE #892: Weird behavior with tracking max and min of ints
## Description

```
import whylogs as why

res = why.log(row={"foo":1})

v = res.view()
v.to_pandas()
```

`ints/max` is 1

Merge them repeatedly 30 times:

```
r = v
for i in range(0, 30):
  r = r.merge(r)

```
and `ints/max` is now 1073741824 (2^30)

andyndang:
Fixed standard aggregators/serializers/deserializers not registering correctly

We create the standard registry too early before these objects are created. Fixed by making the standard registries lazy

## Description

* Fixes: https://github.com/whylabs/whylogs/issues/892
* Had to revert the psutil package since the latest one can't be installed in my environment


jamie256 (approved):
Thanks Andy!

PR #894, opened 2022-09-30
Generalize CompoundMetric for better whylabs support

richard-rogers:
Creates a new `MultiMetric` base class to replace `CompoundMetric`. `MultiMetric` manages multiple `Metric`s for each virtual submetric. They get "uncompounded" together into new columns for each virtual submetric.


python/whylogs/api/writer/whylabs.py
```diff
@@ -70,36 +71,23 @@ def _uncompounded_column_name(column_name: str, metric_name: str, submetric_name
...
     for submetric_name, submetric in metric.submetrics.items():
         new_col_name = _uncompounded_column_name(col_name, metric_name, submetric_name, metric)
         result[new_col_name] = ColumnProfileView({submetric.namespace: submetric})
     return result
 
 
+def _uncompound_multimetric(col_name: str, metric_name: str, metric: MultiMetric) -> Dict[str, ColumnProfileView]:
+    result: Dict[str, ColumnProfileView] = dict()
+    for submetric_name, submetrics in metric.submetrics.items():
+        new_col_name = f"{col_name}.{submetric_name}" if not _v0_compatible_image_feature_flag() else submetric_name
```
richard-rogers: Should check that it's an `ImageMetric`

python/whylogs/api/writer/whylabs.py
```diff
@@ -70,36 +71,23 @@ def _uncompounded_column_name(column_name: str, metric_name: str, submetric_name
...
-        return _uncompound_image_metric(col_name, metric_name, metric)
-
     result: Dict[str, ColumnProfileView] = dict()
     for submetric_name, submetric in metric.submetrics.items():
         new_col_name = _uncompounded_column_name(col_name, metric_name, submetric_name, metric)
         result[new_col_name] = ColumnProfileView({submetric.namespace: submetric})
     return result
 
 
+def _uncompound_multimetric(col_name: str, metric_name: str, metric: MultiMetric) -> Dict[str, ColumnProfileView]:
```
jamie256: I think we should move these _uncompound* methods out of whylabs.py since its not specific to the whylabs extra

python/whylogs/core/metrics/multimetric.py
```diff
@@ -0,0 +1,170 @@
...
+            result[namespace] = deepcopy(rhs[namespace])
+
+    return result
+
+
+class MultiMetric(Metric, ABC):
+    """
+    MultiMetric serves as a base class for custom metrics that consist
+    of one or more metrics. It is handy when you need to do some
+    processing of the logged values and track serveral metrics on
```
jamie256:
```suggestion
    processing of the logged values and track several metrics on
```

python/whylogs/core/metrics/multimetric.py
```diff
@@ -0,0 +1,170 @@
...
+    """
+
+    submetrics: Dict[str, Dict[str, Metric]]
+
+    def __init__(self, submetrics: Dict[str, Dict[str, Metric]]):
+        """
+                Parameters
+                ----------
+                submetrics : Dict[str, Dict[str, Metric]]
+                    The collection of metrics that comprise the CompoundMetric.
```
jamie256: CompoundMetric?

python/whylogs/core/metrics/multimetric.py
```diff
@@ -0,0 +1,170 @@
...
+            else:
+                submetrics[submetric_name] = deepcopy(other.submetrics[submetric_name])
+
+        return submetrics
+
+    def merge(self: MULTI_METRIC, other: MULTI_METRIC) -> MULTI_METRIC:
+        return self.__class__(self.merge_submetrics(other))
+
+    def to_protobuf(self) -> MetricMessage:
+        msg = {}
```
jamie256, line 101: is it a valid case for submetrics to be empty?

python/whylogs/core/metrics/multimetric.py
```diff
@@ -0,0 +1,170 @@
...
+    def get_component_paths(self) -> List[str]:
+        res = []
+        for sub_name, metrics in self.submetrics.items():
+            for namespace, metric in metrics.items():
+                for comp_name in metric.get_component_paths():
+                    res.append(f"{sub_name}:{namespace}/{comp_name}")
+
+        return res
+
+    def to_summary_dict(self, cfg: SummaryConfig) -> Dict[str, Any]:
```
jamie256, line 119: we should make `cfg` here optional

jamie256 (commented):
Looks better. Do we still use compound metric?

I see a notebook failure, and left some minor comments.

python/whylogs/core/metrics/multimetric.py
richard-rogers: I'll do a separate PR to make it optional for all metrics. It looks like it's currently required everywhere.

python/whylogs/core/metrics/multimetric.py
richard-rogers: Hmmm... it works for most methods. How does `MetricMessage` feel about being empty?

richard-rogers:
> Looks better. Do we still use compound metric?

I plan to move the NLP metric to `MultiMetric`. Once that is done, nothing will be using the `CompoundMetric`.  I'm on the fence about keeping `CompoundMetric`... it will be unused, but it's simpler. I'd probably vote to drop it unless it already has users.

python/whylogs/core/metrics/multimetric.py
jamie256: I was wondering if the code here handles the empty case ok, jut curious about the edge case.

python/whylogs/api/writer/whylabs.py
jamie256: but also be careful not to cause problems if the image extra is not installed.

jamie256 (dismissed):
Looks good, I think we can address issues mentioned as follow up items since they are not core to your change here. Approved pending green CI / notebook tests.

jamie256 (approved):
LGTM!

PR #903, opened 2022-10-10
Make SummaryConfig optional

richard-rogers:
Makes `SummaryConfig` an optional argument to `to_summary_dict()` with `SummaryConfig()` as the default if it's not passed.

https://app.clickup.com/t/3nn9vh1


jamie256 (approved):
Thank you! Looks good.

PR #904, opened 2022-10-11
Count booleans in cardinality metric, and add tests

### LINKED ISSUE #901: whylogs does not track cardinality or distribution on boolean columns
## Description

In v0 we had cardinality and distribution on boolean features, but in v1 we don't add these by default. We can add these back so that its less confusing why these metrics appear to be missing. We might want to special case these metrics as simple counts likely outperform using hll/kll for binary ranged features.

## Suggestions

Turn on cardinality and distribution metrics for boolean columns when logging with default DatasetSchema in whylogs.

jamie256:
## Description

Booleans are not added to the numpy or pandas views of the PreprocessedColumns and then are not counted by cardinality metric.

## Changes

This adds special handling to look for the counts of booleans when processing cardinality metric, so that we can correctly count the boolean values. We use integer values 0, 1 to count the unique booleans since the HLL sketch does not directly support boolean (and this seemed like the least surprising way to track these, consistent with C++ and that python treats these as numeric types.)

Also added a test to cover this case.

## Related

Closes #901 

### DIFF
--- a/python/tests/core/metrics/test_metrics.py
+++ b/python/tests/core/metrics/test_metrics.py
@@ -1,10 +1,15 @@
 import numpy as np
 import pandas as pd
+import pytest
 
 import whylogs as why
 from whylogs.core import ColumnProfileView
 from whylogs.core.metrics.maths import VarianceM2Result, parallel_variance_m2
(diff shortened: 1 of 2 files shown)


TheMellyBee (approved):
LGTM.

PR #908, opened 2022-10-13
Add the Result Set to Java

TheMellyBee:
## Description
Implements ResultSet in Java

- [X ] I have reviewed the [Guidelines for Contributing](CONTRIBUTING.md) and the [Code of Conduct](CODE_OF_CONDUCT.md).


java/core/src/main/java/com/whylogs/api/logger/resultSets/ResultSet.java
```diff
@@ -0,0 +1,41 @@
...
+import lombok.NoArgsConstructor;
+
+/**
+ * A holder object for profiling results.
+ *
+ * <p>A whylogs.log call can result in more than one profile. This wrapper class simplifies the
+ * navigation among these profiles.
+ *
+ * <p>Note that currently we only hold one profile but we're planning to add other kinds of profiles
+ * such as segmented profiles here.
```
richard-rogers: Don't we have segmented profiles already on the Python side?

java/core/src/main/java/com/whylogs/api/logger/resultSets/ResultSet.java
```diff
@@ -0,0 +1,41 @@
...
+
+  // TODO: implement read and write when I make the reader and writer
+
+  public abstract Optional<DatasetProfileView> view();
+
+  public abstract Optional<DatasetProfile> profile();
+
+  // TODO: Come back for ModelPerformanceMetrics
+
+  // Question: why is the python addMetrics when it only adds the one?
```
richard-rogers: Possibly a typo? It doesn't look like it's called anywhere in the code base, so rename it or maybe drop it?

java/core/src/main/java/com/whylogs/api/logger/resultSets/ResultSet.java
```diff
@@ -0,0 +1,41 @@
...
+  public void addMetric(String name, Metric<?> metric) throws Error {
```
richard-rogers: What is `name` naming? (same question on the Python side). What is the `get()` on line 39 getting?

java/core/src/main/java/com/whylogs/api/logger/resultSets/ResultSet.java
TheMellyBee: KK I'll update that then :)

java/core/src/main/java/com/whylogs/api/logger/resultSets/ResultSet.java
TheMellyBee: hmm I assumed the name was the of the column, but that doesn't make sense. In the python there are no usages for this function found. This might just be deadcode that isn't needed. 

The get is just because this is in an optional.

java/core/src/main/java/com/whylogs/api/logger/resultSets/ResultSet.java
TheMellyBee: Good question. I copied over this header because it is correct with java not having segments yet. 

As for the python side, yes I think segments are done. I'll check with jamie and send up an update on the python side

java/core/src/main/java/com/whylogs/api/logger/resultSets/ResultSet.java
richard-rogers: Keep it in sync with Python :)  (i.e., rename or drop in Python if we decide to go that way)

java/core/src/main/java/com/whylogs/api/logger/resultSets/ResultSet.java
richard-rogers: I added some `add_metric()` methods to other classes, but I don't think I did this one. I'm not sure it makes sense to add a metric to a `ResultSet`. Maybe run this by @jamie256 and drop it (from Python too) if it's dead. Or just resting. Or pining for the fjords...

TheMellyBee: Alright I sycned with @jamie256 this is not currently being used but is meant for non-segmented model performance metrics row count, etc. Let's go ahead and leave this in here for that as it will be coming in the future. 

I love the pining the fjords haha

java/core/src/main/java/com/whylogs/api/logger/BasicCache.java
```diff
@@ -0,0 +1,3 @@
+package com.whylogs.api.logger;
+
+public class BasicCache {}
```
jamie256: I don't think we need this?

java/core/src/main/java/com/whylogs/api/logger/resultSets/ResultSet.java
jamie256: The ResultSet.add_metric adds a named metric to the profile (not associated with a column). The name here is whatever you want to key the metric on as this will be a dataset level metric, but we don't need this yet in the Java port, I would leave out things we don't need to log some input and have it create a profile that can be uploaded to WhyLabs.

java/core/src/main/java/com/whylogs/api/logger/resultSets/ResultSet.java
```diff
@@ -0,0 +1,40 @@
...
+
+  // TODO: Come back for ModelPerformanceMetrics
+
+  public void addMetric(String name, Metric<?> metric) throws Error {
+    if (!this.profile().isPresent()) {
+      throw new Error(
+          "Cannot add " + name + " metric " + metric + " to a result set without a profile");
+    }
+
+    this.profile().get().addMetric(name, metric);
```
jamie256: Avoid if/else or checking isPresent when using Optionals, e.g. something like this:

```suggestion
    DatasetProfile profile = this.profile().orElseThrow(() -> new Error("Cannot add " + name + " metric " + metric + " to a result set without a profile");
    profile.addMetric(name, metric);
```

java/core/src/main/java/com/whylogs/api/logger/resultSets/ViewResultSet.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.Optional;
+import lombok.*;
+
+@EqualsAndHashCode(callSuper = true)
+@Data
+public class ViewResultSet extends ResultSet {
+  @NonNull private final DatasetProfileView view;
+
+  public ViewResultSet(@NonNull DatasetProfileView view) {
+    super();
```
jamie256: do you need this?

java/core/src/main/java/com/whylogs/api/logger/resultSets/ViewResultSet.java
```diff
@@ -0,0 +1,32 @@
...
+    this.view = view;
+  }
+
+  public ViewResultSet(DatasetProfile profile) {
+    super();
```
jamie256: do you need this?

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
```diff
@@ -94,9 +99,10 @@ public void setDatasetTimestamp(Instant datasetTimestamp) {
...
   private void initializeNewColumns(Set<String> colNames) {
     for (String column : colNames) {
-      ColumnSchema columnSchema = this.schema.columns.get(column);
+      ColumnSchema columnSchema = this.schema.getColumns().get(column);
       if (columnSchema != null) {
-        this.columns.put(column, new ColumnProfile(column, columnSchema, this.schema.cache_size));
+        this.columns.put(
+            column, new ColumnProfile<>(column, columnSchema, this.schema.getCacheSize()));
       }
       // TODO: log warning 'Encountered a column without schema: %s", col' in an else
```
jamie256: remove

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
```diff
@@ -94,9 +99,10 @@ public void setDatasetTimestamp(Instant datasetTimestamp) {
 
   private void initializeNewColumns(Set<String> colNames) {
     for (String column : colNames) {
-      ColumnSchema columnSchema = this.schema.columns.get(column);
+      ColumnSchema columnSchema = this.schema.getColumns().get(column);
```
jamie256: why not use` this.schema.get(column)` here, the `getColumns()` is unnecessary, then restructure by using the returned optional.

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
```diff
@@ -94,9 +99,10 @@ public void setDatasetTimestamp(Instant datasetTimestamp) {
...
       if (columnSchema != null) {
-        this.columns.put(column, new ColumnProfile(column, columnSchema, this.schema.cache_size));
+        this.columns.put(
+            column, new ColumnProfile<>(column, columnSchema, this.schema.getCacheSize()));
```
jamie256: maybe too late to save you much here, but the cache only really makes sense in python, where there is a C++ library (whylogs-sketching) that can handle arrays of inputs more efficiently. In whylogs Java the sketching library is also Java so this looks like a premature optimization.

java/core/src/test/java/com/whylogs/api/resultsets/TestProfileResultSet.java
```diff
@@ -0,0 +1,79 @@
...
+
+  private ProfileResultSet defaultResultSet() {
+    HashMap<String, Object> data = new HashMap<>();
+    data.put("test", 1);
+    data.put("test2", "2");
+    DatasetSchema datasetSchema = new DatasetSchema();
+    datasetSchema.resolve(data);
+
+    DatasetProfile datasetProfile =
+        new DatasetProfile(Optional.of(datasetSchema), Optional.empty(), Optional.empty());
```
jamie256: Consider: Even if just for our internal use, it may be worth a ctor overload that takes in 0 or 1 parameters?

java/core/src/test/java/com/whylogs/api/resultsets/TestViewResultSet.java
```diff
@@ -0,0 +1,41 @@
+package com.whylogs.api.resultsets;
+
+import com.whylogs.api.logger.resultSets.ViewResultSet;
```
jamie256: consistency in casing for resultsets vs restultSets?

jamie256 (commented):
left some initial comments.

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
TheMellyBee: Remove completely? Why is that?

java/core/src/test/java/com/whylogs/api/resultsets/TestProfileResultSet.java
TheMellyBee: Yeah I was thinking the same thing. I even have a note on it. I don't like all these optionals as the main workflow.  I was planning on doing it in the next iteration.

java/core/src/test/java/com/whylogs/api/resultsets/TestViewResultSet.java
TheMellyBee: haha good catch

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
TheMellyBee: Ok, given it's already here do you think we should leave it or go ahead?

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
TheMellyBee: Old habit on what was considered better form. I'm happy to just do the get.

java/core/src/main/java/com/whylogs/api/logger/BasicCache.java
TheMellyBee: Given on the other comment we may not need the cache in Java. Given that I was already adding these do we want to continue or is it a worthwhile savings to remove? Basically how likely do we think we will want the cache?

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
TheMellyBee: I'll take out the cache.

java/core/src/main/java/com/whylogs/api/logger/BasicCache.java
TheMellyBee: Clarified in person, I'll be removing as it won't be useful in Java.

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
TheMellyBee: Meant line 101. I'll remove :)

jamie256: clarification: remove the comment:
> // TODO: log warning 'Encountered a column without schema: %s", col' in an else

java/core/src/main/java/com/whylogs/api/logger/resultSets/ViewResultSet.java
TheMellyBee: Nope!

java/core/src/test/java/com/whylogs/api/resultsets/TestViewResultSet.java
TheMellyBee: https://google.github.io/styleguide/javaguide.html#s5.2.1-package-names

It's all lower case

java/core/src/main/java/com/whylogs/core/DatasetProfile.java
TheMellyBee: Ok, well looking into this if we decide to do that then we should do that in a separate PR as it does touch quite a few things.

jamie256 (approved):
Looks better, thanks!

PR #914, opened 2022-10-17
Handle numpy.bool_ when resolving metrics for bool types

### LINKED ISSUE #912: Need to treat np.bool_ types as bool when resolving metrics
## Description
We add cardinality and other numeric metrics for bool and numpy bool, but numpy.bool_ is a common pandas dataframne dtype that we missed adding these metrics for. We should treat these the same and add cardinality metrics for these columns automatically.

jamie256:
## Description

numpy.bool_ is not a subclass of bool and misses our check for the type when resolving metrics so adding this as well.

## Changes

- add np.bool_ to the type check alongside bool when resolving metrics for numeric types.
- added some tests that catch this case.

## Related

Relates to  https://github.com/whylabs/whylogs/pull/904

Closes #912 


TheMellyBee (approved):
Good fix :)

PR #932, opened 2022-10-24
Changing quantiles for KS drift detection from 9 to 100 buckets.

FelipeAdachi:
## Description

In previous experiments, it was determined that increasing the number of buckets for quantiles for KS test drift detection reduces the error compared to scipy's implementation on the complete dataset.

## Changes

## Related


andyndang:
@FelipeAdachi do you have the results from these experiments? Would love to see them

FelipeAdachi:
> @FelipeAdachi do you have the results from these experiments? Would love to see them

@andyndang I did the experiments, but it's a work in progress, I didn't share a final report yet with good descriptions. But until then, I can make a summary here:

I tested KS test varying:

- Distribution Type (Normal, Pareto, Uniform)
- Sample Size (1k, 5k, 10k, 50k, 100k)
- Number of quantile buckets (5,10, 50, 100)
- Drift Magnitude (no drift, small drift, medium drift, large drift)

I compared with Scipy's implementation when run on the complete dataset. Since whylogs gives different results for each run, for each set of parameters I ran whylogs implementation 10 times.

Overall, it looks like:
- increasing the number of buckets reduces the errors
- increasing the sample size increases the error's deviation, for same kll_k

Beware of the y-axis for small, medium and large drifts, most of those are effectively 0. The graphs as they are can be deceiving.

# Normal Distribution

Visualizing the drift injected:

image

And the p-value errors for different number of buckets:

image

# Pareto Distribution

Drift:

image

Pvalue errors:

image

# Uniform

Drift:

image

Pvalue errors:

image

andyndang:
@FelipeAdachi this is awesome. I think we should blog about this!

jamie256 (approved):
Looks good!

PR #934, opened 2022-10-24
Adjusting overall README examples and integrations

murilommen:
## Description

Basically the title.
When reviewing this PR, please check out how the Markdowns are ought to be rendered and see if this is better!


FelipeAdachi (commented):
I personally prefer to have every example listed on the root Examples folder, so the user doesn't have to click on the category to be able to see it.

But we can open this to the team and see which option is preferred overall.

If we were to list every example on the root, one option would be to always update on each category's README, and then just reference them on the root README.

TheMellyBee (dismissed):
Wow!! I love this change. It's fun and easy to read, and the writing flows better. Thank you!

FelipeAdachi (approved):
That's such a great improvement!
LGTM, thanks!

PR #935, opened 2022-10-26
Add segmented profile serialization to rolling logger

### LINKED ISSUE #911: Segments are not serialized with the rolling logger
## Description

The rolling logger lets you define segments, and segmented results will be returned by the log statements, but the accumulated profile in the rolling logger doesn't work in this case.

jamie256:
## Description
Add segmented profile serialization support to rolling logger

## Changes

* Add a SegmentedCache to store and track profiles as the rolling logger encounters new segments
* Replace segment_processing references to ProfileStore with SegmentedCache since these are different use cases.
* Add new tests, tighten the time in existing rolling logger tests
* Fix rolling logger callback to support no writer case for testing or customer serialization hook.
* Add set_dataset_timestamp to ResultSet

## Related

Fixes #911 

### DIFF
--- a/python/tests/api/logger/test_rolling.py
+++ b/python/tests/api/logger/test_rolling.py
@@ -159,6 +159,6 @@ def test_rolling_row_messages_with_segments(tmp_path: Any) -> None:
     initial_callback_count = rolling_callback.call_count
     assert initial_callback_count == 0
 
-    # after explicitly calling close on the logger, we trigger one more flush and callback
+    # after explicitly calling close on the logger, we trigger one more flush which has two segments and expect two callbacks
     rolling_logger.close()
-    assert rolling_callback.call_count > 0
+    assert rolling_callback.call_count == 2
--- a/python/tests/api/logger/test_rolling.py
+++ b/python/tests/api/logger/test_rolling.py
@@ -13,6 +13,8 @@ import whylogs as why
 from whylogs.api.store.local_store import LocalStore
 from whylogs.api.store.query import ProfileNameQuery
 from whylogs.core.errors import BadConfigError
+from whylogs.core.schema import DatasetSchema
+from whylogs.core.segmentation_partition import segment_on_column
 
 
 def test_closing(tmp_path: Any, lending_club_df: pd.DataFrame) -> None:
@@ -58,7 +60,7 @@ def test_rolling(tmp_path: Any) -> None:
 
 def test_rolling_with_callback(tmp_path: Any) -> None:
     rolling_callback = MagicMock()
-    messages = [{"col1": [i], "col2": [i * i * 1.2], "col3": ["a"]} for i in range(10)]
+    messages = [{"col1": i, "col2": i * i * 1.2, "col3": "a"} for i in range(10)]
 
     rolling_logger = why.logger(
         mode="rolling", interval=1, when="S", base_name="test_base_name", callback=rolling_callback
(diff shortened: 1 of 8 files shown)


python/whylogs/api/logger/rolling.py
```diff
@@ -170,12 +187,15 @@ def _get_time_tuple(self) -> time.struct_time:
...
+        if results is None:
             return
-        if self.skip_empty and profile.is_empty:
-            logger.debug("skip_empty is set. Skipping empty profiles")
+
+        profiles = results.get_writables()
+        if not profiles:
             return
+        number_of_profiles = len(profiles)
+        logger.debug(f"about to write {number_of_profiles}.")
```
richard-rogers:
```suggestion
        logger.debug(f"about to write {number_of_profiles} profiles.")
```

richard-rogers (dismissed):

python/whylogs/api/logger/rolling.py
```diff
@@ -170,12 +187,15 @@ def _get_time_tuple(self) -> time.struct_time:
...
-    def _flush(self, profile: DatasetProfile) -> None:
-        if profile is None:
+    def _flush(self, results: ResultSet) -> None:
+        if results is None:
             return
-        if self.skip_empty and profile.is_empty:
-            logger.debug("skip_empty is set. Skipping empty profiles")
+
+        profiles = results.get_writables()
+        if not profiles:
```
murilommen: is this verifying if it's an empty list or None? i think it'd be best to make it explicit

python/whylogs/api/logger/result_set.py
```diff
@@ -90,6 +91,11 @@ def profile(self) -> Optional[DatasetProfile]:
     def get_writables(self) -> Optional[List[Writable]]:
         return [self.view()]
 
+    def set_dataset_timestamp(self, dataset_timestamp: datetime) -> None:
+        profile = self.profile()
+        if profile:
```
murilommen: profile is not None or is not empty?

python/tests/api/logger/test_rolling.py
```diff
@@ -132,3 +132,33 @@ def test_rolling_with_local_store_writes() -> None:
...
+    )
+    rolling_logger.append_writer("local", base_dir=tmp_path)
+    # process the 10 input messages, and wait a second to allow the rolling logger to hit an interval
+    for message in messages:
+        rolling_logger.log(message)
+
+    # without an explicit call to rolling_logger.flush we expect that the elapsed time
+    # is less than the 60s interval, so the callback should not have been triggered
+    initial_callback_count = rolling_callback.call_count
+    assert initial_callback_count == 0
```
murilommen:
```suggestion
    assert rolling_callback.call_count == 0
```

python/whylogs/api/logger/rolling.py
jamie256: It is intentionally doing both, None and [] both get skipped there:
https://docs.python.org/3/library/stdtypes.html#truth-value-testing

but its a weird enough case that probably worth adding some debug logging output here, Done.

python/whylogs/api/logger/result_set.py
jamie256: Hmm thinking about this more, because this is in the base class, it would be less sneaky to allow the exception here. Added an explicit check for None and added an error message.

murilommen (dismissed):
lgtm =)

murilommen (approved):

PR #940, opened 2022-10-27
implements hellinger distance as a drift calculation

FelipeAdachi:
## Description

Implements hellinger distance to be applied between two profile views.

Currently, this is meant to be used as a standalone function to get the Hellinger Distance statistic. Further work must be done to integrate it with the `viz` module, and the decision process of which kind of algorithm/drift test is applied in different circumstances.

This currently enables hellinger for any feature that has a Distribution Metric. FI Metrics-only features are still not supported.

This repository needs 

## Changes

## Related

### DIFF
--- a/python/whylogs/viz/utils/drift_calculations.py
+++ b/python/whylogs/viz/utils/drift_calculations.py
@@ -1,15 +1,23 @@
-from typing import Dict, Optional
+import warnings
+from typing import Dict, Optional, Union
 
 import numpy as np
 from scipy import stats  # type: ignore
+from scipy.spatial.distance import euclidean
 from typing_extensions import TypedDict
 from whylogs_sketching import kll_doubles_sketch  # type: ignore
 
+from whylogs.core.view.column_profile_view import ColumnProfileView  # type: ignore
 from whylogs.core.view.dataset_profile_view import DatasetProfileView  # type: ignore
 from whylogs.viz.utils.frequent_items_calculations import (
     FrequentStats,
     get_frequent_stats,
     zero_padding_frequent_items,
+)
+from whylogs.viz.utils.histogram_calculations import (
+    HIST_AVG_NUMBER_PER_BUCKET,
+    MAX_HIST_BUCKETS,
+    _calculate_bins,
 )
 
 QUANTILES = list(np.linspace(0, 1, 100))
@@ -20,6 +28,13 @@ class ColumnDriftValue(TypedDict):
 
     p_value: float
     test: str
+
+
+class ColumnDriftStatistic(TypedDict):
+    """Statistic for applied algorithm, along with the name of the applied algorithm."""
+
+    statistic: float
+    algorithm: str
 
 
 def _compute_ks_test_p_value(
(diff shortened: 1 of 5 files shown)


richard-rogers (dismissed):
LGTM

python/whylogs/viz/utils/drift_calculations.py
```diff
@@ -1,16 +1,24 @@
...
 from whylogs.core.view.dataset_profile_view import DatasetProfileView  # type: ignore
 from whylogs.viz.utils.frequent_items_calculations import (
     FrequentStats,
     get_frequent_stats,
     zero_padding_frequent_items,
 )
+from whylogs.viz.utils.histogram_calculations import (
+    HIST_AVG_NUMBER_PER_BUCKET,
+    MAX_HIST_BUCKETS,
+    _calculate_bins,
```
murilommen: won't this come from `whylogs.viz.utils` instead of histogram_calculations?

jamie256: Consider moving those constants and `_calculate_bins` to the drift_calculations.py file (or some common location). One concern is that we may require other extras in order to use this method programmatically?

python/whylogs/viz/utils/drift_calculations.py
```diff
@@ -183,9 +198,40 @@ def _get_chi2_p_value(target_view_column, reference_view_column) -> Optional[Col
...
+
+    start = min([target_kll_sketch.get_min_value(), ref_kll_sketch.get_min_value()])
+    end = max([target_kll_sketch.get_max_value(), ref_kll_sketch.get_max_value()])
+    n = target_kll_sketch.get_n() + ref_kll_sketch.get_n()
+    bins, end = _calculate_bins(end=end, start=start, n=n, avg_per_bucket=HIST_AVG_NUMBER_PER_BUCKET, max_buckets=nbins)
+
+    target_pmf = target_kll_sketch.get_pmf(bins)
+    ref_pmf = ref_kll_sketch.get_pmf(bins)
+
+    distance = euclidean(np.sqrt(target_pmf), np.sqrt(ref_pmf)) / np.sqrt(2)
```
jamie256: Might be worth factoring the hellinger distance calculation out into a separate function for better reusability (signature could be two pmfs rather than ColumnProfileViews), can you also add a comment with reference link to the equation used? e.g. https://en.wikipedia.org/wiki/Hellinger_distance

I like that you explicitly use euclidean here as it makes this easier to understand.

Do you need to add numpy as a dependency of whylogs[viz] with this change?

jamie256 (commented):
Very nice to see this get into our drift detection algorithms! Thanks!

python/whylogs/viz/utils/drift_calculations.py
FelipeAdachi: Yes, absolutely, will do that!

oh wow, I haven't realized that numpy is not already a core dependency. If that's the case, then yes, it should be added as a dependency to whylogs[viz]

python/whylogs/viz/utils/drift_calculations.py
FelipeAdachi: Not sure this is what you meant, but yes, to use hellinger calculation programmatically, we'll currently need to install and import [viz]. Everything drift-related ended up in the viz module. We should discuss if this is the correct approach.

python/whylogs/viz/utils/drift_calculations.py
FelipeAdachi: done in the latest commit

python/whylogs/viz/utils/drift_calculations.py
FelipeAdachi: done in the latest commits. we can open another one if something needs to be done around this comment:

`One concern is that we may require other extras in order to use this method programmatically?`

jamie256 (dismissed):
LGTM!

python/whylogs/viz/utils/drift_calculations.py
```diff
@@ -183,9 +196,62 @@ def _get_chi2_p_value(target_view_column, reference_view_column) -> Optional[Col
...
+        The hellinger distance between the two discrete probability distributions.
+        Between 0 (identical distributions) and 1 (completely different distributions).
+    """
+    # https://en.wikipedia.org/wiki/Hellinger_distance
+    distance = euclidean(np.sqrt(target_pmf), np.sqrt(reference_pmf)) / np.sqrt(2)
+    return distance
+
+
+def _get_hellinger_distance(
+    target_view_column: ColumnProfileView, reference_view_column: ColumnProfileView, nbins=None
```
bernease: Super minor, but do we want to add a type to nbins?

python/whylogs/viz/utils/drift_calculations.py
```diff
@@ -183,9 +196,62 @@ def _get_chi2_p_value(target_view_column, reference_view_column) -> Optional[Col
...
+        Reference discrete probability distribution.
+
+    Returns
+    -------
+    float
+        The hellinger distance between the two discrete probability distributions.
+        Between 0 (identical distributions) and 1 (completely different distributions).
+    """
+    # https://en.wikipedia.org/wiki/Hellinger_distance
+    distance = euclidean(np.sqrt(target_pmf), np.sqrt(reference_pmf)) / np.sqrt(2)
```
bernease, line 216: This should be divided by a constant sqrt(2).

python/whylogs/viz/utils/drift_calculations.py
```diff
@@ -183,9 +196,62 @@ def _get_chi2_p_value(target_view_column, reference_view_column) -> Optional[Col
...
+    ref_kll_sketch = ref_dist_metric.kll.value
+
+    if target_kll_sketch.is_empty() or ref_kll_sketch.is_empty():
+        warnings.warn("Distribution sketch must not be empty.")
+        return None
+
+    start = min([target_kll_sketch.get_min_value(), ref_kll_sketch.get_min_value()])
+    end = max([target_kll_sketch.get_max_value(), ref_kll_sketch.get_max_value()])
+    n = target_kll_sketch.get_n() + ref_kll_sketch.get_n()
+    bins, end = _calculate_bins(end=end, start=start, n=n, avg_per_bucket=HIST_AVG_NUMBER_PER_BUCKET, max_buckets=nbins)
```
bernease, line 244: This choosing the number of buckets based on an average number of values per bucket (up to some maximum number of buckets) is new to me. Could we experiment with this using the tests that you're running? I don't see anything wrong with it offhand, but it's just unfamiliar.

bernease (changes requested):
Mostly minor changes, but please do multiply the Hellinger distance by 1/sqrt(2)!

python/whylogs/viz/utils/drift_calculations.py
FelipeAdachi: Yes, that's unfamiliar to me as well, this was carried from v0.

Actually, none of the experiments I ran fall into a case where this avg_per_bucket would be used. But I can run tests to specifically hit that scenario.

python/whylogs/viz/utils/drift_calculations.py
bernease: Nevermind. No idea how I didn't see this. No more early AM reviews.

bernease (approved):
Updating PR to approved because I can not read.

python/whylogs/viz/utils/drift_calculations.py
FelipeAdachi: done in latest commit!

python/whylogs/viz/utils/drift_calculations.py
FelipeAdachi: I will make tests and experiments on this on a follow-up

Issue #902, opened 2022-10-10
ResultSet mergeability


jamie256:
## Description

ResultSet is not mergeable.

## Suggestions

Support merging ResultSet instances so that the complexity of merging multi-profile results can be encapsulated (e.g. if the results are segmented), then it would be easier to use the methods that return ResultSet in a distributed environment.

- [x] I have reviewed the [Guidelines for Contributing](https://github.com/whylabs/whylogs/blob/mainline/.github/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/whylabs/whylogs/blob/mainline/.github/CODE_OF_CONDUCT.md).

jamie256:
Its actually fairly complex with segments, removing the good first issue label.

PR #942, opened 2022-10-28
Add zero() to result_set and DatasetProfileView and test merge

jamie256:
## Description

In a map reduce context, we often need an element we can use a zero element, or seed for accumulation, adding these for DatasetProfileView and ResultSet* with some tests.

## Changes

- DatasetProfileView is already mergeable and adding a zero method makes constructing this easier. Fixed type on timestamps since these can be empty in the underlying DatasetProfile
- Added some tests around merging and checking DatasetProfileView instances are doing the right thing when merging, with zero instances.

## Related

Relates to #902 but doesn't fully address that issue. Will do in a follow up change.


TheMellyBee (approved):
LGTM.

PR #954, opened 2022-11-07
Don't reopen tempfile when writing profiles

### LINKED ISSUE #952: error writing to tempfile on Windows
## Description

Running notebooks locally on windows machines can hit permissions error writing the tempfile while trying to upload to WhyLabs.

jamie256:
Fixes #952 By avoiding Windows NT permissions error when reopening a tempfile

## Description

Add parameters to pass the created tempfile rather than using its name property and then reopening it. 

### DIFF
--- a/python/tests/api/reader/test_local.py
+++ b/python/tests/api/reader/test_local.py
@@ -15,7 +15,8 @@ def reader():
 @pytest.fixture(scope="session")
 def file_path(profile_view):
     tmp_file = NamedTemporaryFile()
-    profile_view.write(path=tmp_file.name)
+    profile_view.write(file=tmp_file)
+    tmp_file.flush()
     yield tmp_file.name
     tmp_file.close()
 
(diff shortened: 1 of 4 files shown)


FelipeAdachi (approved):
LGTM!

But I do have a question (maybe I didn't fully understand it):

I understand that this allows you to pass a bytes object instead of the path when writing a profile. But how is this related to not reopening tempfiles when writing profiles?

jamie256:
> LGTM!
> 
> But I do have a question (maybe I didn't fully understand it):
> 
> I understand that this allows you to pass a bytes object instead of the path when writing a profile. But how is this related to not reopening tempfiles when writing profiles?

Here is some documentation on tempfile, it seems reopening a tempfile based on name is not generally supported, the caller is supposed to use the file returned: https://docs.python.org/3.8/library/tempfile.html

The problem occurs in writers because we first write a tempfile to serialize the profile and then open the tempfile by name to copy it to a remote location (e.g. WhyLabs), so this allows us to maintain that reference to the returned tempfile.

PR #955, opened 2022-11-07
Add separate NaN and inf counts, added tests

### LINKED ISSUE #700: Track NaN and Infinity separately from `null`
## Description

When a system produces NaN or `infinity`, that often indicates bad data or bugs.

## Suggestions

Track these under `n` (count) metric. Maybe under a namespace of counts?

## Related
- Tensorflow tracks the number of `NaN`s: https://github.com/tensorflow/metadata/blob/770c81d26e1bd09faecb2b7940466ad9ddde5483/tensorflow_metadata/proto/v0/statistics.proto#L501

jamie256:
## Description
NaN and inf values can be useful to count separately, add support for splitting these out.

Also added a latency performance test, since some of the None checking was one of the more expensive parts of processing a column.

closes #700 

### DIFF
--- a/python/tests/core/test_performance.py
+++ b/python/tests/core/test_performance.py
@@ -10,6 +10,7 @@ import pandas as pd
 import pytest
 import whylogs_sketching as ds  # type: ignore
 
+import whylogs
 from whylogs.core import ColumnProfile, ColumnSchema
 from whylogs.core.dataset_profile import DatasetProfile
 from whylogs.core.metrics.metrics import MetricConfig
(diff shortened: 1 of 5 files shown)


FelipeAdachi (commented):
I'm thinking this change will affect the Viz' missing cells in the reports:

But I think it could be done in a follow-up.
When it gets merged, I can make an issue for it

FelipeAdachi:
> I'm thinking this change will affect the Viz' missing cells in the reports:
> 
> But I think it could be done in a follow-up. When it gets merged, I can make an issue for it

Actually, nevermind that
I just realized the null count is unaffected by this change

FelipeAdachi (approved):
LGTM!

PR #958, opened 2022-11-08
[feature] Implementing an SQLite Store

murilommen:
## Description

We need an SQLite Store in order to expand usage of our current Profile Store implementation, both in terms of simplicity and perf.

On this PR there was also:
- make `DatasetProfileView.deserialize` use UTC
- modify Fugue's test to comply with `DatasetProfileView.deserialize` having UTC
- allow s3Writer to define configs on init
- make s3Writer return itself


.github/workflows/whylogs-ci.yml
```diff
@@ -85,8 +85,21 @@ jobs:
...
+      - run: |
+          wget https://sqlite.org/2022/sqlite-autoconf-3380500.tar.gz
+          tar -xvf sqlite-autoconf-3380500.tar.gz
+      - run: |
+          cd sqlite-autoconf-3380500
+          ./configure
+          make
+          sudo make install
+          export PATH="/usr/local/lib:$PATH"
+          cd ..
```
naddeoa: Would be better to have an image that just has sqlite, or extend one that we want and install sqlite ontop of it

python/tests/api/store/test_sqlite_store.py
```diff
@@ -0,0 +1,55 @@
...
+    return connection
+
+
+@pytest.fixture
+def store(profile_view, connection):
+    store = SQLiteStore(connection=connection)
+    store.write(profile_view=profile_view, profile_name="my_profile")
+    store.write(profile_view=profile_view, profile_name="my_profile")
+    yield store
+    del store
```
naddeoa: might be worth considering the with pattern. Isn't that how people do resource auto closing in Python?

python/whylogs/api/store/s3.py
```diff
@@ -1,5 +1,66 @@
...
+                for profile_path in profiles_list:
+                    try:
+                        profile_result = self.s3_reader.option(
+                            bucket_name=self.bucket_name, object_name=profile_path
+                        ).read()
+                        profile_view = profile_view.merge(profile_result.view())
+                        logger.debug(f"Profile {profile_path} successfully read")
+                    except Exception as e:
+                        logger.error("Could not read the file {}".format(e))
+                return profile_view
```
naddeoa: Can move to a branch until its 100% done

python/whylogs/api/store/sqlite_store.py
```diff
@@ -0,0 +1,65 @@
...
+from whylogs.api.store import ProfileStore
+from whylogs.api.store.query import DateQuery, ProfileNameQuery
+from whylogs.core import DatasetProfile, DatasetProfileView
+
+logger = logging.getLogger(__name__)
+
+
+class SQLiteStore(ProfileStore):
+    def __init__(self, connection: Optional[sqlite3.Connection] = None):
+        self._db_location = os.getenv("SQLITE_STORE_LOCATION") or ":memory:"
```
naddeoa: Would be great to start a convention early for a Config type that centralizes all of the different env configs

naddeoa: I think make the env var required

python/whylogs/api/store/sqlite_store.py
```diff
@@ -0,0 +1,65 @@
...
+from typing import Optional, Union
+
+from whylogs.api.store import ProfileStore
+from whylogs.api.store.query import DateQuery, ProfileNameQuery
+from whylogs.core import DatasetProfile, DatasetProfileView
+
+logger = logging.getLogger(__name__)
+
+
+class SQLiteStore(ProfileStore):
```
naddeoa, line 13: Take a look at my sqlite wrapper: https://github.com/whylabs/whylogs-container/blob/mainline/src/main/kotlin/ai/whylabs/services/whylogs/util/SqliteManager.kt

Had to do a lot of trial and error to improve perf and landed on this

python/whylogs/api/store/sqlite_store.py
```diff
@@ -0,0 +1,65 @@
...
+            self.conn.commit()
+            logger.debug("Profile successfully written to DB!")
+        except sqlite3.Error as e:
+            logger.error(f"Error: {e}")
+
+    def write(self, profile_view: DatasetProfileView, profile_name: str):
+        init_db_query = (
+            f"CREATE TABLE IF NOT EXISTS {profile_name}(id integer primary key, date datetime, profile blob);"
+        )
+        self.cur.execute(init_db_query)
```
naddeoa: I think profile names have to not matter. We probably just want a single table that has a reference to dataset id as well, and every column that exists would only exist to make querying easier. So, only one table.

python/whylogs/api/store/sqlite_store.py
```diff
@@ -0,0 +1,65 @@
...
+            sql_query = f"SELECT profile FROM '{query.profile_name}' WHERE date BETWEEN '{query.start_date}' AND '{query.end_date}';"
+            response = self.cur.execute(sql_query)
+        profile_view = DatasetProfile().view()
+        for item in response.fetchall():
+            profile_view.merge(DatasetProfileView.deserialize(item[0]))
+        return profile_view
+
+    def _insert_blob(self, profile_view: bytes, profile_name: str):
+        try:
+            profile_date = datetime.utcnow()
```
naddeoa: Should be the dataset_timestamp in the profile

python/whylogs/api/store/sqlite_store.py
```diff
@@ -0,0 +1,65 @@
...
+            data_tuple = (profile_date, profile_view)
+            query = f"INSERT INTO {profile_name} (date, profile) VALUES (?, ?);"
```
naddeoa: Should switch this over to "prepared queries" that we can reuse rather than doing string sub

naddeoa (commented):
Sqlite perf stuff isn't a blocker, we can always optimize that after we get everything working, it shouldn't change everything else.

python/tests/api/store/test_sqlite_store.py
murilommen: accepts both ways now :)

python/whylogs/api/store/sqlite_store.py
murilommen: done on service-side for now, not to blow up whylogs' (i figure it needs to be more discussed, since other things could also benefit from a central config)

python/whylogs/api/store/profile_store.py
```diff
@@ -12,11 +12,11 @@ def list(self) -> List[str]:
         pass
 
     @abstractmethod
-    def get(self, query: DateQuery) -> DatasetProfileView:
+    def get(self, query: Union[DateQuery, DatasetIdQuery]) -> DatasetProfileView:
```
richard-rogers: Perhaps a base class rather than a union?

murilommen: in here i've tried to avoid [dataclass inheritance](https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses). but if you think it's still better to inherit, I can make them regular classes :)

python/whylogs/api/store/sqlite_store.py
```diff
@@ -0,0 +1,126 @@
...
+from whylogs.core import DatasetProfile, DatasetProfileView
+
+logger = logging.getLogger(__name__)
+
+
+class SQLiteStore(ProfileStore):
+    def __init__(self):
+        self._db_location = os.getenv("SQLITE_STORE_LOCATION")
+        if self._db_location is None:
+            raise ValueError("You must define a store with the SQLITE_STORE_LOCATION env var")
```
naddeoa, line 17: We should have a task to centralize all of the config logic behind a single class/interface. Not crucial atm but not having it definitely leads to code duplication, difficulty knowing which parameters are required/exist, and difficulty mocking/testing things when they depend directly on the environment.

python/whylogs/api/store/sqlite_store.py
```diff
@@ -0,0 +1,126 @@
...
+        try:
+            self.cur.close()
+            self.conn.close()
+        except Exception as e:
+            logger.debug(f"Connection not established. Error: {e}")
+
+    def _init_db(self):
+        init_db_query = "CREATE TABLE IF NOT EXISTS profile_store(id string, date datetime, profile blob);"
+        self.cur.execute(init_db_query)
+
```
naddeoa: We should be able to just init this once right? No need to execute an additional query before all of the real work, even if it would be a no-op

python/tests/api/store/test_sqlite_store.py
```diff
@@ -0,0 +1,129 @@
...
+    @classmethod
+    def teardown_class(cls):
+        del os.environ["MERGE_PROFILE_PERIOD_HOURS"]
+        del os.environ["SQLITE_STORE_LOCATION"]
+
+    @pytest.fixture
+    def store(self, profile_view):
+        store = SQLiteStore()
+        store.cur.execute("CREATE TABLE IF NOT EXISTS profile_store(id string, date datetime, profile blob);")
+        query = "INSERT INTO profile_store (id, date, profile) VALUES (?, ?, ?);"
```
naddeoa: This seems weird to me. I don't think this thing should expose the ability to directly execute sql, and we shouldn't need to execute sql to test anything that its doing.

python/whylogs/api/store/query.py
```diff
@@ -1,21 +1,36 @@
...
 
-@dataclass
-class DateQuery:
-    profile_name: str
-    start_date: datetime
-    end_date: Optional[datetime] = field(default=None)
+class BaseQuery(ABC):
+    def __init__(self):
+        self.dataset_id = None
+
```
naddeoa: Any long term vision of what these things will actually share? It seems like the base query shouldn't actually have a dataset_id because I can imagine other types of queries that don't care about that. DateQuery doesn't seem like it should care about that, just date ranges. It would be weird to have a base class and then manually specify all of the things the subclasses have in common too.

An alternative is to not have a base class and just export a type alias 

```python
Query = Union[DateQuery, DatasetIdQuery]
```

python/whylogs/api/store/sqlite_store.py
murilommen: my idea is to leave this to the Profile Store Service only, at least for now, where a config file will be very useful and I can also see more stuff on whylogs in general - like constraints definition - that could use something like a YAML or w/e, so I would prefer to build something at once then

python/whylogs/api/store/query.py
murilommen: if you have two Projects/Dataset Ids - A and B - persisted to the same store, how do you query A by a date range?

python/whylogs/api/store/sqlite_store.py
murilommen: right! will move it onto the init method

python/tests/api/store/test_sqlite_store.py
murilommen: in here i am avoiding to execute the store methods to insert stuff onto the store using the same connection, but at the same time populating its database before actually making it available to the tests.
making this to be able to compare expected vs. actual on `store.list`, and i dont want to `store.write` to later assert `store.list` returns me what it has written -- otherwise those wouldn't be unit tests

lmk how i could do this differently or better though

python/whylogs/api/store/query.py
murilommen: btw, made the suggested change to the classes :)

python/whylogs/api/store/sqlite_store.py
murilommen: my idea is to have it all under a `config.ini`, but only on the profile-store-service, as it's where it will actually make sense to use, since other stuff also require env vars -- such as the s3writer class we rely on for the service.

naddeoa: That's fine. I'm not talking about the way that users specify the config, only the way that we read the config. I've got a similar pattern going on in the whylogs container here 

https://github.com/whylabs/whylogs-container/blob/mainline/src/main/kotlin/ai/whylabs/services/whylogs/core/config/EnvVars.kt

It's just a wrapper that eliminates the need to embed string env names all over the code base and allow for injecting alternatives.

python/whylogs/api/store/query.py
naddeoa: You would also have an AND/OR abstraction to enable more combinations of things. So that might look like this:

```python
filter = Or(
    And(DateQuery(...), IdQuery(id1)),
    And(DateQuery(...), IdQuery(id2)),
)

store.read(filter)
```

python/tests/api/store/test_sqlite_store.py
naddeoa: Why do you want to avoid using the store to populate the store? That's a thing that it should be able to do. The alternative to that is copy/pasting the table creation query into the unit test (and re-copy/pasting it whenever we change the query in the store). It has to match exactly in order to work at all and its invalid for an end user to touch the sql directly. Here are the properties that don't work here:

- User being able to directly execute SQL.
- Tests having to exactly duplicate, by copy/pasting, table creation logic and table schema
- Tests having to exactly duplicate insert statements

If the store just ensures that tables exist after its made and exposes the appropriate methods for inserting things (which I think are both true) then we aren't losing anything. I know its Python so we probably just have to slap a few underscores infront of the conn/cur vars, though I am tempted to use scope to make it really private like 


```python
def make_hi():
  private_var = 'var'
  class Hi:
    def do_stuff():
      print(private_var)

  return Hi()

hi = make_hi()
hi.do_stuff()
```

That might be an anti pattern.

python/whylogs/api/store/query.py
```diff
@@ -1,36 +1,26 @@
...
-        self.dataset_id = dataset_id
-        self.start_date = start_date
-        self.end_date = end_date
-        self.segment_tag = segment_tag
+@dataclass
+class DateQuery:
+    dataset_id: str
+    start_date: datetime
+    end_date: Optional[datetime] = field(default=None)
+    segment_tag: Optional[str] = field(default=None)
```
naddeoa, line 11: I think segment_tag falls into the same bucket. Looks like we can either make a dedicated query type for it or forever copy/paste it to every query type we make.

python/whylogs/api/store/query.py
```diff
@@ -1,36 +1,26 @@
...
-    ):
-
-        super().__init__()
-        self.dataset_id = dataset_id
-        self.start_date = start_date
-        self.end_date = end_date
-        self.segment_tag = segment_tag
+@dataclass
+class DateQuery:
+    dataset_id: str
```
naddeoa, line 8: Meant to remove this since you added the DatasetIdQuery?

python/whylogs/api/store/query.py
murilommen: oh, ok. 
my idea is slightly different, where you'll only need one type of Query object in order to get what you need from the store.
so in your example it would be something like

```python
filter = Or(
    DateQuery(date_range, id1),
    IdQuery(id2),
)
```

python/tests/api/store/test_sqlite_store.py
murilommen: > Why do you want to avoid using the store to populate the store? That's a thing that it should be able to do.

In order to make the tests actually unitary, I didn't want to have a "recursion" with the fixtures. 
So if I for some reason break or modify the way `store.write` works, `store.list` remains intact in the way it is tested and will continue to pass on the checks, and it may be easier to make the error be pointed out to where it's failing - if it ever comes to this point. But maybe that's wrong, that's just how I learned to think when creating unit tests in general. I might have taken it too literally on this. 

> It has to match exactly in order to work at all and its invalid for an end user to touch the sql directly.

With all that being said, all of this makes a ton of sense. Definitely needs improvement.

python/whylogs/api/store/query.py
murilommen: resolving this conversation as this turned out to be a future change, described in: https://github.com/whylabs/whylogs/issues/1003

python/whylogs/api/store/query.py
murilommen: added an issue for it https://github.com/whylabs/whylogs/issues/1003

python/whylogs/api/store/query.py
murilommen: removing `segment_tag` for now as we will need to flesh this out a little more on the following PRs

python/whylogs/api/store/sqlite_store.py
murilommen: also making this a new issue to be addressed separately https://github.com/whylabs/whylogs/issues/1004

python/tests/api/store/test_sqlite_store.py
murilommen: latest commit addresses this :)

naddeoa (approved):

PR #967, opened 2022-11-10
Fluent condition predicate API

richard-rogers:
Replaces the predicate lambdas with an actual class.
* Adds fluent API for predicate construction
* Adds [de]serialization for predicates
* Adds accessors to write predicates agains metrics & profiles


python/whylogs/core/relations.py
```diff
@@ -0,0 +1,167 @@
...
+            return self._udf(x)
+        if op == Relation._and:
+            return self._left(x) and self._right(x)
+        if op == Relation._or:
+            return self._left(x) or self._right(x)
+        if op == Relation._not:
+            return not self._right(x)
+
+        raise ValueError("Unknown ConditionCountMetric predicate")
+
```
richard-rogers, line 132: Are we happy with these names?

python/whylogs/core/relations.py
```diff
@@ -0,0 +1,167 @@
...
+        return f"{_TOKEN[self._op.value]} {self._left.serialize()} {self._right.serialize()}"
+
+    def __str__(self) -> str:
+        if not (self._left or self._right):
+            return f"x {_TOKEN[self._op.value]} {self._value.serialize()}"
+        if self._op == Relation._not:
+            return f"not {str(self._right)}"
+        return f"({str(self._left)}) {_TOKEN[self._op.value]} ({str(self._right)})"
+
+def Not(p: Predicate) -> Predicate:
```
richard-rogers, line 188: maybe `def not_(...)` ?

python/tests/core/metrics/test_condition_count.py
```diff
@@ -262,3 +263,65 @@ def test_condition_count_in_dataset_profile() -> None:
...
+    assert summary["above_min"] == len(data) - 1
+
+
+def test_metric_getter() -> None:
+    data = [1, 2, 3, 4, 5]
+    dist_metric = DistributionMetric.zero(MetricConfig())
+    dist_metric.columnar_update(PreprocessedColumn.apply(data))
+
+    conditions = {
+        "above_min": Condition(X.greater_than(MetricGetter(dist_metric, "min"))),
```
richard-rogers, line 351: Nicer syntax for instantiating `MetricGetter` and `ProfileGetter` ?

python/whylogs/core/relations.py
TheMellyBee: Idk I like the not equal for ease

python/whylogs/core/relations.py
TheMellyBee: I'm not a fan of leq cause I have to rethink it every time. less_eq is worth it to me. (And for all the others)

python/whylogs/core/relations.py
TheMellyBee: Oh yeah, this does seem a bit weird. hmm..... I think it's ok though. Negate could work better

python/tests/core/metrics/test_condition_count.py
TheMellyBee: It's okish. I think I would be more likely to type in Condition(X.greater_than, Metric Getter) then in condition do that. but that's just cause I'm used to sending in separate parameters.

python/whylogs/core/relations.py
richard-rogers: I was referring to the fluent factory methods, not the `Relation` enumeration values since those aren't user-visible.

python/whylogs/core/relations.py
richard-rogers: I think some spelling of `not` would be more familiar to programmers. `not_` seems more consistent with `and_` and `or_`. I just did `Not` to be consistent with `X` and not the reserved `not` keyword...

python/tests/core/metrics/test_condition_count.py
richard-rogers: The `ValueGetter` (:> `MetricGetter`, `ProfileGetter`, `LiteralGetter`) has to be an argument to the `Predicate` rather than the `Condition` to support boolean compound predicates. E.g., `X.less_than(MetricGetter(distribution, "q_75").and_(X.greater_than(MetricGetter(distribution, "mean"))`.  The `Predicate`s can also be used in contexts where there are no `Condition`s.

python/whylogs/core/relations.py
TheMellyBee: Ooh ok, this didn't make sense where it was pointing at. GOt it. I do like those :) I'm happy with them

python/whylogs/core/relations.py
TheMellyBee: Ok, I still think 'negate' would be more clear while keeping the still and not doing the keyword. But I'd be ok with Not

python/tests/core/metrics/test_condition_count.py
TheMellyBee: Ahh yes that make sense.

TheMellyBee (dismissed):
I'm happy with this. Put comments but they aren't blocking. Really neat way to bring in an expression way of work

python/whylogs/core/relations.py
```diff
@@ -0,0 +1,174 @@
...
+        else:
+            self._value = value  # type: ignore
+
+        if op in {Relation._and, Relation._or}:
+            if not left and right:
+                raise ValueError("boolean operators require two operands")
+        self._left = left
+        self._right = right
+
+        if op == Relation._not:
```
richard-rogers: I don't think this is needed

FelipeAdachi (commented):
I'm running the Condition Validators example from mainline with the updated changes to test for backcompatibility.

When reaching this cell (Under Section Apply Validators to Data):
```python
from whylogs.core.schema import DatasetSchema
import whylogs as why

schema = DatasetSchema(validators=validators)
profile = why.log(df, schema=schema).profile()
```

I'm getting this error:

```
  File "/mnt/c/Users/felip/Documents/Projects-WhyLabs/whylogs/python/whylogs/core/validators/condition_validator.py", line 33, in columnar_validate
    valid = condition(x)
TypeError: 'Condition' object is not callable
```

FelipeAdachi (commented):
Can we have a test for using UDF in Predicate?

python/whylogs/core/relations.py
```diff
@@ -0,0 +1,193 @@
...
+    @abstractmethod
+    def __call__(self) -> Union[str, int, float]:
+        raise NotImplementedError
+
+    @abstractmethod
+    def serialize(self) -> str:
+        raise NotImplementedError
+
+
+class LiteralGetter(ValueGetter):
```
FelipeAdachi, line 20: can we have a test for LiteralGetter?

richard-rogers:
> Can we have a test for using UDF in Predicate?

There are a few. See the tests with `is_()`

python/whylogs/core/relations.py
richard-rogers: Almost all the tests exercise `LiteralGetter`.

richard-rogers:
> I'm running the Condition Validators example from mainline with the updated changes to test for backcompatibility.

Should be fixed now :)

FelipeAdachi (commented):
Metric Constraints' `conditons` is a different concept than Condition Count Metrics' `Condition`: The first one is actually a `Relation` if I'm understanding correctly.

This is very confusing terminology. Do we want to do something about it? E.g. deprecating  Metric Constraints `condition` in favor of a `relation` parameter?

Either way, I think this is out of scope for this PR. It's just something I noticed while reviewing this PR.

FelipeAdachi (approved):
my only comment on terminology is not a blocker, so LGTM!

richard-rogers:
> Metric Constraints' `conditons` is a different concept than Condition Count Metrics' `Condition`: The first one is actually a `Relation` if I'm understanding correctly.

I think the `MetricConstraint::condition` may be closer to `Predicate`.

> This is very confusing terminology. Do we want to do something about it? E.g. deprecating Metric Constraints `condition` in favor of a `relation` parameter?

"Condition," "relation," and "predicate" are all a bit muddled unfortunately. I'm not sure we can clean it up in a backward compatible way, but it's worth some thought :)

PR #968, opened 2022-11-11
add method for adding standard metrics to existing Resolver

FelipeAdachi:
## Description

It is cumbersome to always have to create a CustomResolver from scratch, even when you might be interested in adding a single metric to the standard resolver. This enables the user to start from an existing resolver and add a Standard Metric according to column name or data type.

## Changes

## Related


TheMellyBee (dismissed):
Good idea! Much nicer

python/whylogs/core/resolvers.py
```diff
@@ -78,4 +124,10 @@ def resolve(self, name: str, why_type: DataType, column_schema: ColumnSchema) ->
         result: Dict[str, Metric] = {}
         for m in metrics:
             result[m.name] = m.zero(column_schema.cfg)
+
+        if self.additional_metrics:
+            result = self.resolve_additional_standard_metrics(
```
richard-rogers: Should these be added to standard result rather than replacing it?

python/whylogs/core/resolvers.py
```diff
@@ -11,11 +12,45 @@
...
 
+    def __init__(self):
+        self.additional_metrics: List[AdditionalMetric] = []
+
+    def add_standard_metric(
+        self,
+        metric: StandardMetric,
+        column_names: List[String] = [],
+        why_types: List[Union[Fractional, Integral, String]] = [],
+    ):
```
jamie256: Ok, so this can only add a new `StandardMetric`, that wouldn't already resolve for this type or column name on a resolver, so its rather limited functionally on the `StandardResolver `since you would only be able to add a `StandardMetric`, I guess the main use case here is the `unicode_range `and `condition_count `metrics?

Also its a bit confusing what happens when I add the same `StandardMetric` that would already be resolved, or the same `StandardMetric` type gets added multiple times, perhaps bound to a type and then bound to a column name? It looks like they can both be added but only one will be functional, so maybe it doesn't break things but slows down resolving column schema?

python/whylogs/core/resolvers.py
```diff
@@ -67,6 +110,14 @@ def resolve(self, name: str, why_type: DataType, column_schema: ColumnSchema) ->
         result: Dict[str, Metric] = {}
         for m in metrics:
             result[m.name] = m.zero(column_schema.cfg)
+
+        if self.additional_metrics:
+            additional_metrics_result = self.resolve_additional_standard_metrics(
+                result=result, name=name, why_type=why_type, column_schema=column_schema
+            )
+            for key in additional_metrics_result.keys():
+                result[key] = additional_metrics_result[key]
```
jamie256: you can do:
```suggestion
            result.update(additional_metrics_result)
```

python/whylogs/core/resolvers.py
```diff
@@ -78,4 +129,12 @@ def resolve(self, name: str, why_type: DataType, column_schema: ColumnSchema) ->
...
```
jamie256:
```suggestion
            result.update(additional_metrics_result)
```

python/whylogs/core/resolvers.py
```diff
@@ -11,11 +12,45 @@
...
+    def resolve_additional_standard_metrics(
+        self, result: Dict[str, Metric], name: str, why_type: DataType, column_schema: ColumnSchema
+    ) -> Dict[str, Metric]:
+        additional_result = {}
+        for additional_metric in self.additional_metrics:
+            for column_name in additional_metric.column_names:
+                if column_name == name and additional_metric.metric.name not in result:
+                    additional_result[additional_metric.metric.name] = additional_metric.metric.zero(column_schema.cfg)
+            for target_why_type in additional_metric.why_types:
+                if isinstance(target_why_type, type(why_type)) and additional_metric.metric.name not in result:
```
jamie256: So this looks like we can only add one ConditionCountMetric? How would I add both a condition count that checks string columns for email pattern, and a condition count that checks the "Fico Score" column is greater than 400? It looks like here the additional_metric.metric.name would just be "condition_count" and limits me to only have one of these?

python/whylogs/core/resolvers.py
```diff
@@ -11,11 +12,45 @@
...
+    def add_standard_metric(
+        self,
+        metric: StandardMetric,
+        column_names: List[String] = [],
+        why_types: List[Union[Fractional, Integral, String]] = [],
+    ):
+        if not column_names and not why_types:
+            raise ValueError("Either column names or why types must not be empty.")
+        if column_names and why_types:
+            raise ValueError("column_names or why_types should be defined, not both.")
```
jamie256: why can't I add both why_type and column_names? Maybe we don't need that flexibility but maybe its interesting to "Add `frequent_items` to columns named `dti_ratio` or columns of type `string`"

python/whylogs/core/resolvers.py
```diff
@@ -67,6 +110,14 @@ def resolve(self, name: str, why_type: DataType, column_schema: ColumnSchema) ->
...
             result[m.name] = m.zero(column_schema.cfg)
+
+        if self.additional_metrics:
+            additional_metrics_result = self.resolve_additional_standard_metrics(
+                result=result, name=name, why_type=why_type, column_schema=column_schema
+            )
+            for key in additional_metrics_result.keys():
+                result[key] = additional_metrics_result[key]
+
         return result
```
jamie256, line 140: Consider adding an EmptyResolver? I do like the HistogramCountingTrackingResolver as a useful base to add onto but it might be better to allow the caller to skip most columns for a very lightweight profile.

jamie256 (commented):
Mostly questions, changes look good.

python/whylogs/core/resolvers.py
richard-rogers: A column can only have one `ConditionCountMetric`, but the metric instance can host many conditions (we should probably put a configurable limit on how many ;).

FelipeAdachi: but, right now, there's no easy way to assign different metric instances to different columns, right?

Do we want to change this before merging this PR?

python/whylogs/core/resolvers.py
FelipeAdachi: With latest changes, this now expects a SpecResolver, which accepts any kind of Metrics, not just Standard ones.

If the metric to be added is already present in the column's result while resolving the additional specs, it will just skip it.

If both a name and type is passed, the name takes precedence, and will only be added once

python/whylogs/core/resolvers.py
FelipeAdachi: thanks! much better, added in the latest commit

python/whylogs/core/resolvers.py
FelipeAdachi: With the latest changes, adding different conditions to different columns is easier, by using ResolverSpecs.

There's a test for this case in test suite

python/whylogs/core/resolvers.py
FelipeAdachi: yes, now passing both just raises a warning, saying that the name takes precedence.

python/whylogs/core/resolvers.py
```diff
@@ -16,11 +16,85 @@
...
+                raise ValueError("Resolver Spec: must supply a Metric subclass to MetricSpec")
+        self.additional_specs.append(resolver_spec)
+
+    def resolve_additional_specs(
+        self, result: Dict[str, Metric], name: str, why_type: DataType, column_schema: ColumnSchema
+    ) -> Dict[str, Metric]:
+        """Resolves resolver specifications added by add_resolver_spec
+
+        Parameters
+        ----------
```
richard-rogers: This doctoring doesn't appear to be accurate

python/whylogs/core/resolvers.py
```diff
@@ -16,11 +16,85 @@
...
+        Returns
+        -------
+        Dict[str, Metric]
+            Dictionary of metrics related to the additional resolver specificiations.
+        """
+        additional_result = {}
+        for resolver_spec in self.additional_specs:
+            col_name, col_type = resolver_spec.column_name, resolver_spec.column_type
+
+            if (col_name and col_name == name) or (col_name is None and isinstance(why_type, col_type)):  # type: ignore
```
richard-rogers: This looks very similar to the code in `DeclarativeResolver`, and doesn't respect the {en,dis}able members of the `MetricConfig`. Perhaps refactor to share code with `DeclarativeSchema`?  Also, maybe this should be an abstract method in the `Resolver` base class?

python/whylogs/core/resolvers.py
richard-rogers: `DeclarativeSchema` is also a handy way to construct a lightweight profile :)

python/whylogs/core/resolvers.py
```diff
@@ -54,6 +128,13 @@ def resolve(self, name: str, why_type: DataType, column_schema: ColumnSchema) ->
         result: Dict[str, Metric] = {}
         for m in metrics:
             result[m.name] = m.zero(column_schema.cfg)
+
+        if self.additional_specs:
```
richard-rogers: This code seems to be repeated in several places. Consider refactoring into a common implementation.

python/whylogs/core/resolvers.py
```diff
@@ -16,11 +16,85 @@
...
+    def add_resolver_spec(self, resolver_spec: ResolverSpec):
+        """Add resolver specification to existing resolver instance.
+
+        Parameters
+        ----------
+        resolver_spec : ResolverSpec
+            Resolver specification that define the metrics to instantiate for matching columns.
+        """
+        if resolver_spec.column_name and resolver_spec.column_type:
+            logger.warning(
```
richard-rogers: We should move the `ResolverSpec` validation into `ResolverSpec::__post_init__()`.

python/whylogs/core/resolvers.py
richard-rogers: I thought about passing both when designing the `ResolverSpec`. One option would be to signal a warning or error if the specified column name didn't match the specified type (name AND type match semantics). OR match semantics seemed weird, but you can implement it with 2 `ResolverSpec`s if that's what you really want.

python/whylogs/core/resolvers.py
FelipeAdachi: This was added in latest commit!

python/whylogs/core/resolvers.py
FelipeAdachi: With latest changes, the validation is made by `DeclarativeResolver` itself. I'm ok with how it works currently. Let me know if we want to change it.

python/whylogs/core/resolvers.py
FelipeAdachi: Absolutely!

> Perhaps refactor to share code with DeclarativeSchema?

With latest changes, I'm now using DeclarativeResolver to resolve the additional metrics.

> Also, maybe this should be an abstract method in the Resolver base class?

I changed `resolve_additional_specs` to be abstract in Resolver Base Class, and called `super().resolve_additional_specs` in each subclass. Is that what you meant?

But wouldn't this be a breaking change, since that requires us to change every custom resolver created on the examples/tests to add that method?

python/whylogs/core/resolvers.py
FelipeAdachi: In latest changes, I moved this to `resolve_additional_specs` and directly update the result with:

`result.update(self.resolve_additional_specs(name=name, why_type=why_type, column_schema=column_schema))`

Is this better, or do you suggest something else?

python/whylogs/core/resolvers.py
FelipeAdachi: Now that I'm using `DeclarativeResolver` to resolve the additional metrics, I removed this validation. I agree that the validation from `DeclarativeResolver` could be moved to `ResolverSpec::__post_init__()`.

I'm ok with doing this in a follow-up, though. lmk if that's ok

richard-rogers: It will be a slightly breaking change, so it would be good to fix it before `DeclarativeSchema` gets released Tuesday :(

python/tests/core/test_resolver.py
```diff
@@ -0,0 +1,159 @@
...
+        metrics=[
+            MetricSpec(
+                ConditionCountMetric,
+                ConditionCountConfig(conditions=not_4_3_condition),
+            ),
+        ],
+    )
+
+    resolver.add_resolver_spec(legs_not_4_spec)
+    resolver.add_resolver_spec(weights_not_4_2_spec)
```
richard-rogers, line 66: Maybe change `add_resolver_spec` to `add_resolver` or `add_metric`? Maybe pass in `Union[ResolverSpec, List[ResolverSpec]]` ?

FelipeAdachi:
Closing this as #1016 is an updated version of this PR

PR #974, opened 2022-11-14
Java: All the Protobuf

TheMellyBee:
## Description
We need protobuf to communicate for cross platform/language work, thus this protobuf needs to match with the Python v1.

Couple of note this was a skeleton that got put off for other work and then came back to it for the rest (hence the big date gap). Second, long term I want to use an annotation factory to create the metrics and components instead of the switch statement, but for this first release and after talking with @jamie256 and @richard-rogers we will keep to this simple model and refactor in the coming quarter. 

## Changes
- Adds the to/from protobuf to all methods of the tree 
- Adds in serializers and deserializer registries 

## Related
Continues Java development and unblocks branch for readers and writers


java/core/src/main/java/com/whylogs/core/metrics/IntegralMetric.java
```diff
@@ -110,4 +112,21 @@ public IntegralMetric merge(Metric<?> other) {
...
+    for (MetricComponentMessage componentMessage : message.getMetricComponentsMap().values()) {
+      MetricComponent<?> component = MetricComponent.fromProtobuf(componentMessage);
+      components.put(component.getTypeName(), component);
+    }
+
+    MaxIntegralComponent max = (MaxIntegralComponent) components.get("MaxIntegralComponent");
+    MinIntegralComponent min = (MinIntegralComponent) components.get("MinIntegralComponent");
+
+    if (max == null || min == null) {
+      throw new IllegalArgumentException("IntegralMetric must have max and min components");
```
TheMellyBee, line 127: Let's make this an assertion as it indicates a code problem.

richard-rogers (approved):

PR #980, opened 2022-11-18
default performance metrics log full data false

jamie256:
## Description

Splitting out smaller change to stop the default of logging full input data when logging performance metrics, minor refactor and adding optional parameter to include logging the full input.


FelipeAdachi (commented):
One minor question:

The old behavior for log_regression_metrics and log_classification_metrics was to log the full data, right? Isn't it possible that users might be confused if the same call doesn't log the full data anymore?

FelipeAdachi (approved):
LGTM!

jamie256:
> One minor question:
> 
> The old behavior for log_regression_metrics and log_classification_metrics was to log the full data, right? Isn't it possible that users might be confused if the same call doesn't log the full data anymore?

It is a behavior change but I'd characterize this as a bug fix, we need to highlight this in the next release, and invite users to set the `log_full_data` parameter if they want to include profiling the inputs, but callers have to explicitly opt in as the production use case generally has the targets come in later in time for predictions and inputs that were previously already profiled.

PR #990, opened 2022-11-23
Add multiclass model performance metric APIs

bernease:
## Description
Add multiclass classification APIs.

## Changes

- Add two APIs for tracking confusion matrix for multiclass classification model performance. One accepts pandas dataframe + column names (with optional full data logging), other accepts nested arrays or numpy arrays.

## Related


python/whylogs/api/logger/__init__.py
```diff
@@ -78,6 +78,51 @@ def log_classification_metrics(
...
+    data: pd.DataFrame,
+    target_columns: List[str],
+    prediction_columns: List[str],
+    score_columns: List[str],
+    labels: Optional[List[Union[str, bool, float, int]]] = None,
+    schema: Optional[DatasetSchema] = None,
+    log_full_data: bool = False,
+) -> ProfileResultSet:
+    if data is not None:
+        if targets is None:
```
FelipeAdachi, line 146: there is no `targets` variable up until this point, right?
Do we need to check if it is `None`?

python/whylogs/api/logger/__init__.py
```diff
@@ -78,6 +78,51 @@ def log_classification_metrics(
...
+            targets = data.loc[:, target_columns].values.tolist()
+        if predictions is None:
```
FelipeAdachi, line 148: same question as above (also for `scores` right below)

python/whylogs/api/logger/__init__.py
```diff
@@ -78,6 +78,51 @@ def log_classification_metrics(
     return _log_with_metrics(data=data, metrics=model_performance_metrics, schema=schema, include_data=log_full_data)
 
 
+def log_multi_classification_metrics(
```
FelipeAdachi, line 136: Should we have a test for this? Also some docstrings with an example (could be the same for both test and docstring)

python/whylogs/api/logger/__init__.py
```diff
@@ -78,6 +78,51 @@ def log_classification_metrics(
...
+        predictions=predictions,
+        targets=targets,
+        scores=scores,
+        labels=labels,
+    )
+
+    return _log_with_metrics(data=data, metrics=model_performance_metrics, schema=schema, include_data=log_full_data)
+
+
+def log_multi_classification_metrics_raw(
```
FelipeAdachi, line 164: Same comment of testing/docstrings as above

python/whylogs/core/model_performance_metrics/confusion_matrix.py
```diff
@@ -92,12 +92,55 @@ def add(
...
+        if scores is None:
+            if _multiclass:
+                score_row = [1.0] * len(self.labels)
+                scores = [score_row] * len(targets)
+            else:
+                scores = [1.0] * len(targets)
+
+        if _multiclass:
+            if isinstance(scores[0], list):
+                print("received multidimensional scores object")
```
FelipeAdachi, line 112: should we have prints in the code? maybe a warning?

python/whylogs/core/model_performance_metrics/confusion_matrix.py
```diff
@@ -92,12 +92,55 @@ def add(
...
+                if any(length != len(self.labels) for length in [len(row) for row in scores]):
+                    raise ValueError("scores shape must be (n, d) for n rows and d unique labels")
+                # TODO: Until capable of handling new message format, convert scores
+                # to a single list with only predicted class. This will not result in
+                # the correct ROC and score-related values, but confusion matrix accurate.
+                scores = [row[row.index(max(row))] for row in scores]
```
FelipeAdachi, line 118: if for now this leads to incorrect ROC and score-related results, show we warn the user for this case also? I'm not sure if the ROC and score-related values is calculated in whylabs. If so, maybe the warning would be better placed there?

Might be missing context here, just speculating.

FelipeAdachi (commented):
leaving some overall comments!

Issue #1009, opened 2022-12-05
image metric is not correctly registered for deserialization from DatasetProfileView


jamie256:
## Description

You can repro the issue reported by adding this at the end of the Image_Logging notebook (or replace the WhylabsWriter part):
https://colab.research.google.com/github/whylabs/whylogs/blob/mainline/python/examples/advanced/Image_Logging.ipynb
```
from whylogs.core import DatasetProfileView
view.write(path="test.bin")
deserialized_view = DatasetProfileView.read("test.bin")
```

Which results in an error:

```
column_profile_view.py:129, in ColumnProfileView.from_protobuf(cls, msg)
    127         metric_class = _METRIC_DESERIALIZER_REGISTRY[m_name]
    128     else:
--> 129         raise UnsupportedError(f"Unsupported metric: {m_name}")
    130 else:
    131     metric_class = m_enum.value


UnsupportedError: Unsupported metric: image
```

## Suggestions

Modify ImageMetric to correctly register an entry in _METRIC_DESERIALIZER_REGISTRY and allow MultiMetric support for deserialization from ColumnProfileView

- [x] I have reviewed the [Guidelines for Contributing](https://github.com/whylabs/whylogs/blob/mainline/.github/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/whylabs/whylogs/blob/mainline/.github/CODE_OF_CONDUCT.md).

richard-rogers:
`MultiMetric`is not a dataclass, so it isn't calling the `Metric::__post_init__()`. Also, since the registration only happens in the metric constructors, the non-standard metrics aren't registered until at least one gets instantiated. 

I'll add code that registers them on `import`.

PR #1010, opened 2022-12-06
Register non-standard metrics for deserialization 

richard-rogers:
## Description

Register non-standard metrics for custom deserialization.

`MultiMetric` subclasses (`ImageMetric` and `UnicodeRangeMetric`) are not dataclasses, so they were not registering for deserialization. `ConditionCountMetric` is, but for all of these metrics registration happened in the constructor so they couldn't be deserialized unless one had been instantiated.

## Changes

- Register non-standard metrics on `import`

## Related

Relates to #1009

### DIFF
--- a/python/whylogs/core/metrics/metrics.py
+++ b/python/whylogs/core/metrics/metrics.py
@@ -154,6 +154,15 @@ class Metric(ABC):
         return cls(**components)
 
 
+def register_metric(metrics: Union[Metric, Type[METRIC], List[Metric], List[Type[METRIC]]]) -> None:
+    global _METRIC_DESERIALIZER_REGISTRY
+    if not isinstance(metrics, list):
+        metrics = [metrics]  # type: ignore
+
+    for metric in metrics:  # type: ignore
+        _METRIC_DESERIALIZER_REGISTRY[metric.get_namespace()] = metric  # type: ignore
+
+
 @dataclass(frozen=True)
 class IntsMetric(Metric):
     max: MaxIntegralComponent
@@ -203,7 +212,7 @@ class IntsMetric(Metric):
         return self.min.value
 
 
-custom_metric(IntsMetric)
+register_metric(IntsMetric)
 
 
 @dataclass(frozen=True)
@@ -423,7 +432,7 @@ class DistributionMetric(Metric):
         )
 
 
-custom_metric(DistributionMetric)
+register_metric(DistributionMetric)
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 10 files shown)


richard-rogers:
I should probably just get rid of the `__post_init__()`

python/whylogs/core/metrics/metrics.py
```diff
@@ -62,9 +62,10 @@ class MetricConfig:
 _METRIC_DESERIALIZER_REGISTRY: Dict[str, Type[METRIC]] = {}  # type: ignore
 
 
-def custom_metric(metric: Type[METRIC], config: MetricConfig = MetricConfig()) -> Type[METRIC]:  # type: ignore
+def custom_metric(metric: Type[METRIC]) -> Type[METRIC]:  # type: ignore
     global _METRIC_DESERIALIZER_REGISTRY
-    _METRIC_DESERIALIZER_REGISTRY[metric.get_namespace(config)] = metric
+    _METRIC_DESERIALIZER_REGISTRY[metric.get_namespace()] = metric
+    print(f"REGISTERED {metric.get_namespace()}")
```
jamie256: You can use a logger to output but lets remove print statements.

python/whylogs/core/metrics/metrics.py
```diff
@@ -62,9 +62,10 @@ class MetricConfig:
 _METRIC_DESERIALIZER_REGISTRY: Dict[str, Type[METRIC]] = {}  # type: ignore
 
 
-def custom_metric(metric: Type[METRIC], config: MetricConfig = MetricConfig()) -> Type[METRIC]:  # type: ignore
+def custom_metric(metric: Type[METRIC]) -> Type[METRIC]:  # type: ignore
```
jamie256, line 65: Should we rename this method (or add a new alias since its doing more than just "custom" metrics)? The purpose is more like registering additional metrics.

jamie256 (dismissed):
Thanks for getting out this fix!

python/whylogs/core/metrics/metrics.py
richard-rogers: oops! meant to delete debug code

python/whylogs/core/metrics/metrics.py
```diff
@@ -424,6 +423,9 @@ def zero(cls, config: Optional[MetricConfig] = None) -> "DistributionMetric":
         )
 
 
+custom_metric(DistributionMetric)
```
jamie256: Can we name it something like `register_metric(DistributionMetric)` or at least a name without 'custom' if this is required for standard metrics?

richard-rogers: It was actually intended to be used as `@custom_metric`. I will add a `register_metric()`

jamie256: I'm fine if the rename is in a follow up, I thought the code would be easier to read with a name that matched the behavior better here.

jamie256 (dismissed):
LGTM!

jamie256 (dismissed):

jamie256 (approved):

PR #1013, opened 2022-12-12
Fix reversed frequent item upper/lower bounds

richard-rogers:
## Description

It should be the case for `FrequentItem`s that lower bound <= estimate <= upper bound. It looks like we got the upper & lower bounds reversed between the pybind11 interface code and the `FrequentItemsMetric`

`items = [FrequentItem(value=x[0], est=x[1], upper=x[2], lower=x[3]) for x in limited_freq_items]`  in `metrics.py` but 
```
    py::tuple t = py::make_tuple(iter->get_item(),
                                 iter->get_estimate(),
                                 iter->get_lower_bound(),
                                 iter->get_upper_bound());
```
in `fi_wrapper.cpp` (in the `datasketches-cpp` repo).

## Changes

-Upper/lower bounds fixed

## Related

### DIFF
--- a/python/tests/core/metrics/test_metrics.py
+++ b/python/tests/core/metrics/test_metrics.py
@@ -153,7 +153,7 @@ def test_merge_two_profiles_mean(lending_club_df: pd.DataFrame) -> None:
 
 def test_frequent_items_handling_int_as_string() -> None:
     df_gamma = pd.DataFrame({"feature1": np.random.gamma(1, 2, 1000).astype(int)})
-    df_rand =  pd.DataFrame({"feature1": np.random.randint(10000, size=9000)})
+    df_rand = pd.DataFrame({"feature1": np.random.randint(10000, size=9000)})
     df = df_gamma.append(df_rand)
 
     res = why.log(df).view().to_pandas()["frequent_items/frequent_strings"]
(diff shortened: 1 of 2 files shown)


richard-rogers:
It might not be practical to use enough data in the unit tests to trigger the bug :(

python/tests/core/metrics/test_metrics.py
```diff
@@ -155,7 +155,9 @@ def test_frequent_items_handling_int_as_string() -> None:
     df = pd.DataFrame({"int": [1, 1, 1]})
```
jamie256: If you give the frequent item sketch something with a mix of frequent collisions and then also some noise and enough entries the estimate bounds will differ, maybe try this to make sure that the lower != upper estimate for a test that can catch this bug:

```suggestion
    df_gamma = pd.DataFrame({"feature1": np.random.gamma(1, 2, 1000).astype(int)})
    df_rand =  pd.DataFrame({"feature1": np.random.randint(10000, size=9000)})
    df = df_gamma.append(df_rand)
```

jamie256 (dismissed):
Suggested better test input data, but the fix looks good, nice catch!

jamie256 (dismissed):

jamie256 (dismissed):

jamie256 (dismissed):

jamie256 (approved):

PR #1014, opened 2022-12-13
performance improvement for low latency row logging

jamie256:
## Description

This change keeps the row based dictionary logging separate from the pandas columnar logging so that there is less overhead per call when logging messages or integrating whylogs into a service rather than batch processing scenarios.

Taken together these changes reduce latency by about an order of magnitude, but numbers are hardware and input dependent, so your mileage may vary. The included new tests are what were used to identify the pandas series preprocessing as a hot spot when logging single row messages with respect to latency and are what were used to compare the performance improvements.

There are further optimizations, tuning, and cleanup but with the significance of this improvement we are releasing this change after our initial round of testing.

These changes are the same as what exist in the pre-release of whylogs 1.1.17-dev0 (modulo version information and test changes).

## Changes

- Add scalar processing path for row logging.
- Added performance benchmark tests for latency.

### DIFF
--- a/python/whylogs/core/column_profile.py
+++ b/python/whylogs/core/column_profile.py
@@ -49,14 +49,21 @@ class ColumnProfile(object):
         self._cache = []
         self.track_column(old_cache)
 
-    def track_column(self, series: Any) -> None:
-        ex_col = PreprocessedColumn.apply(series)
+    def _process_extracted_column(self, extracted_column: PreprocessedColumn) -> None:
         for metric in self._metrics.values():
-            res = metric.columnar_update(ex_col)
+            res = metric.columnar_update(extracted_column)
             self._failure_count += res.failures
             self._success_count += res.successes
         for validator in self._column_validators:
-            validator.columnar_validate(ex_col)
+            validator.columnar_validate(extracted_column)
+
+    def track_column(self, series: Any) -> None:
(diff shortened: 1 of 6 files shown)


python/whylogs/core/column_profile.py
```diff
@@ -49,14 +49,21 @@ def flush(self) -> None:
...
             self._success_count += res.successes
         for validator in self._column_validators:
-            validator.columnar_validate(ex_col)
+            validator.columnar_validate(extracted_column)
+
+    def track_column(self, series: Any) -> None:
+        ex_col = PreprocessedColumn.apply(series)
+        self._process_extracted_column(ex_col)
+
+    def _track_datum(self, value: Any) -> None:
```
murilommen, line 64: why is this private if it's used outside of this module?

murilommen (approved):
not a huge fan of the if+isinstance, but it looks good for now! left a minor comment, out of curiosity. we can break into issues if you need it asap for release

richard-rogers (approved):

python/whylogs/core/column_profile.py
jamie256: its internal to whylogs, so telegraphing that community users should not consider this part of the public API, and we will likely change this a bit as we optimize and do more testing, but wanted to get this out first in the shape we tested it.

PR #1023, opened 2022-12-14
hellinger calculation edge cases

FelipeAdachi:
Changes to deal with some edge cases + testing for hellinger distance calculations.

- Single Values:
     Calculating hellinger for single values led to number of buckets equal to 1, which led to hellinger distance always 0. Added a min_n_buckets parameter to always have at least 2 buckets

 - min==max==0
    Calculating hellinger between single values [0] and [0] led to `math domain error`, due to trying to get the log of 0 in `get_min_intervals`. changed  `_calculate_bins` to deal with end==0

- empty sketches
   Added test to make sure an error is raised when empty sketches

### DIFF
--- a/python/tests/viz/utils/test_drift_calculations.py
+++ b/python/tests/viz/utils/test_drift_calculations.py
@@ -5,6 +5,9 @@ import pandas as pd
 import pytest
 
 import whylogs as why
+from whylogs.core import ColumnProfileView
+from whylogs.core.metrics import DistributionMetric
+from whylogs.core.metrics.metrics import MetricConfig
 from whylogs.viz.utils.drift_calculations import (
     _compute_chi_squared_test_p_value,
     _compute_ks_test_p_value,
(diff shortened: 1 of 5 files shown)


python/tests/viz/utils/test_drift_calculations.py
```diff
@@ -76,6 +79,31 @@ def test_get_hellinger_distance_float(view_columns):
...
+def test_hellinger_on_empty_sketch():
+    distribution_metric = DistributionMetric.zero(MetricConfig())
+    target_col_view = ColumnProfileView({"distribution": distribution_metric})
+    ref_col_view = ColumnProfileView({"distribution": distribution_metric})
+    with pytest.warns(UserWarning, match="Distribution sketch must not be empty."):
+        _get_hellinger_distance(target_view_column=target_col_view, reference_view_column=ref_col_view)
+
+
+def test_hellinger_single_value():
+    params = [([0], [5], 1.0), ([5], [5], 0.0), ([0], [0], 0.0)]
```
richard-rogers: These could be `@pytest.mark.parametrize`

python/whylogs/viz/configs.py
```diff
@@ -8,6 +8,7 @@
 class HistogramConfig:
     max_hist_buckets: int = 30
     hist_avg_number_per_bucket: int = 4
+    min_n_buckets: int = 2
```
richard-rogers, line 11: Should we add validation to ensure these aren't set less than 2?

bernease (approved):
LGTM!

PR #1024, opened 2022-12-15
helper constraints

FelipeAdachi:
### helper constraints:

- ~~is_complete~~ no_missing_values
- ~~is_unique~~
- is_in_range
- is_non_negative

Edit: is_unique has been removed until we figure out multi-metric constraints

### skip_missing

- refactor some helper constraints and metric constraint validation to concile `skip_missing` with `require_column_existence`

### tests

- tests for additional helper constraints and skip_missing logic


python/tests/core/constraints/factories/test_count_metrics.py
```diff
@@ -1,10 +1,21 @@
 from whylogs.core.constraints.factories import (
     count_below_number,
+    is_complete,
```
jamie256: Consider renaming this something more explicit like "no_nulls" or "no_missing_values"

"is_complete" works but its broad enough one might expect that such a check is for completeness in terms of an expected row count or values in a range, or that there isn't any value missing in a set (e.g. the data contains values for all entries in a set).

python/whylogs/core/constraints/factories/distribution_metrics.py
```diff
@@ -60,6 +61,53 @@ def is_smaller(metric) -> bool:
...
+    ----------
+    column_name : str
+        Column the constraint is applied to
+    lower : float
+        lower bound of defined range
+    upper : float
+        upper bound of defined range
+    skip_missing: bool
+        If skip_missing is True, missing distribution metrics will make the check pass.
+        If False, the check will fail on missing metrics
```
jamie256: suggest adding description to this comment:
"such as on an empty dataset"

python/whylogs/core/constraints/factories/frequent_items.py
```diff
@@ -1,8 +1,30 @@
...
+
+    Parameters
+    ----------
+    column_name : str
+        Column the constraint is applied to
+    """
+
+    def unique(metric) -> bool:
+        frequent_strings = metric.to_summary_dict(SummaryConfig())["frequent_strings"]
+        return frequent_strings[0].est == 1
```
jamie256: I think this breaks on large inputs, you probably need to use something like asserting the lower bound estimate is 1.

python/whylogs/core/constraints/factories/frequent_items.py
```diff
@@ -1,8 +1,30 @@
 from whylogs.core.configs import SummaryConfig
+from whylogs.core.relations import Require
 
 from ..metric_constraints import MetricConstraint, MetricsSelector
 
 
+def is_unique(column_name: str) -> MetricConstraint:
```
jamie256: This is based on a probabilistic sketch, we need to name this something like "is_probably_unique" and it would be better perhaps to use cardinality estimate ~= total count.

jamie256 (commented):
I think we need some naming tweaks to make sure these correctly describe what they are checking, but nice to have these helpers!

python/tests/core/constraints/factories/test_count_metrics.py
FelipeAdachi: done!

python/whylogs/core/constraints/factories/distribution_metrics.py
FelipeAdachi: done!

python/whylogs/core/constraints/factories/frequent_items.py
FelipeAdachi: based on previous conversations, I removed this until we get multi-metric constraints figured out

python/tests/core/constraints/factories/conftest.py
```diff
@@ -18,3 +18,14 @@ def nan_builder():
...
+
+@pytest.fixture
+def empty_builder():
+    data = {
+        "a": [],
+    }
+    df = pd.DataFrame(data)
+    profile_view = why.log(pandas=df).view()
+    empty_builder = ConstraintsBuilder(dataset_profile_view=profile_view)
+    return empty_builder
```
murilommen: unless this is used in more than one test module, it should not be on the wider scope

python/tests/core/constraints/factories/test_count_metrics.py
```diff
@@ -1,10 +1,21 @@
...
     null_percentage_below_number,
     null_values_below_number,
 )
 
 
+def test_no_missing_values(builder):
+    builder.add_constraint(no_missing_values(column_name="animal"))
+    builder.add_constraint(no_missing_values(column_name="weight"))
+    constraint = builder.build()
+    assert constraint.validate() is False
```
murilommen: if it passes on having no_missing_values, why is `constraints.validate()` False?

python/tests/core/constraints/factories/test_distribution_metrics.py
```diff
@@ -1,13 +1,65 @@
...
+    report = constraint.generate_constraints_report()
+    assert report[0].name == "a is in range [1.1,3.2]" and report[0].passed == 1
+
+
+def test_is_non_negative():
+    data = {
+        "animal": ["cat", "hawk", "snake", "cat", "mosquito"],
+        "legs": [4, 2, 0, 4, 6],
+        "weight": [4.3, -1.8, 1.3, 4.1, 5.5e-6],
+        "null_column": [np.nan, np.nan, np.nan, np.nan, np.nan],
```
murilommen: it would be better off specifying the length as in:
```python
"null_column": [np.nan] * 5
```

murilommen: in case this changes for some reason

python/whylogs/core/constraints/factories/distribution_metrics.py
```diff
@@ -15,19 +15,14 @@ def greater_than_number(column_name: str, number: Union[float, int], skip_missin
...
-    def is_greater(metric: DistributionMetric) -> bool:
-        if not metric.kll.value.is_empty():
-            return metric.min >= number
-        else:
-            return True if skip_missing else False
-
     constraint = MetricConstraint(
         name=f"{column_name} greater than number {number}",
-        condition=is_greater,
+        condition=lambda x: x.min > number,
```
murilommen: why has this changed back to lambda?

python/whylogs/core/constraints/factories/distribution_metrics.py
```diff
@@ -73,19 +112,14 @@ def mean_between_range(column_name: str, lower: float, upper: float, skip_missin
...
-    def is_mean_between(metric) -> bool:
-        if not metric.kll.value.is_empty():
-            return lower <= metric.avg <= upper
-        else:
-            return True if skip_missing else False
-
     constraint = MetricConstraint(
         name=f"{column_name} mean between {lower} and {upper} (inclusive)",
-        condition=is_mean_between,
+        condition=lambda x: lower <= x.avg <= upper,
```
murilommen: same, why lambda

python/whylogs/core/constraints/factories/distribution_metrics.py
```diff
@@ -103,19 +137,14 @@ def stddev_between_range(column_name: str, lower: float, upper: float, skip_miss
...
-    def is_stddev_between_range(metric):
-        if not metric.kll.value.is_empty():
-            return lower <= metric.stddev <= upper
-        else:
-            return True if skip_missing else False
-
     constraint = MetricConstraint(
         name=f"{column_name} standard deviation between {lower} and {upper} (inclusive)",
-        condition=is_stddev_between_range,
+        condition=lambda x: lower <= x.stddev <= upper,
```
murilommen: same

python/tests/core/constraints/factories/test_count_metrics.py
FelipeAdachi: constraints.validate() is False if any constraint in `constraints` fails. Since `weight` fails on the check, validate() returns false.
actually, `constraint` should be in plural here.

python/whylogs/core/constraints/factories/distribution_metrics.py
FelipeAdachi: previously, the function needed to make additional validation with regards to missing distribution metrics. now this is covered by `require_column_existence`. This led to a very simple one-line function.

Since it was a single line, I decided to make it a lambda. But now I just recalled a conversation where we agreed that we should prefer functions to lambdas. I'm not sure what the reason was, though. Is it because it's easier to debug? any other reason?

FelipeAdachi: The same comment applies to all the other cases below

python/tests/core/constraints/factories/test_distribution_metrics.py
```diff
@@ -1,13 +1,65 @@
...
+    }
+
+    view = why.log(pd.DataFrame(data)).profile().view()
+    builder = ConstraintsBuilder(dataset_profile_view=view)
+
+    builder.add_constraint(is_non_negative(column_name="weight"))
+    builder.add_constraint(is_non_negative(column_name="legs"))
+    builder.add_constraint(is_non_negative(column_name="animal", skip_missing=False))
+    builder.add_constraint(is_non_negative(column_name="null_column", skip_missing=True))
+    with pytest.raises(ValueError) as e:
```
richard-rogers:
```suggestion
    with pytest.raises(ValueError, match="non_existing_column was not found in set of this profile") as e:
```

Then you can drop the assert on line 52

python/tests/core/constraints/factories/test_distribution_metrics.py
```diff
@@ -1,13 +1,65 @@
...
+        builder.add_constraint(is_non_negative(column_name="non_existing_column", skip_missing=True))
+        assert e.value.args[0].startswith("non_existing_column was not found in set of this profile")
+
+    constraint = builder.build()
+    report = constraint.generate_constraints_report()
+    assert report[0].name == "weight is non negative" and report[0].passed == 0 and report[0].failed == 1
```
richard-rogers: parametrize?

python/whylogs/core/constraints/factories/distribution_metrics.py
richard-rogers: I think you could do `Require("min").greater_than(number)`.  Lambdas are often easier to code, but cannot be [de]serialized.

python/whylogs/core/constraints/factories/distribution_metrics.py
```diff
@@ -135,20 +164,18 @@ def quantile_between_range(column_name: str, quantile: float, lower: float, uppe
...
     """
 
     def quantile_in_range(metric):
-        if not metric.kll.value.is_empty():
-            quantile_value = metric.kll.value.get_quantile(quantile)
-            result: bool = lower <= quantile_value <= upper
-            return result
-        else:
-            return True if skip_missing else False
+        quantile_value = metric.kll.value.get_quantile(quantile)
```
richard-rogers, line 171: The `Predicate`s grab values to use in expressions from the metric summary, so I don't think they could handle this case. You're limited to the quantiles hard-coded in the summary.

python/whylogs/core/constraints/factories/distribution_metrics.py
FelipeAdachi: replaced lambda to Predicate in latest commit

python/whylogs/core/constraints/factories/distribution_metrics.py
FelipeAdachi: replaced lambda to Predicate in latest commit

python/whylogs/core/constraints/factories/distribution_metrics.py
FelipeAdachi: replaced lambda to Predicate in latest commit

python/tests/core/constraints/factories/test_distribution_metrics.py
FelipeAdachi: done! thanks!

python/whylogs/core/constraints/factories/distribution_metrics.py
FelipeAdachi: maybe in the future, we could do serializable versions fetching from the hard-coded values in the summary.
Leaving this as is for now.

python/tests/core/constraints/factories/test_distribution_metrics.py
FelipeAdachi: done! thanks!

jamie256 (approved):
LGTM!

PR #1032, opened 2022-12-22
maintenance/0.7.x spark missing columnGroups.head fix

jamie256:
## Description

If you logged performance metrics and no other columns the spark profiling session in v0 would try to split up the set of non-performance metric related columns into batches of 100, and would access the first batch via the Sequence's head (which is empty and throws):

```
java.util.NoSuchElementException: head of empty stream
```

## Changes

- Instead of pulling the head into the batch here, we always process batches of the "other" columns in the second step of the sequence, which already does a map over the sequence and handles the empty case.

tested on Databricks cluster and generating performance metrics with other columns and with only the prediction, target and score columns.

Note: this is a whylogs v0 spark fix, and covers scala/java and whyspark (v0 java + python).

### DIFF
--- a/java/.bumpversion.cfg
+++ b/java/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.2.0-b1
+current_version = 0.2.1-b0
 commit = True
 tag = False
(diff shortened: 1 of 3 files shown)


andyndang (approved):
LGTM

PR #1033, opened 2022-12-22
Multi-column constraint proposal

richard-rogers:
## Description

Refactor to support multi-column constraints

## Changes

Adds `DatasetConstraint` and some other refactoring

A `DatasetConstraint` implements constraints that aren't attached to a specific column. It has a `Callable` that takes a `DatasetProfileView`. It returns a `bool` and a dictionary mapping `column_name/metric_name` to the metrics used in evaluating the constraint. The latter is necessary for generating the report. 

It looks like there's a similar case of a `MetricConstraint` that applies to all metrics of a specified type across all columns. ~I've changed the implementation of this to find the set of metrics the constraint applies to by using the `MetricSelector`, then attaching the constraint to each specific column. This might not be the best approach... maybe converting it to a `DatasetConstraint` would be better.~
`MetricConstraintWrapper` turns such a `MetricConstraint` into the `Callable` for an equivalent `DatasetConstraint`.

I'm implementing a `PrefixCondition` that implements the prefix expression syntax used to serialize the `Predicate` expressions with the addition of a few arithmetic operators. This is probably a reasonable serialization format, but might not be the ultimate user interface for specifying constraints...
It also might want a better name :)

## Related

### DIFF
--- a/python/tests/core/metrics/test_condition_count.py
+++ b/python/tests/core/metrics/test_condition_count.py
@@ -17,7 +17,8 @@ from whylogs.core.metrics.condition_count_metric import Relation as Rel
 from whylogs.core.metrics.condition_count_metric import relation as rel
 from whylogs.core.metrics.metric_components import IntegralComponent
 from whylogs.core.metrics.metrics import OperationResult
-from whylogs.core.predicate_parser import parse_predicate
+from whylogs.core.metrics.unicode_range import UnicodeRangeMetric
+from whylogs.core.predicate_parser import _tokenize, parse_predicate
 from whylogs.core.preprocessing import PreprocessedColumn
 from whylogs.core.relations import Not, Predicate, Require
 from whylogs.core.resolvers import Resolver
@@ -300,13 +301,14 @@ def test_condition_count_in_dataset_profile() -> None:
     assert summary["condition_count/digit"] > 0
 
 
-def _build_profile(data: List[int]) -> DatasetProfile:
+def _build_profile(data: List[int], col_name: str = "col1") -> DatasetProfile:
     """build up a "reference profile" to compare against"""
-    row = {"col1": data}
+    row = {col_name: data}
     frame = pd.DataFrame(data=row)
     prof = DatasetProfile()
     prof.track(pandas=frame)  # track once to discover columns
-    prof.add_metric("col1", DistributionMetric.zero(MetricConfig()))
+    prof.add_metric(col_name, DistributionMetric.zero(MetricConfig()))
+    prof.add_metric(col_name, UnicodeRangeMetric.zero())
     prof.track(pandas=frame)  # track again to populate the metric
     return prof
 
@@ -362,8 +364,39 @@ def test_metric_getter() -> None:
 
 
 @pytest.mark.parametrize(
+    "input,expected",
+    [
+        ('~ x "abc def"', ["~", "x", '"abc def"']),
+        (r'~ x "abc\"def"', ["~", "x", r'"abc"def"']),
+        (r'  ==   x  " \"\" \" "', ["==", "x", r'" "" " "']),
+        (r'  ==   x  "\"\" \""', ["==", "x", r'""" ""']),
+        (r'  ==   x  "\x"', ["==", "x", r'"\x"']),
+        (r'  ==   x  "\"\x"', ["==", "x", r'""\x"']),
+        (r'  ==   x  "\\"', ["==", "x", r'"\\"']),
+        (r'  ==   x  "\\\""', ["==", "x", r'"\\""']),
+        ("::distribution/n", ["::distribution/n"]),
+        (":col1:distribution/n", [":col1:distribution/n"]),
+        ("::multimetric/subname:distribution/n", ["::multimetric/subname:distribution/n"]),
+        (":col1:multimetric/subname:distribution/n", [":col1:multimetric/subname:distribution/n"]),
+        (
+            ":foo bar:distribution/n :col name:multimetric/subname:distribution/n",
+            [":foo bar:distribution/n", ":col name:multimetric/subname:distribution/n"],
+        ),
+        (
+            r":annoying\:name:distribution/n  :annoying\: name:multimetric/subname:distribution/n",
+            [":annoying:name:distribution/n", ":annoying: name:multimetric/subname:distribution/n"],
+        ),
+    ],
+)
+def test_expression_tokenizer(input: str, expected: List[str]) -> None:
+    assert _tokenize(input) == expected
+
+
+@pytest.mark.parametrize(
     "predicate,serialized",
     [
+        (X.matches('abc"def'), r'~ x "abc\"def"'),
+        (X.matches("abc def"), '~ x "abc def"'),
         (X.matches("[a-zA-Z]+"), '~ x "[a-zA-Z]+"'),
         (X.fullmatch("[a-zA-Z]+"), '~= x "[a-zA-Z]+"'),
         (X.equals("42"), '== x "42"'),
@@ -372,6 +405,7 @@ def test_metric_getter() -> None:
         (X.equals(42.0), "== x 42.0"),
         (X.less_than(42), "< x 42"),
         (X.less_or_equals(42), "<= x 42"),
+        (X.greater_than(-42), "> x -42"),
         (X.greater_than(42), "> x 42"),
         (X.greater_or_equals(42), ">= x 42"),
         (X.not_equal(42), "!= x 42"),
@@ -389,6 +423,7 @@ def test_metric_getter() -> None:
         (Require("mean").equals(42.0), "== mean 42.0"),
         (Require("mean").less_than(42), "< mean 42"),
         (Require("mean").less_or_equals(42), "<= mean 42"),
+        (Require("mean").greater_than(-42), "> mean -42"),
         (Require("mean").greater_than(42), "> mean 42"),
         (Require("mean").greater_or_equals(42), ">= mean 42"),
         (Require("mean").not_equal(42), "!= mean 42"),
@@ -406,6 +441,8 @@ def test_serialization(predicate: X, serialized: str) -> None:
 @pytest.mark.parametrize(
     "predicate",
     [
+        (X.matches('abc"def')),
+        (X.matches("abc def")),
         (X.matches("[a-zA-Z]+")),
         (X.fullmatch("[a-zA-Z]+")),
         (X.equals("42")),
@@ -415,6 +452,7 @@ def test_serialization(predicate: X, serialized: str) -> None:
         (X.less_than(42)),
         (X.less_or_equals(42)),
         (X.greater_than(42)),
+        (X.greater_than(-42)),
         (X.greater_or_equals(42)),
         (X.not_equal(42)),
         # (X.is_(even)),
@@ -432,6 +470,7 @@ def test_serialization(predicate: X, serialized: str) -> None:
         (Require("mean").less_than(42)),
         (Require("mean").less_or_equals(42)),
         (Require("mean").greater_than(42)),
+        (Require("mean").greater_than(-42)),
         (Require("mean").greater_or_equals(42)),
         (Require("mean").not_equal(42)),
         # (Require().is_(even)),
@@ -444,3 +483,29 @@ def test_serialization(predicate: X, serialized: str) -> None:
 def test_deserialization(predicate: X) -> None:
     serialized = predicate.serialize()
     assert parse_predicate(serialized).serialize() == serialized
+
+
+@pytest.mark.parametrize(
+    "col_name,expression",
+    [
+        ("col1", "== n 5"),
+        ("col1", "== n :col1:distribution/n"),
+        ("col1", "== n ::distribution/n"),
+        ("annoying name", "== n 5"),
+        ("annoying name", "== n :annoying name:distribution/n"),
+        ("annoying:name", "== n 5"),
+        ("annoying:name", r"== n :annoying\:name:distribution/n"),
+        (":very : annoying: name:", r"== n :\:very \: annoying\: name\::distribution/n"),
+        (":very : annoying: name:", r"== min :\:very \: annoying\: name\::unicode_range/string_length:distribution/n"),
+        (":", r"== n :\::distribution/n"),
+    ],
+)
+def test_metric_getter_deserialization(col_name: str, expression: str) -> None:
+    data = [0, 2, 3, 4, 5]
+    prof = _build_profile(data, col_name)
+    metric = prof.view().get_column(col_name).get_metric("distribution")
+    assert metric
+
+    predicate = parse_predicate(expression, profile=prof, metric=metric)
+    assert predicate(metric)
+    assert predicate.serialize() == expression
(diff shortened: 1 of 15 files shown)


jamie256 (commented):
Can you get the whylogs-ci working?

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -258,6 +329,24 @@ def _generate_metric_report(
...
                 return ReportResult(name=constraint_name, passed=1, failed=0, summary=None)
 
+    def _generate_dataset_report(
+        self,
+        profile_view: DatasetProfileView,
+        constraint: DatasetConstraint,
+        with_summary: bool,
+    ) -> ReportResult:
+        (result, summary) = constraint.validate_profile(profile_view)
+        if not result:
```
richard-rogers: Pull `ReportResult` construction into its own method.

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -21,23 +21,35 @@ class MetricsSelector:
     column_name: Optional[str] = None  # <- TODO: return a collection of columns
     metrics_resolver: Optional[Callable[[DatasetProfileView], List[Metric]]] = None
 
-    def column_profile(self, profile: DatasetProfileView) -> Optional[ColumnProfileView]:
+    def column_profile(self, profile: Optional[DatasetProfileView]) -> Optional[ColumnProfileView]:
         if profile is None:
             return None
         column_profile_view = profile.get_column(self.column_name)
         if column_profile_view is None:
```
richard-rogers:
```suggestion
        if self.column_name and column_profile_view is None:
```

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -171,19 +185,109 @@ def validate_profile(self, dataset_profile: DatasetProfileView) -> Tuple[bool, O
...
+    def _get_metric_summary(self, metrics: Dict[str, Metric]) -> Optional[Dict[str, Any]]:
+        summary: Dict[str, Any] = dict()
+        for path, metric in metrics:
+            metric_summary = pretty_display(metric.namespace, metric.to_summary_dict())
+            for component, value in metric_summary.items():
+                summary[f"{path}/{component}"] = value
+        return summary
+
+    def validate_profile(self, dataset_profile: DatasetProfileView) -> Tuple[bool, Optional[Dict[str, Any]]]:
+        validate_result, metrics = self.condition(dataset_profile, self.require_column_existence)
```
richard-rogers: Move missing column handling here and drop the parameter.

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -201,14 +305,13 @@ def validate(self, profile_view: Optional[DatasetProfileView] = None) -> bool:
...
-        for metric_name in metric_names:
-            metric_constraint = self.dataset_constraints[metric_name]
-            (result, _) = metric_constraint.validate_profile(profile)
+
+        for constraint in self.dataset_constraints:
+            (result, _) = constraint.validate_profile(profile)
             if not result:
-                logger.info(f"{metric_name} failed on dataset")
-                return False
-        return True
```
richard-rogers, line 227: Nope, this was an early return!

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -171,19 +185,109 @@ def validate_profile(self, dataset_profile: DatasetProfileView) -> Tuple[bool, O
         return (validate_result, metric_summary)
 
 
+@dataclass
+class DatasetConstraint:
```
richard-rogers, line 204: Look at pulling out a common base class with `MetricConstraint`

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -171,19 +185,109 @@ def validate_profile(self, dataset_profile: DatasetProfileView) -> Tuple[bool, O
         return (validate_result, metric_summary)
 
 
+@dataclass
```
richard-rogers: Are `MetricConstraint` and `DatasetConstraint` getting value from being `@dataclass` ?

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -323,28 +441,42 @@ def get_metric_selectors(self) -> List[MetricsSelector]:
...
-                f"{column_name} was not found in set of this profile, the list of columns are: {self._dataset_profile_view.get_columns()}"
-            )
+    def add_constraint(self, constraint: Union[MetricConstraint, DatasetConstraint], ignore_missing: bool = False) -> "ConstraintsBuilder":
+        print("=" * 80)
+        if isinstance(constraint, MetricConstraint):
+            # check that the column exists and we have a column profile view
+            column_name = constraint.metric_selector.column_name
+            column_profile_view = self._dataset_profile_view.get_column(column_name)
+            if column_profile_view is None and not ignore_missing and constraint.metric_selector.metrics_resolver is None:
+                raise ValueError(
```
richard-rogers, line 622: can `column_name` be `None`?

python/whylogs/core/constraints/metric_constraints.py
richard-rogers: The ctors are nice. The `Callable`s will be the hard part of [de]serialization though, so not super-helpful there.

python/whylogs/core/constraints/metric_constraints.py
richard-rogers: The `name` and `require_column_existence` are common. `validate_profile()` could be an abstract method. I think I vote to hold off on extracting a base class until we need another would-be subclass...

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -171,20 +172,145 @@ def validate_profile(self, dataset_profile: DatasetProfileView) -> Tuple[bool, O
...
+
+    if with_summary:
+        return ReportResult(name=constraint_name, passed=1, failed=0, summary=metric_summary)
+    return ReportResult(name=constraint_name, passed=1, failed=0, summary=None)
+
+
 class Constraints:
     column_constraints: Dict[str, Dict[str, MetricConstraint]]
-    dataset_constraints: Dict[str, MetricConstraint]
+    dataset_constraints: List[DatasetConstraint]
```
richard-rogers, line 468: Potential compatibility failure. `ConstraintsBuilder::add_constraint()` should maintain compatible semantics, so hopefully nobody is messing with `Constraints::dataset_constraints` directly.

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -282,17 +405,14 @@ def generate_constraints_report(
 
                 results.append(metric_report)
 
-        metric_names = self.dataset_constraints.keys()
```
richard-rogers, line 301: Line 267/390 is a bug. If `column_constraints` is empty but `dataset_constraints` is not (even in the old version where `dataset_constraint` contained `MetricConstraint`s), the reports for the dataset constraints would not be generated.

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -201,12 +327,11 @@ def validate(self, profile_view: Optional[DatasetProfileView] = None) -> bool:
                 if not result:
```
richard-rogers, line 491: Line 193/319 is a bug. `column_constraints` can be empty while `dataset_constraints` is not.

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -171,26 +172,174 @@ def validate_profile(self, dataset_profile: DatasetProfileView) -> Tuple[bool, O
...
+        self._expression = expression.split()  # TODO: Bug! splits spaces in string literals
+        self._profile = None
+        self._metric_map: Dict[str, Metric] = dict()
+
+    def _interpret(self, i: int) -> Tuple[Any, int]:
+        token = self._expression[i]
+        # ...
+        if token.startswith(":"):
+            # TODO: lexical validation
+            column_name, path = token[1:].split(":", 1)
```
richard-rogers: Can column names contain `:` ?

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -171,26 +172,174 @@ def validate_profile(self, dataset_profile: DatasetProfileView) -> Tuple[bool, O
...
+class PrefixCondition:
+    """
+    Interpret expressions in the form used to serialize Predicate expressions.
+    This is probably a reasonable serialization format for DatasetConstraint
+    conditions, but might not be the best user interface for creating conditions
+    in client code.
+    """
+
+    def __init__(self, expression: str) -> None:
+        self._expression = expression.split()  # TODO: Bug! splits spaces in string literals
```
richard-rogers: This (and the `Predicate` parser) needs to respect quoted string literals. It will probably need [un]escaping too.

python/whylogs/core/constraints/metric_constraints.py
jamie256: the names aren't restricted by whylogs

python/whylogs/core/constraints/metric_constraints.py
jamie256: I think that's fair.

jamie256 (dismissed):
I think we can try this out, exciting!

We should follow up with some examples and documentation in this area.

python/whylogs/core/predicate_parser.py
```diff
@@ -4,37 +4,64 @@
...
+
+def _unescape_token(input: str) -> str:
+    if input[0] == '"':
+        return unescape_quote(input)
+    if input[0] == ":":
+        return unescape_colon(input)
+    return input
+
+
+_TOKEN_RE = re.compile(r'(?::(?:[^:]|\\:)*:[_a-zA-Z0-9]+/[_a-zA-Z0-9:/]+)|(?:[^ ":][^ ]*)|"(?:[^\\"]|\\[^"]|\\")*"')
```
richard-rogers, line 30: Probably wants a comment :)

python/whylogs/core/predicate_parser.py
```diff
@@ -4,37 +4,64 @@
...
+
+def _tokenize(input: str) -> List[str]:
+    return [_unescape_token(t) for t in _TOKEN_RE.findall(input)]
 
 
 def _get_component(token: List[str], i: int) -> Tuple[str, int]:
     return token[i], i + 1  # either dummy variable x or metric component name in metric::to_summary_dict()
 
 
+_METRIC_REF = re.compile(r"::[_a-zA-Z0-9]+/(?:[_a-zA-Z0-9]+:[_a-zA-Z0-9]+/)?[_a-zA-Z0-9]+")
```
richard-rogers, line 45: Also comment-worthy

python/tests/core/metrics/test_condition_count.py
```diff
@@ -361,9 +363,40 @@ def test_metric_getter() -> None:
...
+        (":col1:distribution/n", [":col1:distribution/n"]),
+        ("::multimetric/subname:distribution/n", ["::multimetric/subname:distribution/n"]),
+        (":col1:multimetric/subname:distribution/n", [":col1:multimetric/subname:distribution/n"]),
+        (
+            ":foo bar:distribution/n :col name:multimetric/subname:distribution/n",
+            [":foo bar:distribution/n", ":col name:multimetric/subname:distribution/n"],
+        ),
+        (
+            r":annoying\:name:distribution/n  :annoying\: name:multimetric/subname:distribution/n",
+            [":annoying:name:distribution/n", ":annoying: name:multimetric/subname:distribution/n"],
```
richard-rogers, line 389: Add test with column name `\:`

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -171,26 +174,295 @@ def validate_profile(self, dataset_profile: DatasetProfileView) -> Tuple[bool, O
...
+    def __init__(self, constraint: MetricConstraint) -> None:
+        self.constraint = constraint
+
+    def __call__(self, profile: DatasetProfileView) -> Tuple[bool, Dict[str, Metric]]:
+        valid, _ = self.constraint.validate_profile(profile)
+        metrics = self.constraint.metric_selector.apply(profile)  # Metrics used during evaluation
+        columns = profile.get_columns()
+        metric_map: Dict[str, Metric] = dict()
+        # Find the columns the referenced Metrics live in
+        for metric in metrics:
```
richard-rogers, line 258: Consider interchanging these loops

python/whylogs/core/constraints/metric_constraints.py
richard-rogers: Probably should be frozen

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -171,26 +174,295 @@ def validate_profile(self, dataset_profile: DatasetProfileView) -> Tuple[bool, O
...
+        if token.startswith('"'):
+            return token[1:-1], i + 1
+
+        if bool(_INT_RE.fullmatch(token)):
+            return int(token), i + 1
+
+        if bool(_FLOAT_RE.fullmatch(token)):
+            return float(token), i + 1
+
+        raise ValueError("Unparsable expression: '{self._expression}' at token {i}: '{token}'")
```
FelipeAdachi: missing `f"...`?

richard-rogers: nice catch!

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -171,26 +174,295 @@ def validate_profile(self, dataset_profile: DatasetProfileView) -> Tuple[bool, O
...
+        """
+        relational operators: ~ ~= == < <= > >= !=
+        boolean operators:    and or not
+        arithmetic operators: + - * ** / // %
+
+        literals: 42 3.14 "blah"
+
+        metric component references: :column_name:metric_namespace/metric_component
+
+        returns a map of :column_name:metric_namespace -> Metric
```
FelipeAdachi: returns a bool indicating whether the expression passed and a map of....

FelipeAdachi (commented):
I'm wondering if the name `DatasetConstraint` might cause confusion with the upcoming `Dataset Level Metrics`, considering they won't necessarily be related.

Will Dataset Metric need its own class of Constraint? If so, we should probably reserve the name DatasetConstraint for that. If the current implementation of DatasetConstraint will also work for Dataset Metrics, maybe it's okay?

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -171,26 +174,297 @@ def validate_profile(self, dataset_profile: DatasetProfileView) -> Tuple[bool, O
...
+        return valid, metric_map
+
+
+_INT_RE = re.compile(r"[-+]?\d+")
+_FLOAT_RE = re.compile(r"[-+]?(\d+(\.\d*)?|\.\d+)([eE][-+]?\d+)?")
+
+
+class PrefixCondition:
+    """
+    Interpret expressions in the form used to serialize Predicate expressions.
```
FelipeAdachi, line 272: Maybe we could add the information that it uses prefix (polish) notation, and add the link:

https://en.wikipedia.org/wiki/Polish_notation

richard-rogers: I'm not sure it's good as the user interface... I'm thinking of it more as a serialization format. I tried a version using operator overloading, which sort of worked but had some infelicities. I think a fluent API would be awkward with the arithmetic operators. @labzhinova suggested an expression wizard UI in the front end. That would be nice, but doesn't answer how to write them in whylogs. Still pondering :)

jamie256 (dismissed):
Still good.

jamie256 (approved):

PR #1035, opened 2022-12-22
Ensure timezone aware datetime only

andyndang:
## Description

Ensure that we only accept tz-aware datetime object across the system

## Changes

- Add checks to tz aware datetime


jamie256 (approved):
This caught a few more bugs here too in examples and v0 converters, thanks!

PR #1037, opened 2022-12-23
Support computing segmented performance metrics in top level API

jamie256:
## Description

The `log_regression_metrics` and `log_classification_metrics` methods accept a DatasetSchema but didn't support computing the performance metrics on segmentation schemes.

This change supports segmented performance metrics from these top level methods, like this:
```
results = log_classification_metrics(
        df,
        target_column=target_column,
        prediction_column=prediction_column,
        schema=DatasetSchema(segments=segment_on_column("feature_1"))
    )

```
## Changes

- minor refactor to reduce duplication between different perf metrics.
- removed error on segments defined in perf metrics
- implemented tests and segmented perf metrics calculation.

### DIFF
--- a/python/poetry.lock
+++ b/python/poetry.lock
@@ -3208,7 +3208,7 @@ lazy-object-proxy = [
     {file = "lazy_object_proxy-1.7.1-pp37.pp38-none-any.whl", hash = "sha256:d66906d5785da8e0be7360912e99c9188b70f52c422f9fc18223347235691a84"},
 ]
 livereload = [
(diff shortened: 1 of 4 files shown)


bernease (approved):
LGTM based on my understanding of segments! I think we're a little too dependent on pandas here, but that can be a fix for another day. Perhaps even packaged with the multiclass metrics API (which intends to introduce a non-pandas option and will need to be extended for segments)..

jamie256:
> LGTM based on my understanding of segments! I think we're a little too dependent on pandas here, but that can be a fix for another day. Perhaps even packaged with the multiclass metrics API (which intends to introduce a non-pandas option and will need to be extended for segments)..

Thanks Bernease! I agree we should expand this beyond the pandas but lets do those things in follow up changes.

PR #1043, opened 2023-01-05
Fix UnicodeRangeMetric submetric updates

richard-rogers:
## Description

This fixes how `UnicodeRangeMetric` passes data to its submetrics.  Some submetrics required data computed in `PreprocessedColumn.apply()`, so they were upset with the manually constructed preprocessed columns.

Also corrects the upper/lower case Unicode range definitions.

## Changes

- Pass data to submetrics with `PreprocessedColumn::apply()`
- Reverses the upper/lower case Unicode range definitions


jamie256 (approved):
LGTM!

PR #1045, opened 2023-01-06
Experimental embeddings metric

richard-rogers:
## Description

Embedding metric

## Changes

- Adds an `EmbeddingMetric`


bernease (commented):
Looks good so far! Nice that PreprocessedColumn will have a 2D array, I was worried things would be much more complicated.

There is a component that is inherently multicolumn and much more complex that uses prediction label as well as reference label here (plot_closest_references_by_label() in my notebook), but I totally understand omitting it for first version of this.

Only a few little typos which I'm omitting for now.

python/whylogs/experimental/core/metrics/embedding_metric.py
```diff
@@ -0,0 +1,128 @@
...
+    @property
+    def namespace(self) -> str:
+        return "embedding"
+
+    def columnar_update(self, data: PreprocessedColumn) -> OperationResult:
+        X = data.list.objs  # TODO: throw if not 2D
+        if not X:
+            return OperationResult.ok(0)
+
+        X_ref_dists = self.deistance_fn.value(X, self.references.value)
```
bernease: Typo: self.d**e**istance_fn

python/whylogs/experimental/core/metrics/embedding_metric.py
```diff
@@ -0,0 +1,128 @@
...
+    labels: Optional[List[str]] = None
+    distance_fn: DistanceFunction = cosine
+
+    def __post_init__(self) -> None:
+        if len(self.references.shape()) != 2:
+            raise ValueError("Embedding reference matrix must be 2 dimensional")
+
+        if self.labels:
+            if len(self.labels) == self.references.shape()[1]:
+                raise ValueError(f"Number of labes ({len(self.labels}) must match number of reference vectors ({self.references.shape()[1]")
```
bernease: Typo: labe[l]s

python/whylogs/experimental/core/metrics/embedding_metric.py
```diff
@@ -0,0 +1,129 @@
...
+    @property
+    def namespace(self) -> str:
+        return "embedding"
+
+    def _update_submetrics(self, submetric: str, data: PreprocessedColumn) -> None:
+        for key in self.submetrics[submetric].keys():
+            self.submetrics[submetric][key].columnar_update(data)
+
+    def columnar_update(self, data: PreprocessedColumn) -> OperationResult:
+        X = data.list.objs  # TODO: throw if not 2D
```
jamie256: TODO: Preprocess into a collection for vectors on the NumpyView, assuming this is placeholder or is there a reason this is expected to process off the objs list?

python/whylogs/experimental/core/metrics/embedding_metric.py
```diff
@@ -0,0 +1,129 @@
+from dataclasses import dataclass
+from enum import Enum
+from io import BytesIO
+from typing import List, Optional
+
+import numpy as np
+from sklearn.metrics.pairwise import cosine_distances, euclidean_distances
```
jamie256, line 8: Both these distance functions are well known and not a lot of code let's not require sklearn for these.

Suggest we create numpy based implementation, e.g.:
```python
def cosine_distance(v1: np.ndarray, v2: np.ndarray) -> float:
    # TODO: add checks that v1 and v2 are valid for calculations (shapes aligned).
    return 1.0 - np.dot(v1, v2)/(np.linalg.norm(v1)*np.linalg.norm(v2))
```

python/whylogs/experimental/core/metrics/embedding_metric.py
```diff
@@ -0,0 +1,129 @@
...
+    references: MatrixComponent
+    labels: List[str]
+    distance_fn: DistanceFunction
+
+    def __post_init__(self):
+        submetrics = {
+            f"{label}_distance": {
+                "distribution": StandardMetric.distribution.zero(),
+                "counts": StandardMetric.counts.zero(),
+                "types": StandardMetric.types.zero(),
```
jamie256, line 104: what's your motivation for the 'types' metric here, wouldn't there already be a top level types?

python/whylogs/experimental/core/metrics/embedding_metric.py
```diff
@@ -0,0 +1,129 @@
...
+            }
+            for label in self.labels
+        }
+        submetrics.update(
+            {
+                "closest": {
+                    "frequent_items": StandardMetric.frequent_items.zero(),
```
jamie256, line 111: Can you add a comment to how you will use "closest" and the "labelN_distance", I think what you are doing is providing a way to monitor the distribution of the closest vectors, and then separately we have distributions of the distances for each reference?

An short example notebook would also help provide more context.

python/whylogs/experimental/core/metrics/embedding_metric.py
```diff
@@ -0,0 +1,129 @@
...
+        for i in range(X_ref_dists.shape[1]):
+            closest.append(self.lables[X_ref_closest[i]])
+            self._update_submetrics(f"{i}_distance", PreprocessedColumn.apply(X_ref_dists[i]))
+
+        self._update_submetrics("closest", PreprocessedColumn.apply(closest))
+
+        return OperationResult.ok(1)
+
+    @classmethod
+    def zero(cls, cfg: Optional[EmbeddingConfig] = None) -> "EmbeddingMetric":
```
jamie256, line 215: The zero is a little weird without reference vector, can you merge a zero EmbeddingMetric with a properly initialized one? If merging with zero doesn't work, might be least surprising behavior to raise an error.

jamie256 (commented):
Nice start, this approach looks interesting to try out!

* Suggest we use namespace: `experimental/metrics/embedding_metric.py`
* You might need to wrap this in an extra e.g. whylogs[embeddings] if this functionality uses non-core dependent packages.
* Needs an example notebook so we can try it out (toy data is fine).

python/whylogs/experimental/core/metrics/embedding_metric.py
richard-rogers: WhyLabs monitor ability

python/whylogs/experimental/core/metrics/embedding_metric.py
richard-rogers: Let's check with @bernease, but I would guess that merging metrics with differently shaped references or different labels should fail. A `zero()`ed metric is fine as long as it was passed a compatible `EmbeddingConfig`. I can hack `merge()` to return the other metric if one was default configured.

python/whylogs/experimental/core/metrics/embedding_metric.py
richard-rogers: I got used to avoiding it from the NLP work :)  It should be fine here though.

richard-rogers:
> * Suggest we use namespace: `experimental/metrics/embedding_metric.py`

My thought was to have the tree under `experimental` match the tree under `whylogs` so when experimental code graduates to production you can just delete the `.experimental` from the imports.

python/whylogs/experimental/core/metrics/embedding_metric.py
richard-rogers: Actually, this doesn't seem to be working with the `_process_scalar_value()` plumbing :(

python/whylogs/experimental/core/metrics/embedding_metric.py
bernease: No thoughts on zero metric, but yes, merging metrics with different references should fail.

python/whylogs/experimental/core/metrics/embedding_metric.py
bernease: @jamie256, happy to walk you through the design notebook I created that shows this to hear your thoughts first.

Yes, these distributions are meant to be monitored using the platform. We may want to discuss whether we want to add functions to extend that to gain a similar level of value out of whylogs-only users.

python/whylogs/experimental/core/metrics/embedding_metric.py
jamie256: You might want to allow merging with zero to take the other metric, but sounds good, and raising an error when different references or shape makes sense. Thanks for explaining.

python/whylogs/experimental/core/metrics/embedding_metric.py
jamie256: “Scalar” is a little weird for naming the processing a vector, more I meant I would expect the preprocessing would put the vectors into a NumpyView, it can be a TODO to follow up if the list objects works, but it’s also an “everything” else bucket so I wouldn’t throw if not a vector but skip processing it

richard-rogers: I made a change in `dataset_profile.py` to get `ndarray` into the Numpy view. Let me know what you think of it.

python/whylogs/experimental/core/metrics/embedding_metric.py
```diff
@@ -0,0 +1,204 @@
...
+            references = MatrixComponent.from_protobuf(msg.metric_components["references"])
+            msg.metric_components.pop("references")
+            serialize_references = True
+        else:
+            references = np.zeros((1, 1))
+            serialize_references = False
+
+        submetrics = EmbeddingMetric.submetrics_from_protobuf(msg)
+
+        # TODO: this doesn't guarantee order :(  Fix it, or don't [de]serialize
```
richard-rogers: Do we need to deserialize the labels in reference vector order, i.e., maintain the relationship between labels and reference matrix rows? If so, we'll probably need to serialize them explicitly rather than recovering them from the submetric names. Deserialized metrics aren't updatable, but they are supposed to be mergeable. I'd have to remove the? same-label check in the merge code if we can't guarantee the label ordering. Do we need/want the labels for visualization

richard-rogers: We could sort the labels and then apply the same permutation to the reference matrix rows

bernease: We will need to maintain relationship for visualization and interpretation purposes. Sorted labels should be fine. This seems similar to what happened in model performance metric serialization / merging, so you may check there.

python/whylogs/experimental/core/metrics/embedding_metric.py
```diff
@@ -0,0 +1,208 @@
...
+                logger.warning("Attempt to merge with unconfigured EmbeddingMetric; ignored")
+                return self
+            if self.references.value.shape == (1, 1):
+                logger.warning("Attempt to merge with unconfigured EmbeddingMetric; ignored")
+                return other
+            raise ValueError("Attempt to merge incompatible EbeddingMetrics")
+
+        if (
+            self.labels != other.labels
+            or self.distance_fn != other.distance_fn
```
richard-rogers: Drop the distance function check or [de]serialize the distance function.

python/whylogs/experimental/core/metrics/embedding_metric.py
```diff
@@ -0,0 +1,216 @@
...
+        self.labels = sorted(self.labels)
+        self.references = MatrixComponent(self.references.value[label_indices, :])
+
+    @property
+    def namespace(self) -> str:
+        return "embedding"
+
+    def merge(self, other: "EmbeddingMetric") -> "EmbeddingMetric":
+        if self.references.value.shape != other.references.value.shape:
+            if other.references.value.shape == (1, 1):
```
richard-rogers, line 129: This still doesn't work for merging with deserialized metrics that didn't include their reference matrices. I'll fix it later.

python/whylogs/experimental/core/metrics/embedding_metric.py
jamie256: its still a surprising callstack, but maybe thats more of an internal detail. Its something we can follow up on, so we don't have to block on this issue, and better that it gets attached to the NumpyView now.

python/whylogs/experimental/core/metrics/embedding_metric.py
jamie256: ok, you moved the sklearn into an optional extra, that's acceptable.

python/whylogs/experimental/core/metrics/embedding_metric.py
jamie256: Fair enough. Can you add a comment with TODO or file an issue?

bernease (approved):
LGTM! I want to clean up the notebook quite a bit but have a pressing deadline by EOD. Does it make sense for you to merge this and I can go in behind and improve it?

For now, it'd make sense to just drop the extra stuff after uploading the profiles to WhyLabs.

PR #1046, opened 2023-01-10
drift report v1

FelipeAdachi:
## Description

Changes summary_drift_report to enable user customization os algorithm and algorithm's internal parameters.
Also enables customization for standalone drift calculation.


murilommen (commented):
IMO [column_drift_algorithms.py](https://github.com/whylabs/whylogs/pull/1046/files#diff-29780ae472cd6d735a7c705c3e14fb8cb884443e2924395b4824db526ef1cbd5) could be split into different modules, one for each test. Also take into account how *slow* it can get to have **both** KS and ChiSquared calculated to only on the return statement decide which one to use.

murilommen:
Also thinking that `generate_summaries_with_drift_score` function has got too many conditions inside itself, which can be hard to test/debug if it misbihave for some tricky reason. Hard to determine which one of the conditions are/aren't satisfied, and increase chances of silently failing

jamie256 (approved):
LGTM!

We walked through the code together while github was failing to show the diffs, there are some minor things similar to Murilo's comments I think are worth following up with some refactor and cleanup, but the examples and new functionality look great!

PR #1056, opened 2023-01-18
Support tagging columns as outputs

jamie256:
## Description

After uploading whylogs profiles or performance metrics, we often want to have the columns classified as input or outputs for monitoring and analysis. This change adds a method that lets practioners tag columns from previously uploaded profiles as being outputs. 

## Changes

- Add new method to WhyLabsWriter to tag columns
- convenience method for tagging output columns

## Example

If you previously ran an example notebook such as: https://github.com/whylabs/whylogs/blob/mainline/python/examples/integrations/writers/Writing_Classification_Performance_Metrics_to_WhyLabs.ipynb

Then you have profiles already uploaded to Whylabs, then you can simply run the following to set the "rating" field as an output.

```python
from whylogs.api.writer.whylabs import WhyLabsWriter
whylabs_writer = WhyLabsWriter() 
whylabs_writer.tag_output_columns(["rating", "output_discount"])
```

In this example the "output_discount" column is already treated as an output and so the schema is not updated. The returned status for this method call above would be:

```
(True,
 "{'rating': 'rating updated from input to output', 'output_discount': 'column output_discount was already classified output.'}")
```

### DIFF
--- a/python/poetry.lock
+++ b/python/poetry.lock
@@ -3335,7 +3335,7 @@ lazy-object-proxy = [
     {file = "lazy_object_proxy-1.9.0-cp39-cp39-win_amd64.whl", hash = "sha256:db1c1722726f47e10e0b5fdbf15ac3b8adb58c091d12b3ab713965795036985f"},
 ]
 livereload = [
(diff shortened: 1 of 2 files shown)


bernease (approved):
Looks great! No suggestions here.

PR #1062, opened 2023-01-23
Support automatic fallback to non-delimited read

### LINKED ISSUE #1058: Attempt fallback to reading profiles as v0 format if magic header is missing
## Description

If you try to read a profile file that was created with whylogs v0 (Java or Python) using whylogs v1, we check for the whylogs header and fail the read.

Also its possible to write non-delimited protobuf files, which also hit an error when read back with the delimited read.

## Suggestions

For scenarios where there are whylogs profiles created from v0 versions of whylogs (either Java or Python), whylogs can attempt a fallback read assuming v0 format profiles, and attempting non-delimited read, and convert them into v1 DatasetProfileViews if successful. Suggest we log a warning in this case, but we can make it work:

```python
from whylogs.core import DatasetProfileView
from whylogs.migration.converters import read_v0_to_view

view: DatasetProfileView = read_v0_to_view(profile_file_path_v0)

```

## Related

jamie256:
## Description
Closes #1058 by catching the DecodeError and attempting a non-delimited file read that works on v0 profile files serialized without a delimited write. While an edge case this should help make our converter utilities handle these cases.

## Changes

- Add try around delimited read in protobuf_utils
- Handle DecodeError, log diagnostic information as warning as this should be a rare edge case
- Attempt to read from the stream at position 0, without delimited header check
- No catching of exceptions in the fallback attempt

### DIFF
--- a/python/whylogs/core/utils/protobuf_utils.py
+++ b/python/whylogs/core/utils/protobuf_utils.py
@@ -2,11 +2,13 @@
 A read/write library for length-delimited protobuf messages.
 
 Based on: https://github.com/soulmachine/delimited-protobuf/blob/main/delimited_protobuf.py"""
+from logging import getLogger
 from typing import IO, Type, TypeVar
 
-from google.protobuf.message import Message
+from google.protobuf.message import DecodeError, Message
 
 T = TypeVar("T", bound=Message)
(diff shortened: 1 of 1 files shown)


FelipeAdachi (approved):
LGTM!

PR #1064, opened 2023-01-25
Error handling converting frequent items from v0

jamie256:
Follow up changes for converting v0 profiles:

* error handling around frequent items sketch, allow dropping these on error.
* fix metadata handling of converted profiles to handle segments
* Added conversion of model performance metrics from v0 to v1
* Added check for RuntimeWarning as some notebook environments won't raise the DecodeError, but we still should fallback to retrying to read such files. see https://docs.python.org/3/library/warnings.html#warnings.catch_warnings

### DIFF
--- a/python/whylogs/core/utils/protobuf_utils.py
+++ b/python/whylogs/core/utils/protobuf_utils.py
@@ -9,7 +9,6 @@ from google.protobuf.message import DecodeError, Message
 
 from whylogs.core.errors import DeserializationError
 
-
 T = TypeVar("T", bound=Message)
 logger = getLogger(__name__)
 
--- a/python/whylogs/core/utils/protobuf_utils.py
+++ b/python/whylogs/core/utils/protobuf_utils.py
@@ -7,6 +7,8 @@ from typing import IO, Type, TypeVar
 
 from google.protobuf.message import DecodeError, Message
 
+from whylogs.core.errors import DeserializationError
+
 T = TypeVar("T", bound=Message)
 logger = getLogger(__name__)
 
@@ -45,7 +47,12 @@ def read_delimited_protobuf(stream: IO[bytes], proto_class_name: Type[T], offset
         )
         stream.seek(0)
         buf = stream.read()
-        msg.ParseFromString(buf)
+        try:
+            msg.ParseFromString(buf)
+        except Exception as e:
+            logger.error(f"{e}: Occured during attempted fallback read of {proto_class_name}.")
+            raise DeserializationError(f"Failed fallback attempt to read {proto_class_name}:{e}")
+
     return msg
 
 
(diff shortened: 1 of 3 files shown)


python/whylogs/core/utils/protobuf_utils.py
```diff
@@ -36,16 +39,45 @@ def read_delimited_protobuf(stream: IO[bytes], proto_class_name: Type[T], offset
...
+        except Exception as decode_error:
+            logger.warning(
+                f"{decode_error}: when reading delimited protobuf file of size "
+                f"{size} bytes as {proto_class_name}, falling back to non-delimited read."
+            )
+            delimited_read_failure = True
+
+        if not delimited_read_failure and len(warning_messages) > 0:
+            for w in warning_messages:
+                if issubclass(w.category, RuntimeWarning):
```
FelipeAdachi, line 58: RuntimeWarning looks pretty general.

Is it possible to reach a case where it's a delimited file and we parsed it correctly, but another type of RuntimeWarning was raised, and then we incorrectly assume we should fall back to not-delimited parsing?

python/whylogs/migration/converters.py
```diff
@@ -106,51 +107,80 @@ def _generate_segment_tags_metadata(
     return segment_message_tags, segment_tags, segment_metadata
 
 
-def read_v0_to_view(path: str) -> DatasetProfileView:
+def read_v0_to_view(path: str, allow_partial: bool = True) -> DatasetProfileView:
```
FelipeAdachi: allow partial defaults to true here, but in `v0_to_v1_view` it defaults to false.
Should we set both to the same value? I'd vote for the default to be False, and in the example we could set it to True and explain.

FelipeAdachi (commented):
~~maybe we should add tests for converters in a followup?~~
Sorry, we already have those. maybe expand, then.

python/whylogs/migration/converters.py
jamie256: I was on the fence there, but sure the read_v0_to_view can make callers pass in allow_partial = True and raise an exception by default on errors, thanks for the comment, updated.

python/whylogs/core/utils/protobuf_utils.py
jamie256: Good point, tightened this to the specific DecodeError message string we see in testing.

jamie256:
> ~maybe we should add tests for converters in a followup?~ Sorry, we already have those. maybe expand, then.

Agreed, lets add more in a follow up, I'm using a few test files from whylogs v0, some of these issues don't repro in our CI or when run locally: https://github.com/whylabs/whylogs/issues/1067

FelipeAdachi (approved):
LGTM!

PR #1070, opened 2023-01-31
Add limit to frequent item string length

richard-rogers:
## Description

Adds `max_frequent_item_size` field to `MetricConfig` to limit the maximum length of frequent item strings.

## Changes

- Adds `max_frequent_item_size` to `MetricConfig` with default value of 128


jamie256 (approved):
LGTM!

PR #1079, opened 2023-02-06
pyspark segments support

### LINKED ISSUE #902: ResultSet mergeability

jamie256:
## Description

Added new method to support profiling segmented results. As this is still in experimental namespace, this change adds a parallel method to the previous `collect_dataset_profile_view` with `collect_segmented_results`: 

```python
from whylogs.api.pyspark.experimental import collect_dataset_profile_view, collect_segmented_results
from whylogs.core.segmentation_partition import segment_on_column
from whylogs.core import DatasetSchema

segmented_results = collect_segmented_results(input_df=df, schema=DatasetSchema(segments=segment_on_column("column_A")))
```

## Changes

- Updates to SegmentedResultSet: supports merging and result set scoped metrics
- Added merge for metadata and metrics
- Added new pyspark grouped profiling using segment columns.
- pyspark schema udf is keyed on column name, and segment key

## Related

### DIFF
--- a/python/tests/api/pyspark/experimental/test_profiler_function.py
+++ b/python/tests/api/pyspark/experimental/test_profiler_function.py
@@ -37,8 +37,6 @@ def _assert_segmented_result_sets_are_equal(results_a: SegmentedResultSet, resul
         if isinstance(segmented_view_b, DatasetProfile):
             segmented_view_b = segmented_view_b.view()
         assert isinstance(segmented_view_b, DatasetProfileView)
-        if segmented_view_a is None and segmented_view_b is None:
-            return
         if segmented_view_a is None or segmented_view_b is None:
             assert segmented_view_a == segmented_view_b
 
--- a/python/tests/api/pyspark/experimental/test_profiler_function.py
+++ b/python/tests/api/pyspark/experimental/test_profiler_function.py
@@ -9,9 +9,9 @@ from whylogs.api.logger.result_set import SegmentedResultSet
 from whylogs.api.pyspark.experimental import (
     collect_column_profile_views,
     collect_dataset_profile_view,
+    collect_segmented_results,
     column_profile_bytes_aggregator,
     whylogs_pandas_map_profiler,
-    collect_segmented_results,
 )
 from whylogs.core import ColumnProfileView, DatasetProfile, DatasetSchema, Resolver
 from whylogs.core.metrics import StandardMetric
@@ -21,6 +21,7 @@ from whylogs.core.view.dataset_profile_view import DatasetProfileView
 from whylogs.core.view.segmented_dataset_profile_view import SegmentedDatasetProfileView
 
 TEST_LOGGER = getLogger(__name__)
+
 
 def _assert_segmented_result_sets_are_equal(results_a: SegmentedResultSet, results_b: SegmentedResultSet) -> None:
     assert type(results_a) == type(results_b)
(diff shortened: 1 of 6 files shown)


python/tests/api/pyspark/experimental/test_profiler_function.py
```diff
@@ -4,25 +4,72 @@
...
+        assert isinstance(profile_a, SegmentedDatasetProfileView)
+        segment = profile_a.segment
+        segmented_view_a = profile_a.profile_view
+        segmented_view_b = results_b.profile(segment)
+        if isinstance(segmented_view_b, DatasetProfile):
+            segmented_view_b = segmented_view_b.view()
+        assert isinstance(segmented_view_b, DatasetProfileView)
+        if segmented_view_a is None and segmented_view_b is None:
+            return
+        if segmented_view_a is None or segmented_view_b is None:
```
richard-rogers, line 40: This seems odd. Seems like it will always either early return or assert?

jamie256: The early exit isn't a real scenario since we expect all the contained profiles to be real profiles and not None: simplified to remove the return.

This function is for convenience and reuse in testing, its just a bunch of asserts testing that the contained dataset profiles in a segmented result set (or other result set) are plausibly similar.

python/tests/api/pyspark/experimental/test_profiler_function.py
```diff
@@ -4,25 +4,70 @@
...
+    for profile_a in profiles_a:
+        assert isinstance(profile_a, SegmentedDatasetProfileView)
+        segment = profile_a.segment
+        segmented_view_a = profile_a.profile_view
+        segmented_view_b = results_b.profile(segment)
+        if isinstance(segmented_view_b, DatasetProfile):
+            segmented_view_b = segmented_view_b.view()
+        assert isinstance(segmented_view_b, DatasetProfileView)
+        if segmented_view_a is None or segmented_view_b is None:
+            assert segmented_view_a == segmented_view_b
```
richard-rogers, line 41: Maybe early return after this assert?

python/tests/api/pyspark/experimental/test_profiler_function.py
```diff
@@ -4,25 +4,70 @@
...
+
+        columns_in_a = segmented_view_a.get_columns()
+        columns_in_b = segmented_view_b.get_columns()
+        if not columns_in_b:
+            assert columns_in_a == columns_in_b
```
richard-rogers, line 46: I might throw in an early return here too...

richard-rogers (approved):

python/tests/api/pyspark/experimental/test_profiler_function.py
jamie256: it doesn't check timestamps or other properties, so I wanted to do the deeper checks later. But you are correct we can simplify this a bit more. I will address some of the test code issues in a follow up change to get this out.

python/tests/api/pyspark/experimental/test_profiler_function.py
jamie256: still want to check the timestamps.

PR #1080, opened 2023-02-07
Add BentoML example to integrations

sagecodes:
- Add BentoML to integration examples
- Add link to BentoML example in examples readme


python/examples/integrations/bentoml/README.md
```diff
@@ -0,0 +1,166 @@
...
+        "sepal width": features[1],
+        "petal length": features[2],
+        "petal width": features[3],
+        "class_output": category,
+        "proba_output": prob,
+    }
+
+    # Log data+outputs with whylogs & write to WhyLabs.ai
+    profile_results = why.log(data)
+    profile_results.writer("whylabs").write()
```
murilommen: ok, wondering if this is the best approach to take with bento, since it will be a live service. wouldn't it be best to have a pre-configured rolling logger?
wondering because it can be hard on WhyLabs' side to handle several (order of 1000s) one-line profile writes, which would come on every `classify` call

python/examples/integrations/bentoml/train.py
```diff
@@ -0,0 +1,22 @@
+import bentoml
+from sklearn import datasets, neighbors
+
+# Load training data set
+iris = datasets.load_iris()
+X, y = iris.data, iris.target
+
+# Create & train the model
+knn = neighbors.KNeighborsClassifier(n_neighbors=5)
+knn.fit(X, y)
```
murilommen, line 10: it would be great to store the training data's profile as reference, using whylogs+WhyLabs :)

murilommen (changes requested):
from what I remember, there are ways to extend/modify the existing Dockerfile for bentoml applications. can't precise how, but check if there's a chance you can do something similar to [the startup method](https://docs.whylabs.ai/docs/fastapi-integration/) i did here

python/examples/integrations/bentoml/train.py
sagecodes: I am going to do that in the blog post! I originally had it in here too, but I felt like it wasn't actually part of integrating with BentoML. I could add it back in, or maybe we make a note saying you can make a reference profile with training data.

python/examples/integrations/bentoml/README.md
sagecodes: Good point. I am having trouble finding how to easily call a startup/shutdown function like that to initialize the rolling logger. I'll ask about in their community.

sagecodes: They don't have lifecycle built in right now, but I could probably init the roller in service.py and use it in the prediction function.

murilommen: are we solved on this?

murilommen (changes requested):
i believe there is an extra `profile.writer` on README. other than that looks good!

python/examples/integrations/bentoml/README.md
sagecodes: Resolved: startup and shutdown logic implemented with a class

murilommen (approved):
LGTM

python/examples/integrations/bentoml/README.md
```diff
@@ -0,0 +1,188 @@
...
+```python
+# create a class to handle startup, shutdown, and prediction logic with Bentoml
+class MyService:
+    def __init__(self):
+        # Initialize the whylogs rolling logger with a 5 minute interval
+        global logger
+        logger = why.logger(
+            mode="rolling", interval=5, when="M", base_name="bentoml_predictions"
+        )
+        logger.append_writer("whylabs")
```
parano, line 65: Hi @sagecodes, we recently introduced the `on_startup` lifecycle hook, which should make the logger initialization a lot cleaner to implement: https://docs.bentoml.org/en/latest/concepts/service.html#lifecycle-hooks

PR #1081, opened 2023-02-08
Work-around to translate ConditionCountMetric to V0-compatible metrics(s)

richard-rogers:
## Description

Work-around missing WhyLabs missing support for V1 profiles for `ConditionCountMetric`.

## Changes

Each condition on a column is turned into ~a~ three "fake" columns named `<magic_prefix><column_name>.<condition_name>.{count, matches,non_matches}`. The columns have a single `ColumnCountsMetric`, in which the `n` component is the total/matching/non-matching count of the condition's evaluations.

## Related

Relates to https://app.clickup.com/t/8669ew4h4


python/whylogs/migration/uncompound.py
```diff
@@ -67,6 +70,23 @@ def _uncompound_multimetric(col_name: str, metric_name: str, metric: MultiMetric
...
 
 
+def _uncompound_condition_count(
+    col_name: str, metric_name: str, metric: ConditionCountMetric
+) -> Dict[str, ColumnProfileView]:
+    result: Dict[str, ColumnProfileView] = dict()
+    for condition_name, count_component in metric.matches.items():
+        new_col_name = f"{col_name}.{condition_name}"
+        new_metric = ColumnCountsMetric(
+            metric.total,  # n
```
andyndang: @richard-rogers can we use named parameters here?

FelipeAdachi (commented):
Does this work right off the bat, or does the platform need special treatment for it?
If it needs extra treatment, how does the platform know that the new column is from a ConditionCountMetric?

richard-rogers:
> Does this work right off the bat, or does the platform need special treatment for it?
> If it needs extra treatment, how does the platform know that the new column is from a ConditionCountMetric?

It works immediately, but gives a very confusing user experience where the match count has to be viewed/monitored as the count of missing values. @jamie256 suggested additional hacks for the UI that would identify faked `ConditionCountMetric` columns and provide a better UX until we get whylogs V1 support in the platform.

python/whylogs/migration/uncompound.py
```diff
@@ -38,6 +41,13 @@ def _v0_compatible_image_feature_flag() -> bool:
     return False
 
 
+def _condition_count_magic_string() -> str:
+    """
+    Column name prefix for uncompounded ConditionCountMetric columns
+    """
+    return "CONDITION_COUNT_METRIC_MAGIC."
```
jamie256: How about "__whylabs." or "__whylabs.condition."?

jamie256 (commented):
This approach should work to adapt condition_count metrics into an existing metric that will work in the platform.

Let's update the PR title, and change the magic string,

jamie256 (approved):
Thanks for the update Richard! LGTM

PR #1086, opened 2023-02-09
Enable overriding the API endpoint hostname

andyndang:
## Description
Enable overriding the TLS hostname so that user can point to one DNS name for calling the API, but using the public DNS name to verify the domain.

This allows WhyLabs API to work in private link environment where users can't configure the endpoint to use the public domain name.

## Changes
- Support endpoint override
- Use a global cache for client so that we don't create a new client every time we call whylabs

### DIFF
--- a/python/whylogs/api/writer/whylabs.py
+++ b/python/whylogs/api/writer/whylabs.py
@@ -1,11 +1,15 @@
+import abc
+import copy
 import datetime
 import logging
 import os
 import tempfile
-from typing import IO, Any, Dict, List, Optional, Tuple
+from typing import IO, Any, Dict, List, Optional, Tuple, Union
+from urllib.parse import urlparse
 
 import requests  # type: ignore
 import whylabs_client  # type: ignore
+from urllib3 import PoolManager, ProxyManager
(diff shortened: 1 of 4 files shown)


python/whylogs/api/writer/whylabs.py
```diff
@@ -82,17 +145,105 @@ class WhyLabsWriter(Writer):
...
+        self._api_client = client
+        _API_CLIENT_CACHE[cache_key] = client
+        logger.debug(f"Created and updated new client for cache key: {cache_key}")
+
+    def _update_hostname_config(self, endpoint_hostname_override: str) -> None:
+        """
+        This method overrides the pool manager's new connection method to add the hostname
+
+        """
+        import urllib3
```
jamie256, line 277: Do you want to add urllib3 as a requirement in the whylabs extra in pyproject.toml file (next to requests and whylabs-client)?

jamie256 (dismissed):
This is so much cleaner, thanks for the nice incidental refactor around key refresh and adding api client caching.

Looks good pending test updates for green CI (some tests are using badly formed api keys).

jamie256 (approved):
thanks for updating the dependencies

PR #1095, opened 2023-02-16
Fix error logging categorical dtype integer values

### LINKED ISSUE #1093: dataframe with categorical dtype column breaks logging
## Description

I have a pandas DataFrame with a column with binary values of 0 and 1. Due to prior processing, this column has a pandas dtype of `category`, which breaks the logging process, with the following error:

`TypeError: 'Categorical' with dtype category does not support reduction 'mean'`

This snippet should be able to reproduce the error:

```python
import pandas as pd
import whylogs as why

data = {
        "animal": ["cat", "hawk", "snake", "cat"],
        "legs": [4, 2, 0, 4],
        "full_time": [0, 1, 0, 0],
    }

df = pd.DataFrame(data)
df["full_time"] = df["full_time"].astype('category')


results = why.log(df)
```

jamie256:
## Description

pandas dataframes containing integer values with a dtype categorical hit error because the values were not converted to int types before passing to distribution metric.

## Changes

We already call astype(float) for the float numerics and to avoid this kind of processing error we need to do the same for integral types in preprocessing.
Added a test that repros the issue and shows string categorical values are already ok.

## Related

Closes #1093 

### DIFF
--- a/python/tests/api/logger/test_logger.py
+++ b/python/tests/api/logger/test_logger.py
@@ -99,6 +99,19 @@ def test_lending_club(lending_club_df: pd.DataFrame) -> None:
     assert len(df) == 151
 
 
+def test_categorical_dtype() -> None:
+    data = {"can_fly": [0, 1, 0, 0], "habitat": ["forest", "forest", "river", "river"]}
+
+    df = pd.DataFrame(data)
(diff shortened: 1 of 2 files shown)


FelipeAdachi (approved):
LGTM!
Thanks!

PR #1110, opened 2023-02-24
Add a new rolling logger type that handles historical data

naddeoa:
The current implementation of rolling logger has a few properties that make it unusable for the python whylogs container:

- It assumes all logged data has a dataset timestamp of now() and merges it into the current (only) profile.
- It has potential to upload data for future dates.
- It conflates upload schedule with dataset profile timestamp.
- It has the potential to tear logging between dataset profiles if logging happens during writing.

The new logger breaks the interface of the current Logger to enable different behavior. Now, the log() method takes in a timestamp that indicates when the data being logged happened, which is used to determine what dataset profile or segment cache the data should be logged to. Internally, it maintains a map of dataset timestamp to dataset profile/segment cache. The dataset timestamp is created by the aggregate_by configuration, where it can be Hourly, Daily, etc. For Hourly, the timestamp will be truncated to the hour.

Dataset timestamps are always the floor of the time interval so no data can be uploaded to future times.

You can separately configure the upload interval and the model type, which is important for the container since it has to distinguish between the two in the context of late data.

The tearing issue is resolved by using a message queue internally for all actions so that they're handled in serial. This does mean that logging is paused while uploading happens but log() is also asynchronous by default now so that won't add additional latency to the caller. The message queue is processed using a Thread. It's up to the caller to close the logger during application shutdown as well, which will start a synchronous series of cleanup events and profile writing.

### DIFF
--- a/python/examples/experimental/Multi dataset logger.ipynb
+++ b/python/examples/experimental/Multi dataset logger.ipynb
@@ -1,0 +1,153 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d026913b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%pip install whylogs[whylabs]"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "id": "c5ac0c04",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from whylogs.api.logger.experimental.multi_dataset_logger.multi_dataset_rolling_logger import MultiDatasetRollingLogger\n",
+    "from whylogs.api.logger.experimental.multi_dataset_logger.time_util import TimeGranularity, Schedule\n",
+    "from whylogs.api.writer.whylabs import WhyLabsWriter\n",
+    "\n",
+    "import os, getpass"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "id": "510d69d6",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Enter your WhyLabs Org ID\n",
+      "org-123\n"
+     ]
+    }
+   ],
+   "source": [
+    "print(\"Enter your WhyLabs Org ID\")\n",
+    "os.environ[\"WHYLABS_DEFAULT_ORG_ID\"] = input()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "id": "592b1f42",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Enter your WhyLabs API key\n",
+      "········\n",
+      "Using API Key ID:  xxxxxxxxxx\n"
+     ]
+    }
+   ],
+   "source": [
+    "print(\"Enter your WhyLabs API key\")\n",
+    "os.environ[\"WHYLABS_API_KEY\"] = getpass.getpass()\n",
+    "print(\"Using API Key ID: \", os.environ[\"WHYLABS_API_KEY\"][0:10])\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "id": "92827496",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Create a logger that will write hourly profiles to WhyLabs every 10 minutes\n",
+    "writer = WhyLabsWriter(\n",
+    "    org_id=os.environ[\"WHYLABS_DEFAULT_ORG_ID\"], \n",
+    "    api_key=os.environ[\"WHYLABS_API_KEY\"], \n",
+    "    dataset_id=\"model-1\"\n",
+    ")\n",
+    "schedule = Schedule(cadence=TimeGranularity.Minute, interval=10)\n",
+    "logger = MultiDatasetRollingLogger(\n",
+    "    aggregate_by=TimeGranularity.Hour, \n",
+    "    write_schedule=schedule,\n",
+    "    writers=[writer],\n",
+    "    # Can optionally provide a dataset schema here as well.\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 10,
+   "id": "e029ae5c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "data = [\n",
+    "    {'col1': 2, 'col2': 6.0, 'col3': 'FOO'},\n",
+    "    {'col1': 57, 'col2': 7.0, 'col3': 'BAR'},\n",
+    "    {'col1': 2, 'col2': 9.0, 'col3': 'FOO'},\n",
+    "    {'col1': 60, 'col2': 1.1, 'col3': 'FOO'},\n",
+    "]"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 11,
+   "id": "337e97f1",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Log a collectio nof rows. It can be a single row, or a pandas data frame as well.\n",
+    "logger.log(data)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "066c2dfc",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Manually trigger a write.\n",
+    "logger.flush()\n",
+    "\n",
+    "# Writes also automatically happen when you close the logger\n",
+    "logger.close()"
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python 3 (ipykernel)",
+   "language": "python",
+   "name": "python3"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.10.8"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 5
+}
--- a/python/whylogs/api/logger/logger.py
+++ b/python/whylogs/api/logger/logger.py
@@ -70,7 +70,14 @@ class Logger(ABC):
         pandas: Optional[pd.DataFrame] = None,
         row: Optional[Dict[str, Any]] = None,
         schema: Optional[DatasetSchema] = None,
+        timestamp_ms: Optional[int] = None,  # Not the dataset timestamp, but the timestamp of the data
     ) -> ResultSet:
+        """
+        Args:
+            timestamp_ms: The timestamp of the data being logged. This defaults to now if it isn't provided.
+             This is used to determine what the dataset timestamp should be. For an hourly model, the dataset
+             timestamp will end up being the start of the hour of the provided timestamp_ms, UTC.
+        """
         if self._is_closed:
             raise LoggingError("Cannot log to a closed logger")
         if obj is None and pandas is None and row is None:
--- a/python/whylogs/api/store/profile_store.py
+++ b/python/whylogs/api/store/profile_store.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from datetime import datetime, timedelta
-from typing import List
+from typing import Generator, List
 
 from whylogs.api.store.query import BaseQuery, DateQuery
 from whylogs.core import DatasetProfileView
@@ -15,12 +15,18 @@ class ProfileStore(ABC):
     def get(self, query: BaseQuery) -> DatasetProfileView:
         pass
 
+    # TODO this has to be updated to be able to handle more types of things. It might be better to just
+    # accept any Writable instead of the concrete types, like Writer does.
     @abstractmethod
     def write(self, profile_view: DatasetProfileView, dataset_id: str) -> None:
         pass
 
     @staticmethod
-    def _get_date_range(query: DateQuery):
+    def _get_date_range(query: DateQuery) -> Generator[datetime, None, None]:
+        if query.end_date is None:
+            yield query.start_date
+            return
+
         for n in range(int((query.end_date - query.start_date).days) + 1):
             yield query.start_date + timedelta(n)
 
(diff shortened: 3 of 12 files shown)


python/whylogs/api/logger/logger.py
```diff
@@ -85,6 +92,7 @@ def log(
 
         profiles = self._get_matching_profiles(obj, pandas=pandas, row=row, schema=active_schema)
 
+        # @Jamie so we never track to the main profile if there are segments defined in the schema?
```
jamie256: there is no such thing as a "main" profile when there are segments, there are multiple dataset profiles each with a unique segment key.

python/whylogs/api/logger/segment_cache.py
```diff
@@ -48,4 +48,5 @@ def flush(self, dataset_timestamp: Optional[datetime]) -> SegmentedResultSet:
         self._cache = dict()
         if dataset_timestamp:
             results.set_dataset_timestamp(dataset_timestamp)
+        # @Jamie shouldn't this be resetting its internal state after? In what sense is this a "flush" if nothing changes. It's just a "get_result_set()".
```
jamie256: self._cache is the internal state, which is recreated after the creation of the SegmentedResulSet above (line 48).

python/whylogs/api/store/profile_store.py
```diff
@@ -15,17 +15,24 @@ def list(self) -> List[str]:
...
+        if query.end_date is None:
+            yield query.start_date
+            return
+
         for n in range(int((query.end_date - query.start_date).days) + 1):
             yield query.start_date + timedelta(n)
 
     def _get_dates(self, query: DateQuery) -> List[datetime]:
         dates = []
+        a = self._get_date_range(query=query)
```
jamie256: is this for debugging? maybe logger.debug it or remove `a`

python/whylogs/api/logger/multi_dataset_logger/multi_dataset_rolling_logger.py
```diff
@@ -0,0 +1,362 @@
...
+        has_pending = False
+        for writer, pending in self._writers.items():
+            has_pending = len(pending) > 0
+        return has_pending
+
+    def _process_flush_message(self, message: FlushMessage) -> None:
+        for dataset_timestamp, container in self._cache.items():
+            self._logger.debug(f"Generating result set for dataset timestamp {dataset_timestamp}")
+
+            for writer, pending in self._writers.items():
```
jamie256: do you need the `writer ` here? Maybe you can use `self._writers.values()`

python/whylogs/api/logger/multi_dataset_logger/multi_dataset_rolling_logger.py
```diff
@@ -0,0 +1,362 @@
...
+                    if not success:
+                        self._logger.error(f"Couldn't write profile: {msg}")
+                        failed = True
+                except Exception as e:
+                    self._logger.exception(e)
+                    failed = True
+
+                if failed:
+                    p.attempts += 1
+                    if p.attempts < 3:
```
jamie256: consider config value for retry attempts?

python/whylogs/api/logger/multi_dataset_logger/multi_dataset_rolling_logger.py
```diff
@@ -0,0 +1,362 @@
...
+        Get the ResultSet of the contained profile/segment.
+
+        This doesn't have any side effects. It generates a ResultSet of whatever
+        is inside when this is called.
+        """
+        try:
+            if isinstance(self._target, SegmentCache):
+                return self._target.flush(dataset_timestamp=self._dataset_timestamp)
+            elif isinstance(self._target, DatasetProfile):
+                return ProfileResultSet(self._target)
```
jamie256: Is it possible this profile has processed 2 out of three columns of a row?

jamie256 (commented):
Left a few minor comments and answers to questions.

Can we add a notebook example showing how to use a multi dataset logger?

python/whylogs/api/logger/multi_dataset_logger/multi_dataset_rolling_logger.py
```diff
@@ -0,0 +1,362 @@
...
+            for row in data:
+                segment_processing(self._schema, row=row, segment_cache=self._target)
+        else:
+            segment_processing(self._schema, data, segment_cache=self._target)
+
+    def _track_profile(self, data: TrackData) -> None:
+        if not isinstance(self._target, DatasetProfile):
+            raise Exception("Dataset profile missing in logger")
+
+        if isinstance(data, List):
```
murilommen: what exactly do you expect here? a List of dicts or so?

python/whylogs/api/logger/segment_cache.py
naddeoa: ok. I just ignore this feature and recreate the segment cache/dataset pofiles in the logger.

python/whylogs/api/store/profile_store.py
naddeoa: woops yea.

python/whylogs/api/logger/multi_dataset_logger/multi_dataset_rolling_logger.py
naddeoa: I think we just put this off for now until we have a reason to set it. It's easy to create that config later and I think three will probably be fine for a long long time. Just another thing to document and decide on for users.

python/whylogs/api/logger/multi_dataset_logger/multi_dataset_rolling_logger.py
naddeoa: You mean something like this?

```python

row1 = {'col1': 2}
row2 = {'col2': 3}
logger.log(row1, timestamp_ms=foo)
logger.log(row2, timestamp_ms=foo-one_hour)
```

Then there would be two profiles with different column layouts yeah. Is that bad?


If you meant some rows go into the segment cache and some go into the dataset profile then that can't happen. If the schema has segments then it just always uses segment cache.

python/whylogs/api/logger/multi_dataset_logger/multi_dataset_rolling_logger.py
naddeoa: No its duplicate type checking. You need this for type safety even though it can't be anything else at this point. Basically sanity checks that should never happen because we catch them in unit tests if we violate those constriants.

naddeoa:
@jamie256 FYI, I had to make another change to the dependencies. For some reason, we were depending on protobuf >= 4.21, but if you do that then it doesn't work locally (at least on ubuntu), it forces you to set the `PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python` env var or you get a runtime error for tests (and probably everything else too).

The CI gets around this by explicitly installing `3.19` protobuf and happens to pick that one over the one that pip installed installed one. I just made the python dependency `3.19` as well. Is there some history behind that or is it just a mistake?

naddeoa:
> Left a few minor comments and answers to questions.
> 
> Can we add a notebook example showing how to use a multi dataset logger?

@jamie256 We don't actually depend on jupyter as a dev dependency? I'll add it in unless there's some reason.

naddeoa:
Added an example for the logger in examples/experimental.

python/whylogs/api/logger/multi_dataset_logger/multi_dataset_rolling_logger.py
jamie256: different layouts is fine. We can merge these correctly. I was wondering more about logging large dataframes that could take significant time and then calling flush while this processing is in flight, but maybe that is a question better scoped to multi-process changes.

jamie256:
> protobuf >= 4.21

We depend on protobuf >= 3.19.4
The package's install dependencies are defined in the pyproject.toml file:
https://github.com/whylabs/whylogs/blob/mainline/python/pyproject.toml#L16

jamie256:
> @jamie256 FYI, I had to make another change to the dependencies. For some reason, we were depending on protobuf >= 4.21, but if you do that then it doesn't work locally (at least on ubuntu), it forces you to set the `PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python` env var or you get a runtime error for tests (and probably everything else too).
> 
> The CI gets around this by explicitly installing `3.19` protobuf and happens to pick that one over the one that pip installed installed one. I just made the python dependency `3.19` as well. Is there some history behind that or is it just a mistake?

The CI installs the compiler, **protoc** https://github.com/whylabs/whylogs/blob/mainline/.github/workflows/whylogs-ci.yml#L55 which is what is used to build the protobuf message definitions from our .proto files, protoc is not a python package. Separately whylogs package depends on protobuf python package (used to serialize/deserialize into the proto messages).

There are back-compat and support considerations here with respect to choice of protoc and protobuf versions, let's discuss outside this PR, I can help manage the depencies and get your tests passing.

For reference here is what I have running locally on mainline:
`make install && poetry shell && lsb_release -a && protoc --version && pip freeze | grep protobuf`

excerpt from output of the `lsb_release -a && protoc --version && pip freeze | grep protobuf` part
         
```
...
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.5 LTS
Release:        20.04
Codename:       focal
libprotoc 3.19.4
mypy-protobuf==3.4.0
protobuf==4.21.12
types-protobuf==4.21.0.5
```

python/examples/experimental/Multi dataset logger.ipynb
```diff
@@ -0,0 +1,118 @@
...
+    "    org_id=\"org-xxxx\", \n",
+    "    api_key=\"xxxxxxxxxx.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\", \n",
+    "    dataset_id=\"model-1\"\n",
+    ")\n",
+    "schedule = Schedule(cadence=TimeGranularity.Minute, interval=10)\n",
+    "logger = MultiDatasetRollingLogger(\n",
+    "    aggregate_by=TimeGranularity.Hour, \n",
+    "    write_schedule=schedule,\n",
+    "    writers=[writer],\n",
+    "    # Can optionally provide a dataset shcema here as well.\n",
```
jamie256:
```suggestion
    "    # Can optionally provide a dataset schema here as well.\n",
```

python/pyproject.toml
```diff
@@ -141,6 +141,7 @@ embeddings = [
 
 [tool.poetry.group.dev.dependencies]
 types-urllib3 = "^1.26.25.5"
+jupyter = "^1.0.0"
```
jamie256: This is ok, but do we need this?

python/examples/experimental/Multi dataset logger.ipynb
```diff
@@ -0,0 +1,118 @@
...
+      "Requirement already satisfied: six>=1.5 in /home/anthony/workspace/whylogs/python/.venv/lib/python3.10/site-packages (from python-dateutil->whylabs-client) (1.16.0)\n",
+      "Installing collected packages: whylabs-client\n",
+      "Successfully installed whylabs-client-0.4.2\n",
+      "Note: you may need to restart the kernel to use updated packages.\n"
+     ]
+    }
+   ],
+   "source": [
+    "%pip install whylabs-client\n",
+    "# Typically you would `pip install whylogs[whylabs]` for dependencies\n",
```
jamie256: This notebook will fail if you run it by itself because there is no whylogs installed, can you include something like:
%pip install 'whylogs[whylabs]'

python/examples/experimental/Multi dataset logger.ipynb
```diff
@@ -0,0 +1,118 @@
...
+   "execution_count": 14,
+   "id": "92827496",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Create a logger that will write hourly profiles to WhyLabs every 10 minutes\n",
+    "writer = WhyLabsWriter(\n",
+    "    org_id=\"org-xxxx\", \n",
+    "    api_key=\"xxxxxxxxxx.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\", \n",
+    "    dataset_id=\"model-1\"\n",
```
jamie256: typically we try to encourage people to use env variables for these values, you could grab the snippet from step 4 in this example: https://github.com/whylabs/whylogs/blob/mainline/python/examples/integrations/writers/Getting_Started_with_WhyLabsV1.ipynb

jamie256 (commented):
Let's clean up the dependencies and get CI green but otherwise looks good!

python/pyproject.toml
naddeoa: How else do you run the example notebooks?

naddeoa: I configure my system to never allow global pip dependencies FYI, so maybe you just have a single jupyter instance on your system that you use for everything? It seems like its part of the dev process here to use notebooks so it seemed right to add it to the dev dependencies instead of making this one package an exception.

python/examples/experimental/Multi dataset logger.ipynb
naddeoa: Actually a little confused about how this is "supposed" to work. If I do that then the notebook will fail right now because we haven't published yet. It actually works right now when run inside of poetry shell though. Is that still the right thing to do?

naddeoa:
Got the protobuf issues worked out.

Here's what happened
- I installed protoc via apt install protobuf-compiler and I got 3.12.x
- I compiled with that old version
- I attempted to run tests
- I got an error saying that I needed a protoc version >=3.19
- I resolved it by installing protobuf==^3.19 because its a runtime error and it doesn't happen with older protobuf

But it looks like upgrading my local protoc to 3.19 (by downloading the release from github) and regenerating the proto files makes the error go away without changing the protobuf python dependency

python/pyproject.toml
```diff
@@ -58,6 +57,7 @@ scikit-learn = {version = "^1.0.2", optional = true}
 # Fugue related dependencies
 fugue = {version = "^0.8.1", optional = true}
 requests = "^2.27"
+protobuf = "^4.22.0"
```
jamie256: This is still an unnecessary change to our install dependencies, we don't need to force users to upgrade their protobuf. 4.22 works but so does protobuf==3.19.4 with the protoc version that builds the message definitions in our CI.

python/pyproject.toml
jamie256: There are many ways to run notebooks, whylogs doesn't specifically require jupyter, .ipynb is an open file format.

whylogs already depends on `ipykernel` and `papermill` as a dev dependencies, which lets us run notebooks in automated tests, and have left it up to people how they want to run the examples. We do lean on colab as a convenient way for users browsing our github repo to click a button to run them. We could add this as convenience for contributors but we run these in colab or an IDE directly while developing.

jamie256 (commented):
The changes you made to pyproject.toml are not required for this PR, otherwise looks good.

jamie256 (dismissed):
Looks good!

jamie256 (approved):

PR #1113, opened 2023-03-01
Performance Estimation - Better upload to WhyLabs

FelipeAdachi:
## Description

Currently, to upload performance estimation results to WhyLabs, the user is required to log manually, by calling e.g. `log({"performance_estimation":0.67})`. This PR adds the ability to write to WhyLabs by calling `estimation_results.writer('whylabs').write()` (or `estimator.estimate(target_result_set).writer('whylabs').write()`), so the user is no longer required to do it manually.

This way, we avoid extra work by the user and also provide a standard for naming by defining a performance estimation result magic string.

## Changes

- change whylabs writer to deal with EstimationResult
- implemented EstimationResult writer to whylabs
- updated notebook example


jamie256 (approved):
LGTM!

PR #1118, opened 2023-03-03
Fix error detecting segments

naddeoa:
The if check wasn't catching the case where segments in the schema were an empty dictionary. That should be treated as having no segments

### DIFF
--- a/python/tests/api/logger/experimental/multi_dataset_logger/test_multi_dataset_rolling_logger.py
+++ b/python/tests/api/logger/experimental/multi_dataset_logger/test_multi_dataset_rolling_logger.py
@@ -1,9 +1,19 @@
 import logging
 from datetime import datetime, timedelta
-from typing import Any, Optional, Tuple
+from typing import Any, List, Optional, Tuple
 
+import numpy as np
 import pytest
 from dateutil import tz
+
+from whylogs.core.resolvers import MetricSpec, ResolverSpec
+from whylogs.core.schema import DatasetSchema, DeclarativeSchema
+from whylogs.experimental.extras.embedding_metric import (
+    DistanceFunction,
+    EmbeddingConfig,
+    EmbeddingMetric,
+)
+from whylogs.experimental.preprocess.embeddings.selectors import PCAKMeansSelector
 
 try:
     from whylogs.api.logger.experimental.multi_dataset_logger.multi_dataset_rolling_logger import (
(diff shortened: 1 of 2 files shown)


python/tests/api/logger/experimental/multi_dataset_logger/test_multi_dataset_rolling_logger.py
```diff
@@ -68,6 +78,35 @@ def test_happy_path() -> None:
     assert writer.write_count == 1
 
 
+def test_no_segments_schema() -> None:
+    writer = TestWriter()
+
+    # Make a fancy embedding schema with no segmentation
+    l: List[List[float]] = [[1.0 for i in range(0, 50)] for i in range(0, 30)]
```
jamie256: can we use a longer name there, single lower case L is hard to distinguish from digit one.

python/whylogs/api/logger/experimental/multi_dataset_logger/multi_dataset_rolling_logger.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, dataset_timestamp: int, schema: Optional[DatasetSchema]) -> N
             self._target = DatasetProfile(dataset_timestamp=self._dataset_timestamp)
 
     def _has_segments(self) -> bool:
-        return self._schema is not None and self._schema.segments is not None
+        return self._schema is not None and bool(self._schema.segments)
```
jamie256, line 52: The explicit bool is not needed: https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations

The check for not None is ok, but you can simplify to:
`return self._schema and self._schema.segments`

jamie256 (dismissed):
Looks fine

python/whylogs/api/logger/experimental/multi_dataset_logger/multi_dataset_rolling_logger.py
naddeoa: mypy doesn't like it if I remove the bool(). It infers the function as trying to return `Dict | None` instead of a boolean.

naddeoa: In fact, I need the `is not None` as well it seems

jamie256: I see you are correct with respect to matching that return type of `bool` and maintaining one liner simplicity.

jamie256 (approved):
LGTM!

PR #1120, opened 2023-03-04
Remove broken make target

naddeoa:
Bumping the dev build number actually happens via `make bump-build`, no need for `make bump-dev`, which doesn't do anything.


jamie256 (approved):
Thanks!

PR #1121, opened 2023-03-07
Add dtype for pd.Series

naddeoa:
There is a lot of logging happening because of missing dtypes. I'm not sure what the right dtype is but we should specify something to silence this error.

```
    ./whylogs/core/preprocessing.py:176: FutureWarning: The default dtype for empty Series will be 'object' instead of 'float64' in a future version. Specify a dtype explicitly to silence this warning.
```


bernease (dismissed):
This looks good and stops the repeated warning from my manual test. Approving.

bernease (approved):

PR #1124, opened 2023-03-07
Embeddings reference selector refactoring

richard-rogers:
## Description

Some `ReferenceSelector` refactoring.

## Changes

- Makes `ReferenceSelector` an abstract base class
- ~Splits [un]supervised cases into separate entry points~ Adds error checking to ensure arguments passed match [un]supervised case
- Splits PCA and K-means into separate classes so NLP can use K-means


jamie256 (dismissed):
LGTM!

bernease:
I'm a little more used to the scikit-learn APIs that has become standard in the Python data science community where the method name is the same for supervised and unsupervised with the exception of the inclusion of y. (And even if you include the y, you warn or ignore.)

I'm okay with pulling this in, but can I get some help understanding *why* we want to make this change?

richard-rogers:
> I'm a little more used to the scikit-learn APIs that has become standard in the Python data science community where the method name is the same for supervised and unsupervised with the exception of the inclusion of y. (And even if you include the y, you warn or ignore.)
> 
> I'm okay with pulling this in, but can I get some help understanding _why_ we want to make this change?



> I'm a little more used to the scikit-learn APIs that has become standard in the Python data science community where the method name is the same for supervised and unsupervised with the exception of the inclusion of y. (And even if you include the y, you warn or ignore.)
> 
> I'm okay with pulling this in, but can I get some help understanding _why_ we want to make this change?

I switched it to the scikit-learn convention and added argument validation code to ensure the arguments passed match the selector's expectation.

jamie256 (approved):
The warning looks better thanks!

bernease (approved):
Looks great! Thank you for adjusting to data scientists' refusal to use different methods for supervised and unsupervised. Engineers are much more flexible. :)

PR #1131, opened 2023-03-08
Fix reading older profiles without tensors counts

### LINKED ISSUE #1130: tensors break TypeCounters read of older dataset profiles
## Description

With the addition of a new required field in this Metric, older profiles will not have a 'tensors' field and break when reading back these profiles.

## Suggestions

We allow reading older profiles from 1.1.29 but default the tensor count to 0.

jamie256:
## Description

Fixes #1130 by creating a default value for the tensors field in TypeCountersMetric

## Changes

- added a test that reproduces the issue, but more realistically any older dataset profile file would not be readable by 1.1.29
- added default factory for the tensors field to instantiate an integral component when this field is not passed to init.

### DIFF
--- a/python/.bumpversion.cfg
+++ b/python/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.1.28
+current_version = 1.1.29
 tag = False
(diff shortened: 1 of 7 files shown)


FelipeAdachi (approved):
LGTM!

PR #1134, opened 2023-03-09
NLP Summarization Example - fix whylabs flag and colab button

FelipeAdachi:
## Description

NLP Summarization Example - fix whylabs flag and colab button


jamie256 (approved):
LGTM!

PR #1138, opened 2023-03-10
Add preprocessed column entry point for homogeneous columns

richard-rogers:
## Description

Adds a `_process_homogeneous_column()` entry point to `PreprocessedColumn` to improve performance for data frames with homogeneous columns (and no missing data/NaNs).

## Changes

- Adds `PreprocessedColumn::_process_homogeneous_column()`
- ~Plumbed through as a new optional argument to `why.log()`~
- Adds `homogeneous_column_names` set to `DatasetSchema` so users can specify which columns take the faster code path


python/tests/core/test_preprocessing.py
```diff
@@ -682,6 +686,27 @@ def test_apply_list(
             assert X.shape == Y.shape
 
 
+def test_homogeneous_column() -> None:
+    d = {"col1": [1, 2, 3, 4], "col2": [5.0, 6.0, 7.0, 8.0], "col3": ["a", "b", "c", "d"]}
+    df = pd.DataFrame(data=d)
+    types = {"col1": (int, ColumnProperties.homogeneous), "col2": (float, ColumnProperties.default)}
```
jamie256: TODO: follow up with a test that has something like:
* "col2": float
* "col2: (float)

python/whylogs/core/preprocessing.py
```diff
@@ -250,6 +256,46 @@ def _process_scalar_value(value: Any) -> "PreprocessedColumn":
...
+
+        if pdc.is_numeric_dtype(series.dtype) and not pdc.is_bool_dtype(series.dtype):
+            if pdc.is_float_dtype(series.dtype):
+                result.numpy.floats = series.astype(float)
+                return result
+            else:
+                result.numpy.ints = series.astype(int)
+                return result
+
+        value = series[0]
```
jamie256, line 282: comment here to draw attention that this is the unique aspect of the homogenous processing

python/whylogs/core/schema.py
```diff
@@ -99,8 +100,13 @@ def __init__(
             )
 
         for col, data_type in self.types.items():
+            if isinstance(data_type, tuple) and len(data_type) == 2 and isinstance(data_type[1], ColumnProperties):
+                dtype = data_type[0]
+            else:
+                dtype = data_type
```
jamie256, line 105: make sure we can handle variable length tuples

jamie256 (commented):
Looks good, just need to verify we can handle variable tuple types and not break.

python/whylogs/core/dataset_profile.py
```diff
@@ -154,6 +155,18 @@ def _do_track(
                     logger.error(
                         f"whylogs was passed a pandas DataFrame with key [{k}] but DataFrame.get({k}) returned nothing!"
                     )
+                    return
+
+                dtype = self._schema.types.get(k)
+                homogeneous = (
+                    dtype is not None
+                    and isinstance(dtype, tuple)
+                    and len(dtype) == 2
```
richard-rogers, line 164: drop length requirement

jamie256 (approved):
LGTM!

PR #1139, opened 2023-03-13
faster embedding submetric data preprocessing

richard-rogers:
## Description

Don't use Pandas series for submetric input. 25x speedup in benchmarking.


bernease (approved):
LGTM!

naddeoa:
Can we get an example snippet here for how the configuration has to look to take advantage of this?

PR #1140, opened 2023-03-13
Make writing loop a little better

naddeoa:
Not a huge deal but we generate fewer redundant result sets if we invert this loop when there are multiple writers.


jamie256 (approved):
Agreed!

PR #1142, opened 2023-03-14
trigger registration of non-standard metrics for profile deserialization

### LINKED ISSUE #1137: If you haven't instantiated embeddings then deserializing profiles with embeddings metrics hits error
## Description

Error is:
```
whylogs/core/view/column_profile_view.py:132, in ColumnProfileView.from_protobuf(cls, msg)
    130         metric_class = _METRIC_DESERIALIZER_REGISTRY[m_name]
    131     else:
--> 132         raise UnsupportedError(f"Unsupported metric: {m_name}")
    133 else:
    134     metric_class = m_enum.value

UnsupportedError: Unsupported metric: embedding
```
To reproduce the error, write a profile with embeddings metric in it. Using whylogs 1.1.30

then run the following code to read and deserialize the file:
```
from whylogs.core import  DatasetProfileView

profile_view = DatasetProfileView.read(profile_file_name)
```
## Suggestions

register embeddings metric so we can deserialize profiles with them without referencing the embeddings metric first.

richard-rogers:
## Description

Triggers registration of non-standard metrics to allow deserialization of profiles containing such metrics.

## Changes

- Adds stubs for Scipy
- Ensure non-standard metrics use stubs for extra dependencies
- `ColumnProfileView::from_protobuf()` imports the non-standard metrics to trigger their registration

## Related

Relates to whylabs/whylogs#1137

Closes whylabs/whylogs#1137

### DIFF
--- a/python/whylogs/core/stubs.py
+++ b/python/whylogs/core/stubs.py
@@ -16,6 +16,11 @@ except ImportError:  # noqa
     if _pd is not None:
         logger.error("Pandas is installed but numpy is not. Your environment is probably broken.")
 
+try:
+    import scipy as _sp
+except ImportError:  # noqa
+    _sp = None  # type: ignore
+
 
 class _StubClass:
     pass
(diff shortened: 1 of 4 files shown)


FelipeAdachi (commented):
The `viz` module also uses scipy. Do we need to make any changes there as well?
If we should, but it isn't a priority and is not breaking anything, I can do it in a follow-up.

richard-rogers:
> The `viz` module also uses scipy. Do we need to make any changes there as well?
> If we should, but it isn't a priority and is not breaking anything, I can do it in a follow-up.

I assume `viz` has Scipy as a hard requirement, so I don't think it needs to make any changes.

jamie256 (approved):
LGTM!

PR #1143, opened 2023-03-14
Add ConformalMetric

richard-rogers:
## Description

Adds a metric that can masquerade as any other metric that uses only standard metric components. This can be used to deserialize custom metrics when the source code of the custom metric is unavailable.  The conformal metric has limited functionality, but should be usable in a read-only profile.

## Changes

- Adds `ConformalMetric`
- Use `ConformalMetric` as last-ditch effort to deserialize a non-standard metric that has not registered as a deserializable metric class


python/whylogs/core/metrics/conformal_metric.py
```diff
@@ -0,0 +1,69 @@
...
+        raise NotImplementedError
+
+    @property
+    def namespace(self) -> str:
+        return self._namespace
+
+    # NOTE: This only works for metrics for which merge is strictly component-wise merging.
+    def merge(self, other: Metric) -> "ConformalMetric":
+        if other.namespace != self._namespace:
+            raise ValueError(f"Attempted to merge {self._namespace} with incompatible {other.namespace} metric")
```
FelipeAdachi, line 26: maybe add a test to cover this case?

richard-rogers:
Keep the branch around as an example or possible future resurrection, but don't merge to mainline at this time.

PR #1147, opened 2023-03-17
Data Validators - Identity Columns and Reservoir Sampling

### LINKED ISSUE #1108: Support keeping a sample of failed row ids for data validators
## Description

When running the data validation as part of profiling, we should support user specifying a row id and then we can keep a buffer of the row ids of any failed row.

FelipeAdachi:
## Description

- Reservoir sampling:
  - Sampling failed conditions defaults to true (sampling size default is 10)
  - If identity column not passed, samples actual validated values
  - If identity column is passed, samples row ids for failed values
- Identity columns
  - Defaults to None
  - If identity column is passed, the row ids are passed to the actions, in addition to being sampled (if sampling is enabled)

### DIFF
--- a/python/whylogs/core/validators/condition_validator.py
+++ b/python/whylogs/core/validators/condition_validator.py
@@ -1,13 +1,20 @@
 import logging
 from dataclasses import dataclass, field
-from itertools import chain
-from typing import Any, Callable, Dict, List, Union
+from typing import Any, Callable, Dict, List, Union, Optional
 
 from whylogs.core.metrics.condition_count_metric import Condition
-from whylogs.core.preprocessing import PreprocessedColumn
 from whylogs.core.validators.validator import Validator
+from whylogs.core.metrics.metrics import MetricConfig
+from whylogs.core.stubs import pd
+import whylogs_sketching as ds
 
 logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class ConditionValidatorConfig(MetricConfig):
+    validator_sampling_size: int = 10
+    identity_column: Optional[str] = None
 
 
 @dataclass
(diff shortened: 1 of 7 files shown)


python/whylogs/core/validators/condition_validator.py
```diff
@@ -1,32 +1,72 @@
...
-from whylogs.core.preprocessing import PreprocessedColumn
+from whylogs.core.metrics.metrics import MetricConfig
+from whylogs.core.stubs import pd
 from whylogs.core.validators.validator import Validator
 
 logger = logging.getLogger(__name__)
 
 
+@dataclass(frozen=True)
+class ConditionValidatorConfig(MetricConfig):
```
jamie256: Should we put this into metric config?

python/whylogs/core/validators/condition_validator.py
```diff
@@ -1,32 +1,72 @@
...
+    """
+    Configuration for a condition validator.
+
+    Args:
+        validator_sample_size (int): The number of failed values to sample. Defaults to 10.
```
jamie256: If used in a datasketch k param, this wouldn't be the count of values but a log(number_of_samples) = k, need to update the description.

python/whylogs/core/validators/condition_validator.py
```diff
@@ -1,32 +1,72 @@
...
-    def columnar_validate(self, data: PreprocessedColumn) -> None:
+    def columnar_validate(self, data: Any, sampling_size: int = 10, identity_values: Any = None) -> None:
         count = 0
         count_failures = 0
-        for x in list(chain.from_iterable(data.raw_iterator())):
+        validate_with_row_id = False
+        if not isinstance(sampling_size, int):
+            raise ValueError("sampling_size must be an integer")
+        if self.enable_sampling:
+            self._sampler = ds.var_opt_sketch(k=sampling_size)
```
jamie256: Can we move to *init__?

jamie256 (commented):
Some initial comments on code walkthrough we did.

python/whylogs/core/validators/condition_validator.py
jamie256: Ok, I am wrong, its actually the number of samples

python/whylogs/core/validators/condition_validator.py
```diff
@@ -41,11 +81,30 @@ def columnar_validate(self, data: PreprocessedColumn) -> None:
...
                     for action in self.actions:
-                        action(self.name, cond_name, x)
+                        if validate_with_row_id:
+                            action(self.name, cond_name, x, identity_values[index])
+                        else:
+                            action(self.name, cond_name, x)  # type: ignore
 
         self.total = count
 
+    def sample_failed_conditions(self) -> List[Any]:
```
jamie256:
```suggestion
    def get_samples(self) -> List[Any]:
```

To match the sketch's name. We aren't actively computing the sample with this, just reading it out of the sketch.

python/whylogs/core/validators/condition_validator.py
```diff
@@ -1,32 +1,72 @@
...
 
-    def columnar_validate(self, data: PreprocessedColumn) -> None:
+    def columnar_validate(self, data: Any, sampling_size: int = 10, identity_values: Any = None) -> None:
         count = 0
         count_failures = 0
-        for x in list(chain.from_iterable(data.raw_iterator())):
+        validate_with_row_id = False
+        if not isinstance(sampling_size, int):
+            raise ValueError("sampling_size must be an integer")
+        if self.enable_sampling:
```
jamie256:
```suggestion
        if self.enable_sampling and self._sampler is None:
```

python/whylogs/core/validators/condition_validator.py
FelipeAdachi: done in latest commit

python/whylogs/core/validators/condition_validator.py
FelipeAdachi: So, it was not as straightforward to simply move to init: when the validator is instantiated, it doesn't have access to the sampling size, as this was present only in the `schema`.

To be able to put this on `post_init`, I had to remove the `sample_size` param from `MetricConfig` and put it as a parameter of the `Validator` itself.

jamie256 (approved):
LGTM!

PR #1149, opened 2023-03-21
pyspark embeddings support on vectors

jamie256:
## Description

the pyspark mapInPandas does not support dataframe that contain vectors, and the whylogs pyspark profiler uses mapInPandas. To support embeddings which might be passed in as DenseVectors we check for any columns of vector type and convert these to arrays inside the whylogs `collect_dataset_profile_view` if present. This creates a transformed dataframe which conforms to the requirements of mapInPandas and is consistent with how whylogs profiles embeddings (counted as tensor types).

As a side effect we now can profile dataframes that contain vector columns, where before this change a vector column would cause an error in the whylogs profiling when using pyspark (even if the configuration did not include embeddings metrics).

## Changes

- Check for VectorUDT columns inside pyspark profiling functions
- transform vectors to arrays before passing to mapInPandas inside whylogs pyspark profiling.
- added tests for vectors and embeddings metrics with pyspark.

### DIFF
--- a/python/tests/api/pyspark/experimental/test_profiler_function.py
+++ b/python/tests/api/pyspark/experimental/test_profiler_function.py
@@ -2,6 +2,7 @@ from decimal import Decimal
 from logging import getLogger
 from typing import Dict
 
+import numpy as np
 import pytest
 
 import whylogs as why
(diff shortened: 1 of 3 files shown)


FelipeAdachi (dismissed):
LGTM!

FelipeAdachi (approved):
LGTM!

PR #1151, opened 2023-03-22
preserve mlflow existing run when using mlflow writer

### LINKED ISSUE #1129: [mlflow] We should not end the run if we didn't start it
By default `end_run` is set to `True`:

https://github.com/whylabs/whylogs/blob/ba50b60dff602f5e67ea32e477904623140c8aa8/python/whylogs/api/writer/mlflow.py#L35

jamie256:
## Description

Fixes #1129 by recording if whylogs started the mlflow run or not and only end it if whylogs also started it and the setting to end the run is True (default value).

## Changes

- check if there is a preexisting mlflow run before creating one in the whylogs mlflow writer
  - only end the run if it was also started locally in the write method and the setting is True
- Added new test to check that we don't end preexisting mlflow runs
- Minor test fix for the MLFlow writer tests to not throw if there are no artifacts or runs directory which is the case in the new test if run standalone.


FelipeAdachi (approved):
LGTM!

PR #1159, opened 2023-03-24
Upgrade python-setup and checkout actions

jamie256:
## Description

Upgrade python-setup action to v4 and use the built in cache for poetry dependencies + smaller CI fixes to speed up execution. This results in significantly faster CI runs, saving more than 10 minutes of time per run and avoids a lot of wasted dependency reinstalls via better cache hit rates.

- Fix a few missed places that should have used env variables
- Fix release workflows saving of version string
- remove two of the slowest notebooks tests


FelipeAdachi (approved):
LGTM!

PR #1162, opened 2023-03-28
Multi-column UDF support

richard-rogers:
## Description

Extends `DatasetSchema` with UDFs to compute new columns from existing columns (prior to resolving column metrics).

## Changes

* Adds new `UdfSchema` that supports dataset (multi-column) UDFs
* Register dataset UDFs with `@register_dataset_udf` decorator
* New privacy-preserving default submetric schema for `UdfMetric` 
* Deprecates `generate_udf_schema()` in favor of `generate_udf_resolvers()`


FelipeAdachi (commented):
I personally like this approach better than the CallbackMetric

python/tests/core/test_declarative_schema.py
```diff
@@ -202,3 +203,29 @@ class UnknownType:
             reference_metrics = set(reference_results.get_column(column).get_metric_names())
             declarative_metrics = set(declarative_results.get_column(column).get_metric_names())
             assert reference_metrics == declarative_metrics
+
+
+def test_augmenters_row() -> None:
+    schema = DeclarativeSchema(
+        STANDARD_RESOLVER,
+        augmenters=[AugmenterSpec(column_name="col1", callbacks={"col2": lambda x: x, "col3": lambda x: x})],
```
FelipeAdachi: what happens if the callback specifies the same col name (col1)? will it overwrite the original column?
what would be the ideal behavior in this case? don't allow it or let it overwrite? Or, maybe, create new cols, like `col1_aug_1` or something similar?

python/whylogs/core/schema.py
```diff
@@ -76,6 +85,7 @@ def __init__(
         schema_based_automerge: bool = False,
         segments: Optional[Dict[str, SegmentationPartition]] = None,
         validators: Optional[Dict[str, List[Validator]]] = None,
+        augmenters: List[AugmenterSpec] = [],
```
jamie256: Move to an experimental subclass of DatasetSchema

jamie256 (commented):
Lets try these out a bit to see how they work with some of the scenarios we discussed around CV and NLP.

Minor comment about moving to experimental namespace while we are iterating on this API, but don't need to do that yet.

python/tests/core/test_declarative_schema.py
richard-rogers: Good question :)

python/whylogs/core/schema.py
```diff
@@ -22,6 +22,15 @@
 T = TypeVar("T", bound="DatasetSchema")
 
 
+@dataclass
+class AugmenterSpec:
+    column_name: Optional[str] = None  # TODO: maybe make this a regex
+    column_type: Optional[Any] = None
+    callbacks: Dict[str, Callable[[Any], Any]] = field(
```
richard-rogers: maybe `callbacks: List[Callable[[Any], Tuple[str, Any]]]` where the returned string is the new column name.

python/whylogs/core/schema.py
```diff
@@ -161,6 +173,30 @@ def _resolve_pdf(self, df: pd.DataFrame, force_resolve: bool = False) -> bool:
 
         return dirty
 
+    def _augment(
```
richard-rogers: Maybe just make *this* the callback...

richard-rogers:
s/callback/udf/
s/augmenter/?/ -- computer, extender, ...

python/whylogs/core/schema.py
jamie256: Default value here should not be an empty list (current implementation will create a shared instance across instances of DatasetSchema). Add default_factory here to get separate instances of the list for each DatasetSchema instance.

You may need to add imports for `field` and `default_factory`
```suggestion
        augmenters: List[AugmenterSpec] = field(default_factory=list),
```

python/whylogs/core/schema.py
```diff
@@ -22,6 +22,15 @@
 T = TypeVar("T", bound="DatasetSchema")
 
 
+@dataclass
+class AugmenterSpec:
```
jamie256: Suggest renaming the class: data augmentation is potentially a familiar term with respect to training ML models and we don't want to invoke that concept here.

What about naming it something like: Transforms, Transformer, ColumnTransformer, DerivedColumnSpec, DerivedColumn?

python/whylogs/core/schema.py
jamie256: also renaming applies to the parameter here similarly to the class name (lets move away from augment*

python/whylogs/core/schema.py
jamie256: This operation is a pre-profiling transform step, seems like it should live in Preprocessing conceptually. Having the pandas / row split in the signature is a leaky abstraction, it requires that the integrator know about this internal detail of how we split the inputs. Why not just pass the original data before the split, or we could pass them the extracted column where bound?

python/whylogs/core/schema.py
```diff
@@ -161,6 +173,30 @@ def _resolve_pdf(self, df: pd.DataFrame, force_resolve: bool = False) -> bool:
...
+    def _augment(
+        self, pandas: Optional[pd.DataFrame] = None, row: Optional[Mapping[str, Any]] = None
+    ) -> Tuple[Optional[pd.DataFrame], Optional[Mapping[str, Any]]]:
+        new_columns = dict(row) if row else None
+        if row is not None:
+            for column, value in row.items():
+                for augmenter in self.augmenters:
+                    col_name, col_type = augmenter.column_name, augmenter.column_type
+                    why_type = self.type_mapper(type(value))
+                    if (col_name and col_name == column) or (col_name is None and isinstance(col_type, why_type)):  # type: ignore
```
jamie256: Factor this out into separate function. Maybe this is a transforms resolve step?

You can simplify` (col_name and col_name == column)` to `(col_name == column)`

python/whylogs/core/schema.py
```diff
@@ -161,6 +173,30 @@ def _resolve_pdf(self, df: pd.DataFrame, force_resolve: bool = False) -> bool:
...
+                        for new_col, callback in augmenter.callbacks.items():
+                            new_columns[new_col] = callback(value)  # type: ignore
+
+        if pandas is not None:
+            for column in pandas.keys():
+                for augmenter in self.augmenters:
+                    col_name, col_type = augmenter.column_name, augmenter.column_type
+                    why_type = pandas.dtypes[column]
+                    if (col_name and col_name == column) or (col_name is None and isinstance(col_type, why_type)):  # type: ignore
```
jamie256: same comment as above, also will let us avoid the duplicated if statement.

python/whylogs/core/schema.py
```diff
@@ -161,6 +173,30 @@ def _resolve_pdf(self, df: pd.DataFrame, force_resolve: bool = False) -> bool:
...
+                        for new_col, callback in augmenter.callbacks.items():
+                            pandas[new_col] = pandas[column].map(callback)
```
jamie256: Let not add the derived columns to the input pandas dataframe which modifies the input. Can we keep these in a separate dataframe?

python/tests/core/test_declarative_schema.py
jamie256: Current implementation will overwrite the original input. We should at least emit an INFO log message that the derived column name exists in the input already.

jamie256 (commented):
Needs to be moved into experimental and not break/add to existing core classes yet to protect integrations from accidentally relying on this experimental extensibility point.

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -0,0 +1,83 @@
+import logging
+from copy import deepcopy
+from dataclasses import dataclass, field
+from typing import Any, Callable, Dict, List, Mapping, Optional, Tuple
+
+from whylogs.core.datatypes import TypeMapper
```
richard-rogers, line 6:
```suggestion
from whylogs.core.datatypes import DataType, TypeMapper
```

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -0,0 +1,83 @@
...
+from whylogs.core.stubs import pd
+from whylogs.core.validators.validator import Validator
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class UdfSpec:
+    column_name: Optional[str] = None  # TODO: maybe make this a regex
+    column_type: Optional[Any] = None
```
richard-rogers:
```suggestion
    column_type: Optional[DataType] = None
```

python/whylogs/core/schema.py
richard-rogers:
```suggestion
                    if col_name == column or (col_name is None and isinstance(col_type, why_type)):  # type: ignore
```

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -0,0 +1,83 @@
...
+        self, pandas: Optional[pd.DataFrame] = None, row: Optional[Mapping[str, Any]] = None
+    ) -> Tuple[Optional[pd.DataFrame], Optional[Mapping[str, Any]]]:
+        new_columns = deepcopy(row) if row else None
+        new_df = pd.DataFrame(pandas) if pandas is not None else None
+        if row is not None:
+            for column, value in row.items():
+                for udf_spec in self.udf_specs:
+                    col_name, col_type = udf_spec.column_name, udf_spec.column_type
+                    why_type = self.type_mapper(type(value))
+                    if (col_name and col_name == column) or (col_name is None and isinstance(col_type, why_type)):  # type: ignore
```
richard-rogers:
```suggestion
                    if col_name == column or (col_name is None and isinstance(col_type, why_type)):  # type: ignore
```

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -0,0 +1,83 @@
...
+                            if new_col in row:
+                                logger.info(f"Overwriting column {new_col}")
+                            new_columns[new_col] = udf(value)  # type: ignore
+
+        if pandas is not None:
+            for column in pandas.keys():
+                for udf_spec in self.udf_specs:
+                    col_name, col_type = udf_spec.column_name, udf_spec.column_type
+                    why_type = pandas.dtypes[column]
+                    if (col_name and col_name == column) or (col_name is None and isinstance(col_type, why_type)):  # type: ignore
```
richard-rogers:
```suggestion
                    if col_name == column or (col_name is None and isinstance(col_type, why_type)):  # type: ignore
```

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -0,0 +1,89 @@
...
+            resolvers=resolvers,
+            types=types,
+            default_config=default_config,
+            type_mapper=type_mapper,
+            cache_size=cache_size,
+            schema_based_automerge=schema_based_automerge,
+            segments=segments,
+            validators=validators,
+        )
+        self.udf_specs = udf_specs or []
```
richard-rogers:
```suggestion
        self.udf_specs = deepcopy(udf_specs) or []
```

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def copy(self) -> DatasetSchema:
+        copy = super().copy()
+        copy.udf_specs = deepcopy(self.udf_specs)
+        return copy
+
+    def _run_udfs_on_row(self, row: Mapping[str, Any], new_columns: Mapping[str, Any]) -> None:
+        for column, value in row.items():
+            for udf_spec in self.udf_specs:
```
richard-rogers: Change this to map from column name -> UDF list and map from column type -> UDF list for efficiency

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -0,0 +1,296 @@
...
+
+    You must specify exactly one of either col_name or col_type. col_name will attach
+    a UDF to the named input column(s). col_type will attach a UDF to all
+    input columns of the specified type. The decorated function will automatically
+    be a UDF in the UdfSchema.
+
+    Specify udf_name to give the output of the UDF a name. udf_name
+    defautls to the name of the decorated function. Note that all lambdas are
+    named "lambda", so omitting udf_name on more than one lambda will result
+    in name collisions.
```
FelipeAdachi, line 133: can we add a brief description on what the `metrics` param is for?

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    Decorator to easily configure UDFs for your data set. Decorate your UDF
+    functions, then call generate_udf_dataset_schema() to create a UdfSchema
+    that includes the UDFs configured by your decorator parameters. The decorated
+    function will automatically be a UDF in the UdfSchema.
+
+    Specify udf_name to give the output of the UDF a name. udf_name
+    defautls to the name of the decorated function. Note that all lambdas are
+    named "lambda", so omitting udf_name on more than one lambda will result
+    in name collisions.
```
richard-rogers, line 133:
```suggestion
    in name collisions.
    
    If any metrics are passed via the metrics argument, they will be attached
    to the column produced by the UDF via the schema returned by generate_udf_dataset_schema().
    If metrics is None, the UDF output column will get the metrics determined by
    the other resolvers passed to generate_udf_dataset_schema(), or the STANDARD_RESOLVER
    by default.
```

python/whylogs/experimental/core/metrics/udf_metric.py
```diff
@@ -305,6 +305,11 @@ def upper(x):
     return resolvers
 
 
+@deprecated(message="Please use generate_udf_resovlers()")
```
jamie256:
```suggestion
@deprecated(message="Please use generate_udf_resolvers()")
```

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -0,0 +1,208 @@
...
+    default_config: Optional[MetricConfig] = None,
+    type_mapper: Optional[TypeMapper] = None,
+    cache_size: int = 1024,
+    schema_based_automerge: bool = False,
+    segments: Optional[Dict[str, SegmentationPartition]] = None,
+    validators: Optional[Dict[str, List[Validator]]] = None,
+) -> UdfSchema:
+    """
+    Returns a UdfSchema that implements any registered UDFs, along with any
+    other_udf_specs or resovlers passed in.
```
jamie256: nit: typo 'resolvers'

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+
+    resolver_specs = resolvers + _resolver_specs if resolvers is not None else STANDARD_RESOLVER + _resolver_specs
```
jamie256: We had people not want frequent items tracked by default on the original features. The standard resolver is not a good general default for udf scenarios. We need a new default config that is standard -= frequent items.

This might be ok to have as an option, but I want the privacy preserving version as a helper method too.

jamie256 (dismissed):
Left some minor comments.

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def copy(self) -> DatasetSchema:
+        copy = super().copy()
+        copy.multicolumn_udfs = list(self.multicolumn_udfs)
+        return copy
+
+    def _run_udfs_on_row(self, row: Mapping[str, Any], new_columns: Dict[str, Any]) -> None:
+        for spec in self.multicolumn_udfs:
+            if set(spec.column_names).issubset(set(row.keys())):
+                _apply_udfs_on_dataframe(row, spec.udfs, new_columns)
```
richard-rogers:
```suggestion
                _apply_udfs_on_row(row, spec.udfs, new_columns)
```

python/whylogs/experimental/core/udf_schema.py
richard-rogers:
```suggestion
    other_udf_specs or resolvers passed in.
```

jamie256 (approved):
LGTM!

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -0,0 +1,210 @@
...
+            new_columns[new_col] = None
+            logger.exception(f"Evaluating multi-column UDF {new_col} failed")
+
+
+def _apply_udfs_on_dataframe(pandas: pd.DataFrame, udfs: Dict, new_df: pd.DataFrame) -> None:
+    for new_col, udf in udfs.items():
+        try:
+            new_df[new_col] = udf(pandas)
+        except Exception:  # noqa
+            # new_df[new_col] = pandas[pandas.keys()[0]].apply(lambda x: None)
```
richard-rogers: column with single `None`

jamie256 (dismissed):
LGTM!

jamie256 (approved):

PR #1163, opened 2023-03-28
constraints generation

FelipeAdachi:
## Description

initial implementation of Constraints Generation as a function under the `experimental` namespace.
First group auto-generated constraints:

- Count Metrics: If reference contain no `null` values, create `no_missing_values` constraints for given column
- Types Metrics: If type count is non-zero for a list of data types, creates `column_has_non_zero_types` constraints for the types list
- Condition Count Metrics: If column has a set of conditions attached to it, and the conditions didn't fail once, creates a `condition_meets` constraint

This is just the first few ideas for auto-generated constraints. More to follow.

### DIFF
--- a/python/tests/experimental/constraints_generation/test_constraints_generation.py
+++ b/python/tests/experimental/constraints_generation/test_constraints_generation.py
@@ -39,8 +39,7 @@ def reference_profile_view():
             ),
         ],
     )
-    schema = DeclarativeSchema(STANDARD_RESOLVER)
-    schema.add_resolver(legs_spec)
+    schema = DeclarativeSchema(STANDARD_RESOLVER + [legs_spec])
 
     results = why.log(df, schema=schema)
     profile_view = results.view()
--- a/python/tests/experimental/constraints_generation/test_constraints_generation.py
+++ b/python/tests/experimental/constraints_generation/test_constraints_generation.py
@@ -1,0 +1,67 @@
+import pandas as pd
+import pytest
+import whylogs as why
+import pandas as pd
+from whylogs.core.constraints import ConstraintsBuilder
+from whylogs.experimental.constraints_generation import generate_constraints_from_reference_profile
+from whylogs.core.relations import Predicate
+from whylogs.core.metrics.condition_count_metric import (
+    Condition,
+    ConditionCountConfig,
+    ConditionCountMetric,
+)
+from whylogs.core.resolvers import (
+    STANDARD_RESOLVER,
+    MetricSpec,
+    ResolverSpec,
+)
+from whylogs.core.schema import DeclarativeSchema
+
+
+@pytest.fixture
+def reference_profile_view():
+    data = {
+        "animal": ["cat", "hawk", "snake", "cat", "mosquito"],
+        "legs": [4, 2, 0, 4, 6],
+        "weight": [4.3, 1.8, 1.3, 4.1, 5.5e-6],
+    }
+    df = pd.DataFrame(data)
+
+    def even(x) -> bool:
+        return x % 2 == 0
+
+    regex_conditions = {"legs_even": Condition(Predicate().is_(even))}
+
+    legs_spec = ResolverSpec(
+        column_name="legs",
+        metrics=[
+            MetricSpec(
+                ConditionCountMetric,
+                ConditionCountConfig(conditions=regex_conditions),
+            ),
+        ],
+    )
+    schema = DeclarativeSchema(STANDARD_RESOLVER)
+    schema.add_resolver(legs_spec)
+
+    results = why.log(df, schema=schema)
+    profile_view = results.view()
+    return profile_view
+
+
+def test_constraints_generation(reference_profile_view):
+    suggested_constraints = generate_constraints_from_reference_profile(reference_profile_view=reference_profile_view)
+    # since the profile is the reference itself, every condition should pass
+    builder = ConstraintsBuilder(dataset_profile_view=reference_profile_view)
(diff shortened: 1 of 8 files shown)


python/whylogs/experimental/constraints_generation/condition_counts.py
```diff
@@ -0,0 +1,24 @@
...
+    ----------
+    column_name : str
+        Name of the column
+    column_profile : ColumnProfileView
+        Profile of the column
+    """
+    constraints = []
+    condition_count_metric = column_profile.get_metric("condition_count")
+    for condition, matches in condition_count_metric.matches.items():
+        if matches.value == condition_count_metric.total.value:
```
jamie256, line 22: I would consider the case where matches.value == 0 is also a strong signal that this is condition count could be considered a constraint. e.g. "No emails" kind of thing?

python/whylogs/experimental/constraints_generation/types_metrics.py
```diff
@@ -0,0 +1,25 @@
...
+    column_name : str
+        Name of the column
+    column_profile : ColumnProfileView
+        Profile of the column
+    """
+    constraints = []
+    types_metric = column_profile.get_metric("types")
+    non_zero_empty_types = []
+    for key, component in vars(types_metric).items():
+        if component.value != 0:
```
jamie256, line 22: When the reference profile has a count of 0 for a type, assert that the production data also has 0 of these types.

That way if the data is sparse, we don't cause failure of the constraint (e.g. mixed columns containing strings, integers, fractions, but we only get 2 rows).

jamie256 (commented):
Some minor suggestions, but looks like a great start!

python/tests/experimental/constraints_generation/test_constraints_generation.py
```diff
@@ -0,0 +1,65 @@
...
+        "animal": ["cat", "hawk", "snake", "cat", "mosquito"],
+        "legs": [4, 2, 0, 4, 6],
+        "weight": [4.3, 1.8, 1.3, 4.1, 5.5e-6],
+    }
+    df = pd.DataFrame(data)
+
+    def even(x) -> bool:
+        return x % 2 == 0
+
+    regex_conditions = {"legs_even": Condition(Predicate().is_(even))}
```
richard-rogers: this is not a regex :)

python/tests/experimental/constraints_generation/test_constraints_generation.py
```diff
@@ -0,0 +1,65 @@
...
+        column_name="legs",
+        metrics=[
+            MetricSpec(
+                ConditionCountMetric,
+                ConditionCountConfig(conditions=regex_conditions),
+            ),
+        ],
+    )
+    schema = DeclarativeSchema(STANDARD_RESOLVER)
+    schema.add_resolver(legs_spec)
```
richard-rogers: You can do `schema = DeclarativeSchema(STANDARD_RESOLVER + [legs_spec])` and skip the `add_resovler()`

python/tests/experimental/constraints_generation/test_constraints_generation.py
```diff
@@ -0,0 +1,65 @@
...
+    report = constraints.generate_constraints_report()
+    assert all(result.passed == 1 for result in report)
+    assert report[0].name == "animal has no missing values"
+    assert report[1].name == "animal types count non-zero for ['string']"
+    assert report[2].name == "legs has no missing values"
+    assert report[3].name == "legs types count non-zero for ['integral']"
+    assert report[4].name == "legs meets condition legs_even"
+    assert report[5].name == "weight has no missing values"
+    assert report[6].name == "weight types count non-zero for ['fractional']"
+    print("h")
```
richard-rogers: remove print

python/tests/experimental/constraints_generation/test_constraints_generation.py
FelipeAdachi: fixed!thanks!

python/whylogs/experimental/constraints_generation/condition_counts.py
FelipeAdachi: That's a good suggestion!
Added it as an issue to do in a follow-up:
https://github.com/whylabs/whylogs/issues/1164

python/whylogs/experimental/constraints_generation/types_metrics.py
FelipeAdachi: Good suggestion!
Added an issue for it, to be done in a follow-up:
https://github.com/whylabs/whylogs/issues/1165

richard-rogers: Maybe have an optional/configurable minimum number of rows required to generate constraints?

python/tests/experimental/constraints_generation/test_constraints_generation.py
FelipeAdachi: fixed!

python/whylogs/experimental/constraints_generation/types_metrics.py
jamie256: I meant sparse data in production, like if you had a system monitoring a web service, you might get only a few calls in some time window so the system might be less noisy if we assert 0 for types we haven't seen in the training or reference data, vs assert > 0 for types we have seen.

jamie256: Follow up is fine.

jamie256 (approved):
LGTM!

PR #1166, opened 2023-03-28
Fix monthly grouping in the multi logger, add year support

naddeoa:
Apparently the first of the month is 1, not 0.

Also adding year support for demo purposes.

I added the exception case to make sure that something goes wrong if the user isn't
using types and supplies something that isn't a valid time granularity.


jamie256 (approved):
LGTM!

PR #1170, opened 2023-03-29
Fix to embeddings freq_items and handling of numpy strings

bernease:
Frequent items sketches were empty due to the performance improvement change to wrapping closest array in numpy. We should check out if there's still a numpy solution and the effect of this on performance, but this at least fixes the bug.

This turned up a few other bugs. This PR includes the following:
 * numpy subviews are now always ndarrays (there were cases where they could be pd.Series)
 * unit tests to type check all subviews
 * TypeCountersMetric includes numpy string subview now
 * CardinalityMetric includes numpy string subview now
 * ~np.ndarray and pd.Series use different degrees of freedom in their variance computations. This PR is consistent with the previous DoF behavior, but this issue warrants further consideration...~ this is dealt with in #1173 
 * FrequentItemsMetric includes numpy string subview (the original bug that prompted this PR)

### DIFF
--- a/python/tests/core/test_preprocessing.py
+++ b/python/tests/core/test_preprocessing.py
@@ -5,9 +5,9 @@ from logging import getLogger
 from typing import Any, List, Optional, Union
 
 import numpy as np
+import numpy.testing as npt
 import pandas as pd
 import pytest
-from pandas.testing import assert_series_equal
 
 from whylogs.core.preprocessing import (
     ListView,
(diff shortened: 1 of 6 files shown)


python/tests/experimental/extras/test_embedding_metric.py
```diff
@@ -75,11 +77,13 @@ def test_embedding_metric_merge_happy_case() -> None:
     metric2.columnar_update(data)
     merged = metric1.merge(metric2)
     summary = merged.to_summary_dict()
+    print(summary)
```
richard-rogers: remove before merging

bernease:
Richard convinced me that we should expand scope to fixing of NumpyView to stop taking pd.DataFrames and only np.ndarrays as contracted by our MyPy signature. I have added that potential change in this PR, but it requires lots of cleanup for test cases.

Richard was excited to do this and I will not stop him! But it is strange that tests like `test_merge_single_values_profile_mean` are giving slightly different values. Needs investigating.

python/tests/core/test_preprocessing.py
```diff
@@ -53,11 +53,36 @@ def assert_numpy_view_is_empty(view: NumpyView) -> None:
...
+            assert isinstance(subview, list)
+
+
+def assert_numpy_view_correct_types(view: ListView) -> None:
+    for subview in [view.ints, view.floats, view.strings]:
+        if subview is not None:
+            assert isinstance(subview, np.ndarray)
+
+
+def assert_pandas_view_correct_types(view: ListView) -> None:
```
jamie256, line 68: type hint here for the parameter `view` should be `PandasView`

python/tests/core/test_preprocessing.py
```diff
@@ -53,11 +53,36 @@ def assert_numpy_view_is_empty(view: NumpyView) -> None:
     assert_subview_is_empty(view.strings)
 
 
+def assert_list_view_correct_types(view: ListView) -> None:
+    for subview in [view.ints, view.floats, view.strings, view.tensors, view.objs]:
+        if subview is not None:
+            assert isinstance(subview, list)
+
+
+def assert_numpy_view_correct_types(view: ListView) -> None:
```
jamie256, line 62: type hint here for the parameter `view` should be `NumpyView`

python/whylogs/core/metrics/column_metrics.py
```diff
@@ -45,7 +45,7 @@ def columnar_update(self, data: PreprocessedColumn) -> OperationResult:
         integral = 0
         if data.numpy.ints is not None:
             integral += len(data.numpy.ints)
-        elif data.list.ints is not None:
+        if data.list.ints is not None:
```
jamie256, line 48: What scenario allows both NumpyView.ints and ListView.ints?

These should be mutually exclusive. Same comment for NumpyView.floats and ListView.floats.

jamie256 (approved):
Some minor comments, the functionality looks correct for the bug fix.

python/tests/core/test_preprocessing.py
jamie256: lets address these in follow up: https://github.com/whylabs/whylogs/issues/1176

python/tests/core/test_preprocessing.py
jamie256: lets address these in follow up: https://github.com/whylabs/whylogs/issues/1176

python/whylogs/core/metrics/column_metrics.py
jamie256: This is ok, but shouldn't occur with the normal apply() call.

PR #1172, opened 2023-03-30
Make write schedules optional on the multi logger & Add a function for getting the current state of the logger

naddeoa:
## Description

There would always be a write schedule before, which is a little awkward
when trying to use this thing in notebook demos since it might go off in
the middle of what you're doing. Now you can set it to `None` instead
and just manually trigger writes

The driving use case for this is notebook experimentation where a
scheduled write cadence doesn't make sense, and writing remotely may not
either.

### DIFF
--- a/python/whylogs/api/logger/experimental/multi_dataset_logger/multi_dataset_rolling_logger.py
+++ b/python/whylogs/api/logger/experimental/multi_dataset_logger/multi_dataset_rolling_logger.py
@@ -195,7 +195,7 @@ class MultiDatasetRollingLogger(MessageProcessor[LoggerMessage]):
     def __init__(
         self,
         aggregate_by: TimeGranularity = TimeGranularity.Hour,
-        write_schedule: Schedule = Schedule(cadence=TimeGranularity.Minute, interval=10),
+        write_schedule: Optional[Schedule] = Schedule(cadence=TimeGranularity.Minute, interval=10),
         schema: Optional[DatasetSchema] = None,
         writers: List[Writer] = [],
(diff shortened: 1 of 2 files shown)


python/whylogs/api/logger/experimental/multi_dataset_logger/multi_dataset_rolling_logger.py
```diff
@@ -209,13 +224,14 @@ def __init__(
...
+            if write_schedule.cadence == TimeGranularity.Second:
+                raise Exception("Minimum write schedule is five minutes.")
 
-        if write_schedule.cadence == TimeGranularity.Minute and write_schedule.interval < 5:
-            raise Exception("Minimum write schedule is five minutes.")
+            if write_schedule.cadence == TimeGranularity.Minute and write_schedule.interval < 5:
+                raise Exception("Minimum write schedule is five minutes.")
 
-        self._timer = FunctionTimer(write_schedule, self.flush)
+            self._timer = FunctionTimer(write_schedule, self.flush)
```
jamie256, line 234: consider adding a log statement for when there is no schedule? It could be logged as a warning if you aren't in a notebook environment.

jamie256 (approved):
LGTM!

python/whylogs/api/logger/experimental/multi_dataset_logger/multi_dataset_rolling_logger.py
jamie256: Can do in follow up, not blocking experimental module changes here.

PR #1173, opened 2023-03-31
DistributionMetric m2 update fix

jamie256:
## Description

The default value of ddof is different between numpy and pandas arrays but we use the returned variance in a single way. This leads to an incorrect update step for the M2 variable which requires knowing if the variance was normalized with N or N-1.

## Changes
- specify ddof parameter in the var() method, always use 1.
- fix the update steps for M2 to be aware of the ddof value assumed.
- statistics.pvariance also uses N-1 normalization and had the wrong M2 update, so fixed there
- add test to cover numpy, pandas, list input on toy data.

### DIFF
--- a/python/tests/core/metrics/test_metrics.py
+++ b/python/tests/core/metrics/test_metrics.py
@@ -1,4 +1,5 @@
 from logging import getLogger
+
 import numpy as np
 import pandas as pd
 import pytest
@@ -14,6 +15,7 @@ from whylogs.core.metrics.metrics import (
 from whylogs.core.preprocessing import PreprocessedColumn
 
 TEST_LOGGER = getLogger(__name__)
+
 
 def test_distribution_metrics_numpy() -> None:
     dist = DistributionMetric.zero(MetricConfig())
@@ -43,19 +45,21 @@ def test_distribution_metrics_series() -> None:
     assert dist.mean.value == data.mean()
     assert dist.variance == data.var()
 
+
 def test_distribution_variance_m2() -> None:
     import statistics
+
     dist_list = DistributionMetric.zero(MetricConfig())
     dist_pandas = DistributionMetric.zero(MetricConfig())
     dist_numpy = DistributionMetric.zero(MetricConfig())
     test_input = [1, 2, 3, 4]
-    
+
     list_test_input = PreprocessedColumn()
     list_test_input.list.ints = test_input
     n = len(test_input)
     mean = sum(test_input) / n
     variance = statistics.variance(test_input)
-    m2 = (n-1) * variance
+    m2 = (n - 1) * variance
     TEST_LOGGER.info(f"statistic package using input {test_input} has variance={variance}, m2={m2}, n={n}")
     pandas_test_input = PreprocessedColumn.apply(pd.Series(test_input))
     numpy_test_input = PreprocessedColumn.apply(np.array(test_input))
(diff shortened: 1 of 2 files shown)


python/tests/core/metrics/test_metrics.py
```diff
@@ -42,6 +46,41 @@ def test_distribution_metrics_series() -> None:
...
+    dist_list = DistributionMetric.zero(MetricConfig())
+    dist_pandas = DistributionMetric.zero(MetricConfig())
+    dist_numpy = DistributionMetric.zero(MetricConfig())
+    test_input = [1, 2, 3, 4]
+
+    list_test_input = PreprocessedColumn()
+    list_test_input.list.ints = test_input
+    n = len(test_input)
+    mean = sum(test_input) / n
+    variance = statistics.variance(test_input)
```
richard-rogers: should comment the DoF used here just for clarity

python/whylogs/core/metrics/metrics.py
```diff
@@ -291,7 +291,7 @@ def columnar_update(self, view: PreprocessedColumn) -> OperationResult:
                 n_b = len(lst)
                 if n_b > 1:
                     mean_b = statistics.mean(lst)
-                    m2_b = statistics.pvariance(lst) * (n_b - 1)
+                    m2_b = statistics.pvariance(lst) * (n_b)
```
richard-rogers: why call `pvariance()` instead of `variance`?

jamie256: I thought changing the function called is a larger risk change, and either works as long as we have the appropriate normalization

python/whylogs/core/metrics/metrics.py
```diff
@@ -273,7 +273,7 @@ def columnar_update(self, view: PreprocessedColumn) -> OperationResult:
                     if n_b > 1:
                         n_b = len(arr)
                         mean_b = arr.mean()
-                        m2_b = arr.var() * (n_b - 1)
+                        m2_b = arr.var(ddof=1) * (n_b - 1)
```
richard-rogers, line 276: It looks like v0 always did the sample variance (1 degree of freedom). So I take it we're declaring the v1 using population variance sometimes the bug and changing to sample variances in all cases despite the minor difference from previously computed profiles?

That works for me. I like the tests here, so I vote we merge this one and I'll rebase @bernease's PR on this one.

python/whylogs/core/metrics/metrics.py
richard-rogers: Seems better to call the function that actually computes the one we want :)

richard-rogers:
Note that with the preprocessing fix, there will never be pandas series as input. Just ndarray and list.

python/tests/core/metrics/test_metrics.py
jamie256: The next line shows the normalization in the m2 assignment, and the statistics.variance doesn't have a DoF parameter, but I added a comment pointing out that this is a sample variance calculated in statistics module https://docs.python.org/3/library/statistics.html#statistics.variance

python/whylogs/core/metrics/metrics.py
jamie256: It is only used in the update of m2 here, so what we want is something consistent with the update, but sure, lets keep it consistent with using sample variance. Updated.

richard-rogers (dismissed):

python/whylogs/core/metrics/metrics.py
jamie256: correct, and also considering our bug of letting pandas series be stored in the NumpyView was a most common code path, the effective ddof was 1 when logging with pandas dataframes (which also covers pyspark).

richard-rogers (dismissed):

richard-rogers (approved):

PR #1178, opened 2023-03-31
Add warning in multi_dataset_logger when there is no write schedule

naddeoa:
Just a small warning in case people mistakenly disable write schedules.


python/whylogs/api/logger/experimental/multi_dataset_logger/multi_dataset_rolling_logger.py
```diff
@@ -232,6 +232,11 @@ def __init__(
                 raise Exception("Minimum write schedule is five minutes.")
 
             self._timer = FunctionTimer(write_schedule, self.flush)
+        else:
+            self._logger.warning(
+                f"No write schedule defined for logger. Profiles will only be written after calls to flush()."
```
jamie256: don't need the f-string prefix without parameters

jamie256 (commented):
Looks fine.

to check all the pre-commit hooks locally try running
`make pre-commit`

in `python` directory

jamie256 (approved):

PR #1180, opened 2023-04-03
Add whylogs SessionManager

murilommen:
## Description

A `SessionManager` to make it easier to manage credentials to authenticate with WhyLabs

## Changes

- Adding a `SessionManager` as a singleton for the `init` method
- Added `UserSession` and `GuestSession` as containers for each auth flow, as well as helper methods to them


python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,113 @@
...
+    CreateSessionResponse,
+    SessionsApi,
+)
+from whylabs_client.api_client import ApiClient, Configuration
+
+from .auth_file import get_auth_file_path, get_configparser_object
+from .variables import Variables
+
+_auth_path = get_auth_file_path(auth_path=Path(f"{Path.home()}/.whylabs/auth.ini"))
+DEFAULT_WHYLABS_HOST = "https://songbird.development.whylabsdev.com"
```
naddeoa: Just a temporary dev thing?

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,113 @@
...
+class GuestSession:
+    def __init__(self, session_id: str):
+        self.session_id = session_id
+        Variables.set_variable_to_config_file(key="session_id", value=self.session_id, auth_path=_auth_path)
+
+
+@dataclass
+class UserSession:
+    org_id: str
+    dataset_id: str
```
naddeoa: Will this be a thing? I kind of assumed that dataset would continue to live in the api latyer rather than have a single value since it's far more likely to change than the org/api key

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,113 @@
...
+    api_key: str
+
+    def __post_init__(self):
+        Variables.set_variable_to_config_file(key="org_id", value=self.org_id, auth_path=_auth_path)
+        Variables.set_variable_to_config_file(key="dataset_id", value=self.dataset_id, auth_path=_auth_path)
+        keyring.set_password("whylabs", "api_key", self.api_key)
```
naddeoa: Does this end up prompting them for a password when we call `get`? And is this the right spot to put some of the `if_notebook` checks? I assume we'll have to do that for some of the UX for entering settings using `input()`.

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,113 @@
...
+
+def _create_session_id(user_id: str) -> str:
+    config = Configuration()
+    config.host = os.getenv("WHYLABS_HOST") or DEFAULT_WHYLABS_HOST
+
+    client = ApiClient(config)
+    api = SessionsApi(client)
+    response: CreateSessionResponse = api.create_session(CreateSessionRequest(user_id))
+
+    print(f"Created session {response.id}")
```
naddeoa: Do we use `logging` for this stuff?

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,113 @@
...
+    session_id = Variables.get_variable_from_config_file(config=config, key="session_id")
+    if session_id is None:
+        user_id = str(uuid.uuid4())
+        session_id = _create_session_id(user_id=user_id)
+    return GuestSession(session_id=session_id)
+
+
+def create_session(anonymous: Optional[str] = None) -> Union[GuestSession, UserSession]:
+    if anonymous is None:
+        anonymous = input("Do you want to create an anonymous session? [True/False] (default: False) ")
```
naddeoa: ah this. But I think we probably want gate it in that `if_notebook` check in case they run this in a normal file after. Not even sure what would happen.

naddeoa: Especially if we have a while True loop lol

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,113 @@
...
+        if SessionManager.__instance is not None:
+            raise Exception("There is an active Session, use Session.get_instance() instead")
+        else:
+            SessionManager.__instance = self
+            self.session = create_session(anonymous=self._anonymous)
+
+    @staticmethod
+    def get_instance(anonymous: Optional[str] = None):
+        if SessionManager.__instance is None:
+            SessionManager(anonymous=anonymous)
```
naddeoa, line 115: Doesn't this have to be set to `Sessionmanager.__instance`?

naddeoa (commented):
Sanity check before we send all of this code. Is this basically doing what pydantic's BaseSettings class does automatically?

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,113 @@
+import os
+import uuid
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Optional, Union
+
+import keyring
```
jamie256: for new run time dependencies (such as keyring) please do the following:
1. `poetry add keyring`
2. `poetry update`

Also test in a non-dev environment, e.g. after you build a wheel off your branch you can drag it onto colab as a local file, then install that wheel and try to use it, minimally run:
`import whylogs`

The problem with this change is that this dependency is already a dev dependency (pulled in by twine), so things will work locally for whylogs developers poetry env, but keyring is missing in the pyproject.toml file's install time depedencies, so new builds would break whylogs import if the env doesn't already have the required packages.

python/whylogs/api/whylabs/session_manager.py
murilommen: better to! will change that

python/whylogs/api/whylabs/session_manager.py
murilommen: is is set on `SessionManager`'s init

python/whylogs/api/whylabs/session_manager.py
murilommen: right! i'm not 100% confident in adding keyring just yet. will test it on other environments (ubuntu, wsl, windows) shortly

python/tests/api/whylabs/test_session_manager.py
```diff
@@ -0,0 +1,109 @@
...
+    create_session,
+)
+
+
+class TestSessionManager:
+    @classmethod
+    def setup_class(cls):
+        os.environ["WHYLABS_API_KEY"] = "api_key"
+        os.environ["ORG_ID"] = "org_id"
+        os.environ["WHYLABS_API_ENDPOINT"] = "https://songbird.development.whylabsdev.com"
```
andyndang: should leave our dev endpoint here

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,115 @@
...
+    def __post_init__(self):
+        Variables.set_variable_to_config_file(key="org_id", value=self.org_id, auth_path=_auth_path)
+        Variables.set_variable_to_config_file(key="api_key", value=self.api_key, auth_path=_auth_path)
+
+
+def _create_session_id(user_id: str) -> str:
+    config = Configuration()
+    config.host = os.getenv("WHYLABS_HOST") or DEFAULT_WHYLABS_HOST
+
+    client = ApiClient(config)
```
andyndang: don't create the API client every time. It's expensive - should reuse the connection pool in general

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,115 @@
...
+
+def _get_guest_session() -> GuestSession:
+    session_id = Variables.get_variable_from_config_file(auth_path=_auth_path, key="session_id")
+    if session_id is None:
+        user_id = str(uuid.uuid4())
+        session_id = _create_session_id(user_id=user_id)
+    return GuestSession(session_id=session_id)
+
+
+def create_session(anonymous: Optional[bool] = None, interactive: bool = False) -> Union[GuestSession, UserSession]:
```
andyndang: would this fail gracefully if they can't connect to our endpoint?

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,115 @@
...
+
+    @staticmethod
+    def is_active() -> bool:
+        if SessionManager.__instance is None:
+            return False
+        return True
+
+
+def init(anonymous: Optional[bool] = None, interactive: bool = False) -> None:
+    SessionManager.get_instance(anonymous=anonymous, interactive=interactive)
```
andyndang: Do we always create new sessions if `anonymous` flag is enabled? We're not pulling config from the file regardless?

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,115 @@
...
+        user_id = str(uuid.uuid4())
+        session_id = _create_session_id(user_id=user_id)
+    return GuestSession(session_id=session_id)
+
+
+def create_session(anonymous: Optional[bool] = None, interactive: bool = False) -> Union[GuestSession, UserSession]:
+    if not anonymous:
+        return _get_logged_session(interactive=interactive)
+    else:
+        return _get_guest_session()
```
andyndang: We should enable re-using anonymous sessions

python/whylogs/api/whylabs/variables.py
```diff
@@ -0,0 +1,26 @@
+from pathlib import Path
+from typing import Optional
+
+from .auth_file import get_configparser_object
+
+
+class Variables:
+    @staticmethod
+    def get_variable_from_input(variable_name: str) -> str:
+        variable = input(f"What is your {variable_name}? ")
```
andyndang, line 11: This isn't good. For API key you should use `getPass` or it'll get saved in the notebook https://docs.python.org/3/library/getpass.html

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,115 @@
...
+from whylabs_client.api_client import ApiClient, Configuration
+
+from whylogs.api.whylabs.auth_file import get_auth_file_path
+from whylogs.api.whylabs.notebook_check import is_notebook
+from whylogs.api.whylabs.variables import Variables
+
+DEFAULT_PATH = os.getenv("WHYLOGS_CONFIG_PATH") or f"{Path.home()}/.whylabs/auth.ini"
+_auth_path = get_auth_file_path(auth_path=Path(DEFAULT_PATH))
+
+DEFAULT_WHYLABS_HOST = "https://api.whylabsapp.com"
```
jamie256, line 20: consider:
```suggestion
DEFAULT_WHYLABS_HOST = os.environ.get("WHYLABS_API_ENDPOINT") or "https://api.whylabsapp.com"
```

Consider moding this into a common file so we can use same definition for session manager and the whylabs writer.

python/examples/integrations/Fugue_Profiling.ipynb
```diff
@@ -1071,7 +1071,7 @@
     }
    ],
    "source": [
-    "!ls /tmp/output*.parquet"
+    "# !ls /tmp/output*.parquet"
```
jamie256: ?

jamie256 (commented):
Why do we need to modify the fugue example notebook here

naddeoa (dismissed):

python/whylogs/api/whylabs/session_manager.py
murilommen: but that's what the singleton `SessionManager` is used for. if there is an active session (either logged or guest), it will refer to it instead of creating a new one

python/whylogs/api/whylabs/session_manager.py
murilommen: no, if the anonymous flag is enabled, it will first try to check if there is a session_id on the default path, otherwise it will create it using the API

python/whylogs/api/whylabs/session_manager.py
murilommen: this is only created and called once - when users want to create an anonymous session AND they don't have it on the config file yet

python/whylogs/api/whylabs/session_manager.py
murilommen: ok! makes sense

python/examples/integrations/Fugue_Profiling.ipynb
murilommen: this is the command that made my changes fail. i have no clue why and I couldn't repro this locally. on my branch the tests pass. 
apparently it's something that only happens on Ubuntu environments? since Anthony could narrow it down to this line of code. 
i feel awkward to make this change, but can't find a way to make this work without it.

python/whylogs/api/whylabs/session_manager.py
murilommen: now it will! :D thanks for the comment

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,120 @@
...
+    __instance = None
+
+    def __init__(self, anonymous: Optional[bool] = None, interactive: bool = False):
+        self._anonymous = anonymous
+        self._interactive = interactive
+        if SessionManager.__instance is not None:
+            raise Exception("There is an active Session, use Session.get_instance() instead")
+        else:
+            SessionManager.__instance = self
+            self.session = create_session(anonymous=self._anonymous, interactive=self._interactive)
```
andyndang: so you are always creating a new session if anonymous is true? But that means users will lose their session context if they terminate the notebook. We should store the session ID in the config and let user re-use them in the future in this case rather than creating new session every time

python/whylogs/api/whylabs/session_manager.py
andyndang: I don't think that's the logic here. Looks like it only re-using the session in the same process. but if the process terminates then the session is forever gone

python/whylogs/api/whylabs/session_manager.py
murilommen: but that's what `create_session` does for a guest session. on line 79:
```python
def _get_guest_session() -> GuestSession:
    session_id = Variables.get_variable_from_config_file(auth_path=_auth_path, key="session_id")
    if session_id is None:
        (...)
```

if you have any other suggestions on how i should do it, let me know

python/whylogs/api/whylabs/session_manager.py
murilommen: that's true. but when starting the process over, the first thing will be checking the config file for an existing `session_id`. let's keep it on the same thread, but please let me know if this should be handled differently

python/whylogs/api/whylabs/session_manager.py
murilommen: I've just renamed the method for clarification on this behavior

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,122 @@
...
+        user_id = str(uuid.uuid4())
+        session_id = _create_session_id(user_id=user_id)
+    return GuestSession(session_id=session_id)
+
+
+def get_or_create_session(
+    anonymous: Optional[bool] = None, interactive: bool = False
+) -> Union[GuestSession, UserSession]:
+    if not anonymous:
+        return _get_logged_session(interactive=interactive)
```
andyndang: `get_or_create`?

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,122 @@
...
+
+DEFAULT_WHYLABS_HOST = "https://api.whylabsapp.com"
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+class GuestSession:
+    def __init__(self, session_id: str):
+        self.session_id = session_id
+        Variables.set_variable_to_config_file(key="session_id", value=self.session_id, auth_path=_auth_path)
```
andyndang: The flow for this saving logic is really strange here. You should save it only at user input prompt, not when creating an object like this. By doing this at the object init level you are generating side effects - what if somehow we need to create these sessions manually for whatever reason? But we don't want to overwrite the existing configs?

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,122 @@
...
+
+def _get_logged_session(auth_path: Path = _auth_path, interactive: bool = False) -> UserSession:
+    api_key = os.getenv("WHYLABS_API_KEY") or Variables.get_variable_from_config_file(
+        auth_path=auth_path, key="api_key"
+    )
+    org_id = os.getenv("ORG_ID") or Variables.get_variable_from_config_file(auth_path=auth_path, key="org_id")
+
+    if is_notebook() and interactive is True:
+        api_key = api_key or Variables.get_variable_from_getpass(variable_name="api_key")
+        org_id = org_id or Variables.get_variable_from_input(variable_name="org_id")
```
andyndang: You should only save the variables at this point

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,122 @@
...
+
+    return UserSession(org_id=org_id, api_key=api_key)
+
+
+def _get_or_create_guest_session() -> GuestSession:
+    session_id = Variables.get_variable_from_config_file(auth_path=_auth_path, key="session_id")
+    if session_id is None:
+        user_id = str(uuid.uuid4())
+        session_id = _create_session_id(user_id=user_id)
+    return GuestSession(session_id=session_id)
```
andyndang, line 91: or at this point (see the comment above about not creating side effects in a constructor

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,122 @@
...
+    try:
+        client = ApiClient(config)
+        api = SessionsApi(client)
+        response: CreateSessionResponse = api.create_session(CreateSessionRequest(user_id))
+
+        logger.debug(f"Created session {response.id}")
+        return response.id
+    except ApiException as e:
+        logger.error(e)
+        raise e
```
andyndang, line 55: This is bad. Nothing is catching the error above. 

We should always fall back to local mode if this call cannot be made and just log out a warning

python/whylogs/api/whylabs/session_manager.py
murilommen: in the case of a logged session it will always be "get", since it basically tries to get it from either an env var, config file or input (if in interactive mode)

python/whylogs/api/whylabs/session_manager.py
murilommen: ok, I'll move it out of the init instead!

python/whylogs/api/whylabs/session_manager.py
murilommen: What does local mode mean for the anonymous login? this API is only called when creating a GuestSession.
also, if not with ApiException, what else could I use to catch errors here? I'm using the same pattern indicated on other songbird calls across our examples

andyndang: Got it. I thought this was getting called everytime

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,119 @@
...
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class GuestSession:
+    session_id: str
+
+
+@dataclass
+class UserSession:
```
andyndang: Would rename this to `ApiKeySession` to make it explicit since there are other ways a user can auth with a service

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,119 @@
...
+
+class SessionManager:
+    __instance = None
+
+    def __init__(self, anonymous: Optional[bool] = None):
+        self._anonymous = anonymous
+        if SessionManager.__instance is not None:
+            raise Exception("There is an active Session, use Session.get_instance() instead")
+        else:
+            SessionManager.__instance = self
```
andyndang: this should be after `get_or_create_session` because that method can throw exception

python/whylogs/api/whylabs/auth_file.py
```diff
@@ -0,0 +1,27 @@
...
+    try:
+        auth_file_path = auth_path.resolve(strict=True)
+    except FileNotFoundError:
+        auth_file_path = _create_blank_auth_file(auth_path=auth_path)
+    return auth_file_path
+
+
+def get_configparser_object(auth_path: Path) -> configparser.ConfigParser:
+    config = configparser.ConfigParser()
+    config.read(auth_path)
```
andyndang: Does this throw exception if the file is corrupted or missing?

python/whylogs/api/whylabs/auth_file.py
```diff
@@ -0,0 +1,27 @@
+import configparser
+from pathlib import Path
+
+
+def _create_blank_auth_file(auth_path: Path) -> Path:
```
andyndang, line 5: what happens if this method fails due to file system permission? No home folder?

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,119 @@
...
+    SessionsApi,
+)
+from whylabs_client.api_client import ApiClient, Configuration
+
+from whylogs.api.whylabs.auth_file import get_auth_file_path
+from whylogs.api.whylabs.notebook_check import is_notebook
+from whylogs.api.whylabs.variables import Variables
+
+DEFAULT_PATH = os.getenv("WHYLOGS_CONFIG_PATH") or f"{Path.home()}/.whylabs/auth.ini"
+_auth_path = get_auth_file_path(auth_path=Path(DEFAULT_PATH))
```
andyndang: This is not a good practice to run code with side effect at import. 

This will break whylogs if `get_auth_file_path` fails. There are different failure modes:
- Permission issues
- The file OS could hang and block the main application

You should move this logic to JIT - only runs this check when user calls `why.init` (or whenever the `init` flow runs)

andyndang (changes requested):
More comment around init/error handliing

python/whylogs/api/whylabs/auth_file.py
```diff
@@ -0,0 +1,39 @@
...
+    return auth_file_path
+
+
+def get_configparser_object(auth_path: Path) -> configparser.ConfigParser:
+    try:
+        config = configparser.ConfigParser()
+        config.read(auth_path)
+        return config
+    except (configparser.Error, OSError, IOError, PermissionError) as e:
+        raise ValueError(f"Error reading config file from {auth_path}: {str(e)}") from e
```
andyndang, line 39: Shouldn't you log a warning and return an empty object instead?

python/whylogs/api/whylabs/auth_file.py
```diff
@@ -0,0 +1,39 @@
...
+        config.read(auth_path)
+        config["whylabs"] = {}
+
+        # writes config file to disk
+        with open(auth_path, "w") as configfile:
+            config.write(configfile)
+        return auth_path
+
+    except (configparser.Error, OSError, IOError, PermissionError) as e:
+        raise ValueError(f"Error creating auth file at {auth_path}: {str(e)}") from e
```
andyndang, line 22: Shouldn't you warn instead of raise an error here? Otherwise the user can't fall back to the env variables

andyndang (changes requested):
Just want to make sure we are on the same page here - I want to make sure that when user does:

```
why.init(...)
```

they will not be BLOCKED from using whylogs because of initialization errors due to invalid configs/network errors/file system errors. They might not be able to write to whylabs or use anonymous session, but they should never be blocked from using the main functionalities

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,126 @@
...
+        logger.debug(f"Created session {response.id}")
+        return response.id
+    except ApiException as e:
+        logger.error(e)
+        raise e
+
+
+def _get_logged_session(auth_path: Optional[Path] = None) -> ApiKeySession:
+    api_key = os.getenv("WHYLABS_API_KEY")
+    org_id = os.getenv("ORG_ID")
```
andyndang, line 60: this should be prefixed with "WHYLABS_"

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,126 @@
...
+
+    if api_key is None or org_id is None:
+        _auth_path = _get_default_authentication_path()
+        api_key = api_key or Variables.get_variable_from_config_file(auth_path=auth_path, key="api_key")
+        org_id = org_id or Variables.get_variable_from_config_file(auth_path=auth_path, key="org_id")
+
+        if is_notebook():
+            if api_key is None:
+                api_key = Variables.get_variable_from_getpass(variable_name="api_key")
```
andyndang, line 69: you should capture these strings as constrants. It's really easy to have a typo

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,126 @@
...
+                Variables.set_variable_to_config_file(key="api_key", value=api_key, auth_path=_auth_path)
+            if org_id is None:
+                org_id = Variables.get_variable_from_input(variable_name="org_id")
+                Variables.set_variable_to_config_file(key="org_id", value=org_id, auth_path=_auth_path)
+
+    if api_key is None or org_id is None:
+        raise ValueError(
```
andyndang, line 76: I just realized that there's no option for users to opt out of WhyLabs. We would want to push `why.init()` to be everywhere so we would need an escape path for local-only users. Not a blocker for this PR but we should track this as a missing feature for this to be GA

python/whylogs/api/whylabs/session_manager.py
```diff
@@ -0,0 +1,126 @@
...
+    if api_key is None or org_id is None:
+        _auth_path = _get_default_authentication_path()
+        api_key = api_key or Variables.get_variable_from_config_file(auth_path=auth_path, key="api_key")
+        org_id = org_id or Variables.get_variable_from_config_file(auth_path=auth_path, key="org_id")
+
+        if is_notebook():
+            if api_key is None:
+                api_key = Variables.get_variable_from_getpass(variable_name="api_key")
+                Variables.set_variable_to_config_file(key="api_key", value=api_key, auth_path=_auth_path)
+            if org_id is None:
```
andyndang, line 71: It'll be weird to use an API key from a config file and an org ID from an environment variable. It should be all or nothing.

It's worth printing out where you're pulling the variables from as well, if not in the debug code. It's important for debugging

andyndang (commented):
Overall it looks better now but I'll leave the main approval to Jamie

murilommen:
Closing this, as it's going to be handled by #1267

PR #1190, opened 2023-04-06
Add out-of-vocabulary counting to BoW metric

richard-rogers:
## Description

Adds optional out-of-vocabulary counting to `BagOfWordsMetric`.

## Changes

```
vocab = ["a", "b", "c"]
config = BagOfWordsConfig.set_vocabulary(vocab)
resolvers = [
    ResolverSpec(column_name="document", metrics=[MetricSpec(BagOfWordsMetric, config)])
]
schema = DeclarativeSchema(resolvers)
doc = ["a", "b", "c", "1", "2", "a", "c", "2"]
profile = why.log(row={"document": doc}, schema=schema)
```

The vocabulary Bloom filter is saved at `config.filename`. It can be deserialized with `BagOfWordsConfig.load_vocabulary()`. If a vocabulary is set, the `BagOfWordsMetric` tracks the counts of OoV terms encountered in the logged documents. If no vocabulary is set, the OoV terms are neither counted nor tracked.


python/whylogs/experimental/extras/nlp_metric.py
```diff
@@ -1,8 +1,11 @@
 from copy import deepcopy
 from dataclasses import dataclass, field
 from itertools import chain
+from tempfile import NamedTemporaryFile
 from typing import Any, Dict, List, Optional, Tuple, Union
 
+from pybloomfilter import BloomFilter
```
FelipeAdachi, line 7: We're now requiring the `nlp` extra to be installed to anyone who wants to use BagofWords, even if not interested in OOV functionalities (as is the case with some of our previous examples). Just raising the question if we think this is ok, or would we like to import this only when actually using OOV features?

richard-rogers: Right now, the BoW and SVD metrics are in the same file. SVD requires numpy. I could split BoW into a separate file/extra, but the I don't Bloom filter package is an onerous dependency.

python/whylogs/experimental/extras/nlp_metric.py
```diff
@@ -195,13 +200,47 @@ def _all_strings(value: List[Any]) -> bool:
...
+        filename = filename or _get_temp_file_name()
+        vocabulary = BloomFilter(len(vocab), error_rate, filename)
+        for v in vocab:
+            vocabulary.add(v)
+        vocabulary.sync()
+        return BagOfWordsConfig(filename=filename, vocabulary=vocabulary)
+
+    @classmethod
+    def load_vocabulary(cls, filename: str) -> "BagOfWordsConfig":
+        vocabulary = BloomFilter.open(filename, "r")
```
richard-rogers, line 232: maybe writable backing file?

python/whylogs/experimental/extras/nlp_metric.py
```diff
@@ -195,13 +200,47 @@ def _all_strings(value: List[Any]) -> bool:
...
+@dataclass(frozen=True)
+class BagOfWordsConfig(MetricConfig):
+    filename: Optional[str] = None
+    vocabulary: Optional[BloomFilter] = None
+
+    @classmethod
+    def set_vocabulary(
+        cls, vocab: List[str], error_rate: float = 0.00001, filename: Optional[str] = None
+    ) -> "BagOfWordsConfig":
+        filename = filename or _get_temp_file_name()
```
richard-rogers, line 223: This will leak files :(

python/whylogs/experimental/extras/nlp_metric.py
```diff
@@ -195,13 +200,47 @@ def _all_strings(value: List[Any]) -> bool:
...
+
+
 @dataclass
 class BagOfWordsMetric(MultiMetric):
     """
     Natural language processing metric -- treat document as a bag of words
     """
 
     fi_disabled: bool = False
+    vocabulary: Optional[BloomFilter] = None
```
richard-rogers, line 243: Might need a a Bloom filter metric component...

python/pyproject.toml
```diff
@@ -68,6 +68,9 @@ scikit-learn = [
 fugue = {version = "^0.8.1", optional = true}
 requests = "^2.27"
 
+# NLP related dependencies
+pybloomfiltermmap3 = {version = "^0.5.7", python = ">=3.5", optional = true}
```
jamie256, line 72: While this is an optional nlp specific dependency, I have concerns about pulling this package in, are there alternatives? This would require more testing on different platforms / os / ISA before we could pull this in as I'm not sure this works on the platforms, os, and ISA matrix that whylogs supports.

jamie256 (changes requested):
Need to fix dependency issue before we can do another round of review.

PR #1191, opened 2023-04-07
add whylogs[all] install extra and update poetry.lock

jamie256:
## Description

There are enough whylogs extras as this point that it is clunky to list them all for install when testing or exploring whylogs in a notebook environment. Adding this target allows a new user to avoid import errors related to missing dependencies.

incidental mypy error fix for pre-commit failure newly detected.


bernease (approved):
LGTM

PR #1192, opened 2023-04-07
Single value line for double_histogram viz

jamie256:
## Description

When there is only one bin with a single value the histogram chart and the reference profile has counts larger than 10, the single value won't be visible. This add a dashed line to show where on the x axis the single value falls. This can be useful when comparing a single row's feature when comparing to a reference profile for that feature.

With this change, here is what the `double_histogram` chart looks like. Without this change, the target profile has a single value that is too small to really see in this chart.

double histogram single value treatment


FelipeAdachi (approved):
LGTM!

Issue #1194, opened 2023-04-10
Template generation should not try to open with write-mode


aq-ashwin:
## Description

Installing whylogs in a read-only file system causes some issues with how the templating code is currently read and compiled.

- System info
  - Operating system and version: Ubuntu 20.04
  - Browser and version: Chrome
- Steps to reproduce:
  - Install whylogs into a virtualenv and make the site-packages read-only.
  - Attempt to create profile_summary with NotebookProfileVisualizer

![image](https://user-images.githubusercontent.com/129331039/230988805-d69c58e2-65cb-4bd0-b222-12c81d17bb5d.png)

## Suggestions

Open template file in read-only mode, i.e. "rt" instead of "r+t". The file is only ever read.
https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/viz/utils/html_template_utils.py#L25

## Related

- [x] I have reviewed the [Guidelines for Contributing](https://github.com/whylabs/whylogs/blob/mainline/.github/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/whylabs/whylogs/blob/mainline/.github/CODE_OF_CONDUCT.md).

PR #1195, opened 2023-04-10
Update to open template file in read-only mode.

aq-ashwin:
Based on https://github.com/whylabs/whylogs/issues/1194.


jamie256 (approved):
Thanks for the contribution Ashwin!

You are correct, there is no need to open this file with r+ since the viz module is only reading it.

PR #1196, opened 2023-04-10
Return whylabs profile id on upload

jamie256:
## Description

When uploading to WhyLabs we get back a returned reference id for the uploaded profile, but whylogs ignores it. This change fixes the result set writer to return back a list of the responses, and updates the whylabs writer to return the reference id when upload is successful in the existing response Tuple(bool, str) for the str message.

For the success case, this will replace the string 'OK' for the whylabs writer's returned status message, and instead you would get something like:

```
from whylogs.api.writer.whylabs import WhyLabsWriter

writer = WhyLabsWriter()
status = writer.write(results.view())
print(status)
```
outputs:
```
(True, 'log-rlnMCbmflpBI38Hc')
```

### DIFF
--- a/python/whylogs/api/logger/result_set.py
+++ b/python/whylogs/api/logger/result_set.py
@@ -126,17 +126,20 @@ class ResultSetWriter:
         self._writer.option(**kwargs)
         return self
 
-    def write(self, **kwargs: Any) -> None:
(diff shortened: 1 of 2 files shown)


FelipeAdachi (approved):
LGTM

PR #1197, opened 2023-04-11
Experimental batch ranking performance metrics

bernease:
Experimental ranking performance metrics method that logs performance metrics as individual columns. This method is a "batch" method in that it is not meant for merging -- not recommended for Spark, distributed systems, or submitting multiple profiles in the same batch time window. The API follows the pattern of `log_classification_metrics` and `log_regression_metrics` except for the swap of target_column and prediction_column (which is unfortunate).

Follow on work for a fully mergeable ranking performance metric method will follow with model performance metrics V1 design work and additional metrics needed for common merging patterns.


richard-rogers (approved):

python/examples/experimental/custom_rank_metrics.ipynb
```diff
@@ -0,0 +1,242 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!pip install --upgrade \"whylogs[whylabs]==1.1.36-dev0\""
```
jamie256, line 9: Need to update this, maybe put 1.2.2 or later?

python/examples/experimental/custom_rank_metrics.ipynb
```diff
@@ -0,0 +1,242 @@
...
+    "column_names = [\"mean_average_precision_k_\"+str(k), \"accuracy_k_\"+str(k), \n",
+    "                \"mean_reciprocal_rank\", \"precision_k_\"+str(k), \"recall_k_\"+str(k), \n",
+    "                \"top_rank\", \"average_precision_k_\"+str(k)]\n",
+    "\n",
+    "whylabs_writer = WhyLabsWriter()\n",
+    "try:\n",
+    "    whylabs_writer.tag_output_columns(column_names)\n",
+    "except:\n",
+    "    print(\"Issue with changing UI via whylogs. \\nUse REST API at \"\n",
+    "          \"https://api.whylabsapp.com/swagger-ui#/Models/PutEntitySchemaColumn.\")"
```
jamie256, line 179: I don't think we should guide people to the REST API directly, can we fix the tag_output_columns if there is an error?

PR #1201, opened 2023-04-12
Fix whylabs config read error handling

jamie256:
## Description

If there is an error reading the server side configuration in WhyLabs the default value should fallback to an older back-compat format when uploading profiles to WhyLabs. Also whylogs reads WhyLabs config on each upload, which if there is a failure results in noisy and slowed down uploads. 

## Changes

- Use older format as fallback when reading configuration fails.
- Only allow using new format if reading config succeeds, and the format is allowed.
- If there is an error reading whylabs config don't keep attempting to read that config and repeat this error on subsequent uploads.
- support avoiding reading whylabs config with an env variable


python/whylogs/api/writer/whylabs.py
```diff
@@ -57,9 +57,13 @@
 
 _US_WEST2_DOMAIN = "songbird-20201223060057342600000001.s3.us-west-2.amazonaws.com"
 _S3_PUBLIC_DOMAIN = os.environ.get("_WHYLABS_PRIVATE_S3_DOMAIN") or _US_WEST2_DOMAIN
+_WHYLABS_SKIP_CONFIG_READ = os.environ.get("_WHYLABS_SKIP_CONFIG_READ") or False
 
 
 def _check_whylabs_condition_count_uncompound() -> bool:
+    global _WHYLABS_SKIP_CONFIG_READ
+    if _WHYLABS_SKIP_CONFIG_READ:
```
murilommen, line 65: if I set this env var with `export _WHYLABS_SKIP_CONFIG_READ=False`, it will be parsed as a string and return True anyways.

jamie256: This is a temporary escape hatch for an edge case while we are reading whylabs side config to know if we need to transform some things to v0 formats. If you don't need to skip the whylabs config read then don't set the var.

While the behavior is surprising in your scenario of setting the var to False (also 0), I don't think this is worth the complexity of parsing and comparing the actual values.

richard-rogers (approved):

murilommen (approved):

PR #1211, opened 2023-04-18
Auto constraints expansion

### LINKED ISSUE #254: Automatic Constraints Generation from Baseline
It would be nice to have a feature for generating constraints automatically based on a set of profiles that represents a baseline, such as the training data, so that it could be applied during inference time.

For instance, the resulting constraints could be set with thresholds based on statistical properties of the baseline, such as mean or median.

FelipeAdachi:
## Description

- `has_zero_count_types` : Change constraint from asserting that counts are not zero for existing type to asserting that counts are zero for non-existing types
- `condition_never_meets`: add constraint if condition is never met in reference
-  `is_probably_unique`: checks if is probably unique based on `counts` and `cardinality`
-  `is_non_negative`
-  `is_in_range`
-  `frequent_strings_in_reference_set`

Also:
- Add Cardinality Thresolds to guide creation generations such as frequent_strings...
- Add `included_columns` and `excluded_columns` to make constraints generation more convenient

Closes https://github.com/whylabs/whylogs/issues/254

### DIFF
--- a/python/whylogs/core/utils/stats_calculations.py
+++ b/python/whylogs/core/utils/stats_calculations.py
@@ -1,8 +1,9 @@
 from typing import Optional, Tuple, Union
 
+from typing_extensions import TypedDict
+
 from whylogs.core.metrics import DistributionMetric
 from whylogs.core.view.column_profile_view import ColumnProfileView
-from typing_extensions import TypedDict
 
 
 def get_distribution_metrics(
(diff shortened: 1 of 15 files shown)


python/whylogs/core/utils/stats_calculations.py
```diff
@@ -17,3 +19,31 @@ def get_distribution_metrics(
     max_val = distribution_metric.kll.value.get_max_value()
     range_val = max_val - min_val
     return min_val, max_val, range_val
+
+
+def is_probably_unique(column_profile: ColumnProfileView) -> bool:
+    cardinality = column_profile.get_metric("cardinality").to_summary_dict()
+    counts = column_profile.get_metric("counts").to_summary_dict()
+    if "cardinality" in column_profile.get_metric_names() and "counts" in column_profile.get_metric_names():
+        if cardinality["lower_1"] <= counts["n"] - counts["null"] <= cardinality["upper_1"]:
```
jamie256, line 28: Let's follow up and increase this to a few standard deviations rather than the 1 stddev in the upper_1 and lower_1 estimates.

jamie256 (approved):
This is exciting Felipe!

There are a few issues for follow up, and lets add an example notebook (or update existing examples). But looks good as first iteration in experimental namespace!

python/whylogs/core/utils/stats_calculations.py
jamie256: https://github.com/whylabs/whylogs/issues/1216

PR #1224, opened 2023-04-27
hackathon2023 constraintshub

richard-rogers:
## Description

Constraints Hub work for Hackathon 2023

## Changes

- Versioned store for config files
- [De]serialize constraint suite (and eventually validators, etc.) to/from the config store
- Workflow/UI for editing, reviewing, & managing constraint suites
- Autogenerate constraint suite to the workflow
- Profile writer that fakes evaluating constraints in WhyLabs & reports in profile columns for monitoring


python/whylogs/experimental/api/writer/cnwriter.py
```diff
@@ -0,0 +1,22 @@
...
+from whylogs.core.utils import deprecated_alias
+from whylogs.core.view.dataset_profile_view import DatasetProfileView
+
+
+class CnWriter(WhyLabsWriter):
+    @deprecated_alias(profile="file")
+    def write(self, file: Writable, **kwargs: Any) -> Tuple[bool, str]:
+        if isinstance(file, (DatasetProfile, DatasetProfileView)):
+            # check if {org,dataset}_id overridden in kwargs
+            constraints_str = ConfigHub(self._org_id, self._dataset_id, "constraints").get_latest()
```
richard-rogers: I suppose we should check if the constraint suite is available :)

python/tests/experimental/api/constraints/test_constraints_yaml.py
```diff
@@ -0,0 +1,97 @@
...
+import pytest
+import os
+import tempfile
+
+from whylogs.experimental.api.constraints import read_constraints_from_yaml, write_constraints_to_yaml
+
+yaml_string = """\
+    
+id: "model-23"
+version: 1
```
richard-rogers, line 30: What is this the version of? If it's the constraint suite, I've used semantic versioning so it would be 1.0.0.

python/tests/experimental/api/constraints/test_constraints_yaml.py
```diff
@@ -0,0 +1,97 @@
...
+hash: abcabc231 # maybe from git?
```
richard-rogers, line 31: I'm not exposing hashes in the `ConfigStore` API, just semvers.

python/tests/experimental/api/constraints/test_constraints_yaml.py
```diff
@@ -0,0 +1,97 @@
...
+            file.write(yaml_string)
+        constraints = read_constraints_from_yaml(os.path.join(temp_dir, input_yaml_name))
+        builder = ConstraintsBuilder(dataset_profile_view=reference_profile_view)
+        builder.add_constraints(constraints)
+        rehydrated_constraints = builder.build()
+        write_constraints_to_yaml(
+            constraints=rehydrated_constraints, output_path=os.path.join(temp_dir, output_yaml_name)
+        )
+        constraints_out = read_constraints_from_yaml(os.path.join(temp_dir, output_yaml_name))
+        assert len(constraints) == len(constraints_out)
```
richard-rogers, line 124: Could check `example.yaml` == `example2.yaml` by lines or file hashes?

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -17,6 +17,13 @@
 ReportResult = namedtuple("ReportResult", "name passed failed summary")
 
 
+class ConstraintsParams(TypedDict):
```
richard-rogers, line 22: This should probably be a `dataclass`

richard-rogers, line 22: Some of these seem redundant with the `MetricConstraint` itself?

python/whylogs/experimental/api/constraints/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+
+constraints_mapping = {
+    "no_missing_values": no_missing_values,
+    "is_in_range": is_in_range,
+    "column_is_probably_unique": column_is_probably_unique,
+}
+
+
+def assemble_constraint(constraint_dict: dict) -> Optional[TypedDict]:
+    constraint_name = None
```
richard-rogers, line 73: I think we could handle prefix expression constraints with something like
```
if "expression" in constraint_dict:  # alternative to factory key
  condition = PrefixCondition(constraint_dict.get("expression"))
  ...
elif "factory" in constraint_dict:
  # as below
```

For hackathon scope, just have the UI be a text entry field where we enter prefix expressions. For production, we can make a better expression builder UI that uses prefix internally.

python/whylogs/experimental/api/constraints/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+    else:
+        logger.warning(f"No constraints found in {input_path}.")
+        return None
+
+
+def write_constraints_to_yaml(constraints: Constraints, output_path: str):
+    constraints_list = []
+    for column_name, column_constraints in constraints.column_constraints.items():
+        for constraint_name, constraint in column_constraints.items():
+            if constraint._params:
```
richard-rogers: Here check if the constraint callable is an instance of `PrefixCondition` and serialize with `- expression: constraint.condition._expression` instead of `factory: ...`

python/whylogs/experimental/extras/confighub.py
```diff
@@ -0,0 +1,130 @@
...
+    @abstractmethod
+    def get_available_versions(self) -> List[Version]:
+        """
+        Returns available versions in ascending order
+        """
+        pass
+
+    @abstractmethod
+    def propose_version(self, config: str, new_version: Version, note: Optional[str] = None) -> None:
+        """
```
richard-rogers, line 59: I think as currently written this can only be called once for a new version. Should I allow multiple calls to update the new version until it's committed (or dropped)? Should there be a separate `update_version(self, config, new_version, note)` method for updating an existing new version until it's committed (or dropped)?

python/whylogs/experimental/extras/confighub.py
```diff
@@ -0,0 +1,130 @@
+import logging
+import os
+from abc import ABC, abstractmethod
+from typing import List, Optional
+
+from semantic_version import Version
+
+logger = logging.getLogger(__name__)
+
+
```
richard-rogers, line 10: This API might not be great if we expect a lot of concurrent editing. Is that a use case we need to support?

python/whylogs/experimental/extras/confighub.py
```diff
@@ -0,0 +1,147 @@
...
+        """
+        Make the proposed config the latest version. ValueError if proposed new_version doesn't exist
+        """
+        pass
+
+
+class LocalGitConfigStore(ConfigStore):
+    def _run_commands(self, commands: List[str]) -> None:
+        for command in commands:
+            output = os.popen(f"cd {self._path}; {command}")
```
richard-rogers, line 76: I should use the fancier `subprocess.Popen()` so I can capture (and ignore) the `stderr` output. It's annoying though.

python/tests/experimental/api/constraints/test_constraints_yaml.py
```diff
@@ -0,0 +1,97 @@
...
+    with tempfile.TemporaryDirectory() as temp_dir:
+        input_yaml_name = "example.yaml"
+        output_yaml_name = "example2.yaml"
+        # Define the path for the YAML file
+        yaml_path = os.path.join(temp_dir, input_yaml_name)
+
+        # Write the YAML string to a file
+        with open(yaml_path, "w") as file:
+            file.write(yaml_string)
+        constraints = read_constraints_from_yaml(os.path.join(temp_dir, input_yaml_name))
```
richard-rogers: We need a `read_constraints_from_yaml_str(yaml: str)` for where the yaml has already been read from the `ConfigStore`.

python/tests/experimental/api/constraints/test_constraints_yaml.py
```diff
@@ -0,0 +1,97 @@
...
+        builder = ConstraintsBuilder(dataset_profile_view=reference_profile_view)
+        builder.add_constraints(constraints)
+        rehydrated_constraints = builder.build()
+        write_constraints_to_yaml(
```
richard-rogers: Likewise, constraints to yaml string that can be written by `ConfigStore`.

python/tests/experimental/api/constraints/test_constraints_yaml.py
FelipeAdachi: that's just a placeholder at the moment, for now I focused only on constraints yaml.

python/tests/experimental/api/constraints/test_constraints_yaml.py
FelipeAdachi: likewise, just placeholders, let's discuss what's needed and what's not

python/tests/experimental/api/constraints/test_constraints_yaml.py
FelipeAdachi: I'm doing that on the test for yaml strings!

python/whylogs/core/constraints/metric_constraints.py
FelipeAdachi: You're right!
In latest commit, I tried to reduce redundancy by keeping only "factory" and specific parameters of factory constraints in ConstraintsParam. The other ones: "name", "column_name", "metric", I'm getting from the MetricConstraint ( and MetricSelector) itself.

python/whylogs/experimental/api/constraints/__init__.py
FelipeAdachi: If `expression` is present, constraint will be assembled using it, provided it has a name.

python/whylogs/experimental/api/constraints/__init__.py
FelipeAdachi: done!

python/tests/experimental/api/constraints/test_constraints_yaml.py
FelipeAdachi: done!

python/tests/experimental/api/constraints/test_constraints_yaml.py
richard-rogers: The `ConfigStore` is keyed with org id, model id, and kind of config (constraints, validators, etc). Those might be good meta data. There's also a commit message, though that's not currently fetchable in the API. I can add it if we want to include it here.

FelipeAdachi: We can decide if the commit message is required or not, but I think it'd be important for the user to show his reasoning on why he's proposing the changes

python/whylogs/experimental/api/writer/cnwriter.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @deprecated_alias(profile="file")
+    def write(self, file: Writable, **kwargs: Any) -> Tuple[bool, str]:
+        if (
+            isinstance(file, (DatasetProfile, DatasetProfileView))
+            and self._cn_store is not None
+            and self._cn_store.exists()
+        ):
+            constraints_str = self._cn_store.get_latest()
+            constraints = read_constraints_from_yaml(constraints_str)
```
richard-rogers:
```suggestion
            constraints = read_constraints_from_yaml(input_str=constraints_str)
```

python/whylogs/experimental/api/constraints/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+            constraint_dict.pop("name")
+        returned_constraint = constraint_function(**constraint_dict)
+        returned_constraint.name = constraint_name
+        return returned_constraint
+    else:
+        logger.warning(f"Constraint factory {constraint_dict.get('factory')} not found.")
+        return None
+
+
+class ConstraintTranslator:
```
richard-rogers, line 93: If this class has no state I'd make the methods bare functions.

python/whylogs/experimental/api/writer/cnwriter.py
```diff
@@ -0,0 +1,86 @@
...
+        ):
+            # turn constraint report into columns and append them to the profile
+            constraints_str = self._cn_store.get_latest()
+            constraints = ConstraintTranslator().read_constraints_from_yaml(input_str=constraints_str)
+            builder = ConstraintsBuilder(file)
+            builder.add_constraints(constraints)
+            report = builder.build().generate_constraints_report()
+            uncompounded_report = _uncompound_report(report)
+            report_view = why.log(row=uncompounded_report).profile().view()
+            file._columns.update(report_view._columns)
```
richard-rogers, line 84: Views are meant to be read only. We should probably clone the original profile and insert the report columns into the clone...

python/whylogs/experimental/extras/confighub.py
richard-rogers: I'm not sure how well this implementation will work on non-POSIX environments. For hackathon I don't care :)

python/whylogs/experimental/extras/confighub.py
richard-rogers: Actually this might work for multiple calls.

python/whylogs/experimental/constraints_hub/backend/routers.py
```diff
@@ -0,0 +1,169 @@
...
+        api_key = os.environ["WHYLABS_API_KEY"]
+    except KeyError:
+        raise Exception(
+            "you must define WHYLABS_DEFAULT_ORG_ID, WHYLABS_DEFAULT_DATASET_ID and WHYLABS_API_KEY environment variables"
+        )
+    translator = ConstraintTranslator()
+    storage_folder_name = "constraints_storage"
+    local_storage_folder = os.path.join(os.getcwd(), storage_folder_name)
+    cs = LocalGitConfigStore(org_id, dataset_id, "constraints", repo_path=local_storage_folder)
+    cs.create()
```
richard-rogers, line 54:
```suggestion
    if not cs.exists():
        cs.create()
```

python/whylogs/experimental/constraints_hub/backend/routers.py
```diff
@@ -0,0 +1,169 @@
...
+                    }
+                )
+                column_list.append(column_scheme)
+        return {"entity_schema": column_list}
+
+
+@router.get("/latest_version")
+def get_latest_version() -> None:
+    latest_yaml = cs.get_latest()
+    return {"constraints_yaml": str(latest_yaml)}
```
richard-rogers:
```suggestion
    return {"constraints_yaml": latest_yaml}
```
I think the `str()` is redundant?

python/whylogs/experimental/constraints_hub/backend/routers.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+@router.post("/push_constraints")
+def push_constraints(
+    yaml_constraint: Annotated[YamlConstraint, Body(example={"constraints_yaml": yaml_example})]
+) -> None:
+    yaml_string = yaml_constraint.constraints_yaml
+    constraints = translator.read_constraints_from_yaml(input_str=yaml_string)
+    cur_ver = cs.get_version_of_latest()
+    new_ver = cur_ver.next_major()
```
richard-rogers, line 220: Maybe a UI widget/parameter to determine which version component gets bumped?

python/tests/experimental/api/constraints/test_constraints_yaml.py
```diff
@@ -0,0 +1,140 @@
...
+- column_name: legs
+  factory: count_below_number
+  number: 25
+  metric: counts
+  name: count of legs lower than 25
+- column_name: weight
+  factory: is_non_negative
+  metric: distribution
+  name: weight is non negative
+- expression: and <= animal:cardinality/lower_1 :animal:counts/n <= :animal:counts/n
```
richard-rogers, line 64:
```suggestion
- expression: and <= :animal:cardinality/lower_1 :animal:counts/n <= :animal:counts/n
```

python/whylogs/experimental/extras/confighub.py
FelipeAdachi: I think that's probably ok for now

python/whylogs/experimental/api/writer/cnwriter.py
FelipeAdachi: wouldn't merging them work?
but for the hackathon I think this is fine either way!

python/whylogs/core/constraints/metric_constraints.py
```diff
@@ -500,26 +518,26 @@ def validate(self, profile_view: Optional[DatasetProfileView] = None) -> bool:
         return True
 
     @deprecated(message="Please use generate_constraints_report()")
-    def report(self, profile_view: Optional[DatasetProfileView] = None) -> List[Tuple[str, int, int]]:
+    def report(self, profile_view: Optional[DatasetProfileView] = None) -> List[Tuple[Optional[str], str, int, int]]:
```
FelipeAdachi, line 521: Also a thought for post-hackathon:
this might break report access to some users, but then again, it's a deprecated function, so maybe it' ok. or would it be better to not carry those changes into report(), and leave it only for generate_constraints_report()?

python/whylogs/experimental/api/writer/cnwriter.py
richard-rogers: The code works, but the side effect of modifying a read-only profile is bad :(  It's easy enough to do the clone & modify, but it can wait 'til after the demo :)

jamie256:
Closing but we should review and pull changes from this branch.

PR #1231, opened 2023-05-05
Improved generate UDF schema

richard-rogers:
## Description

`s/generate_udf_schema/generate_udf_metric_schema/` in case we need to add `generate_udf_dataset_schema()`.
Adds `non_udf_resolvers` argument with default `STANDARD_RESOLVER` for convenience.

### DIFF
--- a/python/whylogs/experimental/core/metrics/udf_metric.py
+++ b/python/whylogs/experimental/core/metrics/udf_metric.py
@@ -188,7 +188,7 @@ def register_metric_udf(
     input columns of the specified type. The decorated function will automatically
     be a UDF in the UdfMetric.
 
-    Specify submetric_name to give the output of the UDf a name. submetric_name
+    Specify submetric_name to give the output of the UDF a name. submetric_name
     defautls to the name of the decorated function. Note that all lambdas are
     named "lambda" so omitting submetric_name on more than one lambda will result
     in name collisions.
--- a/python/whylogs/experimental/core/metrics/udf_metric.py
+++ b/python/whylogs/experimental/core/metrics/udf_metric.py
@@ -236,12 +236,12 @@ def register_metric_udf(
     return decorator_register
 
 
-def generate_udf_schema() -> List[ResolverSpec]:
+def generate_udf_metric_schema(non_udf_resolvers: List[ResolverSpec] = STANDARD_RESOLVER) -> List[ResolverSpec]:
     """
     Generates a list of ResolverSpecs that implement the UdfMetrics specified
-    by the @register_metric_ud decorators. The result only includes the UdfMetric,
-    so you may want to append it to a list of ResolverSpecs defining the other
-    metrics you wish to track.
+    by the @register_metric_ud decorators (in additon to any non_udf_resolvers
+    passed in). The result only includes the UdfMetric, so you may want to append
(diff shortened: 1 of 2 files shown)


jamie256 (commented):
I think we should leave the original name as it works, and create a new function that returns the schema, this still requires someone to construct the schema when they shouldn't have to know about it. Can we have a helper that you could pass to the log statement to simplify the calling pattern?

```
why.log(df, schema=standard_udf_schema())
```

richard-rogers:
> I think we should leave the original name as it works, and create a new function that returns the schema, this still requires someone to construct the schema when they shouldn't have to know about it. Can we have a helper that you could pass to the log statement to simplify the calling pattern?
> 
> ```
> why.log(df, schema=standard_udf_schema())
> ```

That is the calling pattern this PR allows `why.log(df, schema=generate_udf_metric_schema())`. I'm not opposed to improving the name, but I want to avoid confusion between the UDF submetric schema and dataset UDF schema (I'm still on the fence about including both, but I'm updating the computed column PR with similar UDF decorator.)

Er, I guess you do need to include the `DeclarativeSchema` ctor, but I can change it to `generate_udf_metric_schema() -> DeclarativeSchema` directly now that you can pass in any extra resolver specs you need...

python/whylogs/experimental/core/metrics/udf_metric.py
```diff
@@ -287,4 +287,4 @@ def upper(x):
         )
         resolvers.append(ResolverSpec(None, col_type, [MetricSpec(UdfMetric, config)]))
 
-    return resolvers
+    return DeclarativeSchema(non_udf_resolvers + resolvers)
```
richard-rogers, line 324: Consider splitting into separate functions or passing the other `DeclarativeSchema` ctor arguments.

jamie256 (commented):
I like this new method. But I think we are removing something people might be using. For example this change breaks our existing examples and notebooks that used the `generate_udf_schema()`

Can we have the existing method generate both, eventually? I suggest we leave in the original method unchanged and have this be a new method side by side:
`def generate_udf_metric_schema(non_udf_resolvers: List[ResolverSpec] = STANDARD_RESOLVER) -> DeclarativeSchema`

If we don't like the other method we should mark as deprecated rather than just remove it. Experimental namespace means it might go away but if its not an underscore prefixed method we need to mark it deprecated and remove with next minor version bump not in a patch bump.

python/whylogs/experimental/core/metrics/udf_metric.py
```diff
@@ -236,12 +238,11 @@ def decorator_register(func):
     return decorator_register
 
 
-def generate_udf_schema() -> List[ResolverSpec]:
+def generate_udf_metric_schema(non_udf_resolvers: List[ResolverSpec] = STANDARD_RESOLVER) -> DeclarativeSchema:
```
jamie256: The default value in the signature should be None and then set inside to avoid global instance of the list.

python/whylogs/experimental/core/metrics/udf_metric.py
jamie256: we can do more refactors in follow up, this should cover initial cases, thanks!

jamie256 (approved):

PR #1235, opened 2023-05-11
Add new apis to enable guest sessions from whylogs

naddeoa:
This change introduces the concept of a session that can be initialized in order to enable a new UX that integrates with WhyLabs, without requiring any initial user sign up. The new apis work as follows

```python
import whylogs as why

why.init(anonymous=True)

result_set = why.log_reference("test1", {"foo": "bar", "baz": 1})

result_sets = why.batch_log_reference({
    'wine target': wine_target,
    'wine reference': wine_reference
})
```

The new apis here are `init`, `log_reference`, and `batch_log_reference`.

The `init` api is optional. Existing users can continue to do what they're doing today. It can be called with `anonymous=True` to create an anonymous session, or it can be called without that flag to create a normal session. A session is mostly an internal detail. It's something that is available to other parts of the library if they want to know the user's authentication/configuration status. For example, `log_reference` will check to see if there is a session and if so, it will use that session to upload profiles instead of just returning the profiles locally, also printing out the whylabs url that they can be viewed at. At the moment, `log_reference` and `batch_log_reference` are the only logging apis that are aware of sessions. We can follow up with other apis.

The `log_reference` and `batch_log_reference` apis are intended to log reference profiles, as opposed to normal batch profiles that have timestamps. They require an alias along with their data. The only real distinction between batch profiles and reference profiles is the downstream use in whylabs and the absence of time in reference profiles, but they're identical structurally.


python/whylogs/api/logger/__init__.py
```diff
@@ -36,6 +37,68 @@ def log(
...
+        # Don't throw, just don't upload
+        diagnostic_logger.error(f"Failed to upload reference profile: {e}")
+
+    return result_set
+
+
+Loggable = Union[pd.DataFrame, Dict[str, Any], List[pd.DataFrame], List[Dict[str, Any]]]
+
+
+def batch_log_reference(
```
andyndang, line 72: can you document what this methods do? It's unclear from the signature

andyndang (commented):
Left a comment on the api client creation

PR #1237, opened 2023-05-12
udf_metric_schema() supports all schema arguments

### LINKED ISSUE #1240: udf_metrics need try/except around execution
## Description

Some udf metrics might wrap a library that can throw, and we don't want to block other metrics from proceeding. Better would be to log this and continue rather than let an unhandled exception raise.

e.g. division by zero error in a udf.

richard-rogers:
## Description

- Pass through all schema constructor arguments to `udf_metric_schema()`
- #1240 add try/except around UDF evaluation

Closes #1240


jamie256 (approved):
LGTM!

It is better to protect against unruly UDFs, thanks!

PR #1238, opened 2023-05-12
Allow hiding condition count metrics from profile

richard-rogers:
## Description

Adds metric config option to turn off uncompounding a condition count metric instance so we can deprecate validators.

### DIFF
--- a/python/whylogs/migration/uncompound.py
+++ b/python/whylogs/migration/uncompound.py
@@ -114,7 +114,7 @@ def _uncompound_multimetric(col_name: str, metric_name: str, metric: MultiMetric
 def _uncompound_condition_count(
     col_name: str, metric_name: str, metric: ConditionCountMetric, flags: Optional[FeatureFlags] = None
 ) -> Dict[str, ColumnProfileView]:
-    if not _uncompound_condition_count_feature_flag(flags):
+    if metric.hide_from_profile or not _uncompound_condition_count_feature_flag(flags):
         return dict()
 
     result: Dict[str, ColumnProfileView] = dict()
--- a/python/tests/migration/test_uncompound.py
+++ b/python/tests/migration/test_uncompound.py
@@ -1,6 +1,7 @@
 import os
 
 import pandas as pd
+import pytest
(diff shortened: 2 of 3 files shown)


FelipeAdachi (commented):
This prevents from uncompounding the Condition Count metric and being sent to WhyLabs, but the metrics will still be present in the profile, right?

If I'm not mistaken, we had feedbacks in the past where we'd like the Metric not to be in the profile altogether, since it was not needed - the conditions were only used to trigger the actions.

We can make the decision of not supporting that case, but just wanted to point out that currently, this change does not fully address that.

richard-rogers:
> This prevents from uncompounding the Condition Count metric and being sent to WhyLabs, but the metrics will still be present in the profile, right?

Yes, this only prevents creating new uncompounded columns; the metric (and submetrics) will still be present in the original column. I vote against removing the metric from the column.

python/whylogs/migration/uncompound.py
```diff
@@ -114,7 +114,7 @@ def _uncompound_multimetric(col_name: str, metric_name: str, metric: MultiMetric
 def _uncompound_condition_count(
     col_name: str, metric_name: str, metric: ConditionCountMetric, flags: Optional[FeatureFlags] = None
 ) -> Dict[str, ColumnProfileView]:
-    if not _uncompound_condition_count_feature_flag(flags):
+    if metric.hide_from_profile or not _uncompound_condition_count_feature_flag(flags):
```
jamie256, line 117: The request we had for data validators was a non-whylabs use case. The uncompound callstack is not the correct place to remove the metrics because it does not cover the callstacks that write profiles with the local/s3/gcs writers. If we want to consolidate data validators and condition count we probably need to use a different implementation to hide these metrics?

But no objections to this change on its own.

jamie256 (approved):
LGTM, might not be sufficient to satisfy the motivating scenario of deprecating data validators.

jamie256:
I think we want the approach in #1247 closing in favor of the other PR.

PR #1242, opened 2023-05-18
constraints generator fixes - customizable hll_stddev and error handling

FelipeAdachi:
## Description

- changes `column_is_probably_unique` to accept `hll_stddev` param (defaults to 3 - it used to be 1)

This change was motivated by the fact that the `is_probably_unique` constraint was yielding a high false negative with a `hll_stddev` of 1

- prevent missing metrics from raising an error and instead skip with warning

When a column has a different type between reference and target profiles, the auto constraints generator suggests constraints that requires metrics that are missing in the target profile. That would lead to an error when calling `add_constraint`. To prevent from crashing, this PR doesn't raise an error in this case and instead skips the constraint to be added, along with a warning. 

## Changes

## Related

### DIFF
--- a/python/.bumpversion.cfg
+++ b/python/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.1.37
+current_version = 1.1.41
 tag = False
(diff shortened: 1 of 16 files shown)


jamie256 (approved):
Thanks Felipe, looks better!

PR #1244, opened 2023-05-20
Have UDF base schema not include frequent items by default

jamie256:
## Description

We need an easier calling pattern for integrations using the UDF schemas that are privacy preserving. Here's one way we could support that.

### DIFF
--- a/python/whylogs/core/resolvers.py
+++ b/python/whylogs/core/resolvers.py
@@ -197,6 +197,36 @@ STANDARD_RESOLVER = [
     ResolverSpec(column_type=AnyType, metrics=COLUMN_METRICS),
 ]
 
+STANDARD_PRIVACY_PRESERVING_RESOLVER = [
+    ResolverSpec(
+        column_type=Integral,
+        metrics=COLUMN_METRICS
+        + [
+            MetricSpec(StandardMetric.distribution.value),
+            MetricSpec(StandardMetric.ints.value),
+            MetricSpec(StandardMetric.cardinality.value),
+        ],
+    ),
+    ResolverSpec(
+        column_type=Fractional,
+        metrics=COLUMN_METRICS
+        + [
+            MetricSpec(StandardMetric.distribution.value),
+            MetricSpec(StandardMetric.cardinality.value),
+        ],
+    ),
+    ResolverSpec(
+        column_type=String,
+        metrics=COLUMN_METRICS
+        + [
+            MetricSpec(StandardMetric.unicode_range.value),
+            MetricSpec(StandardMetric.distribution.value),
+            MetricSpec(StandardMetric.cardinality.value),
+        ],
+    ),
+    ResolverSpec(column_type=AnyType, metrics=COLUMN_METRICS),
+]
+
 LIMITED_TRACKING_RESOLVER = [
     ResolverSpec(
         column_type=Integral,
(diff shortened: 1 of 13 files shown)


python/whylogs/experimental/core/metrics/udf_metric.py
```diff
@@ -345,7 +346,7 @@ def upper(x):
     """
 
     resolvers = generate_udf_resolvers()
-    non_udf_resolvers = non_udf_resolvers if non_udf_resolvers is not None else STANDARD_RESOLVER
+    non_udf_resolvers = non_udf_resolvers if non_udf_resolvers is not None else STANDARD_PRIVACY_PRESERVING_RESOLVER
```
richard-rogers: see line 64

jamie256: Can you elaborate?

richard-rogers: It's defaulting to `STANDARD_RESOLVER` for `UdfMetric` submetrics.

jamie256: Right, which I think is what we want. We don't want the full metrics on the raw inputs, so that the focus is on the UDF output which should get the standard resolver.

richard-rogers: That will still include frequent items on the UDF outputs.

jamie256: its a bit heavy handed here, basically we are lowering the number of metrics you get on columns even when the UDFs aren't targeting them if you have a mix of UDFs and non-UDF columns. I'm not suggesting this goes straight to mainline, but I'd like to test with this in a dev build since the API is closer to the simple calling pattern I think we want to get to for UDF cases.

jamie256: We don't have to merge this at all, but I think I'll create a dev build here to get some feedback on the impact to callers integrating with UDFs. Maybe a thumbs up here would do for me, mostly just looking for imput from @richard-rogers before I create the next dev build

python/whylogs/experimental/core/metrics/udf_metric.py
```diff
@@ -345,7 +346,7 @@ def upper(x):
     """
 
     resolvers = generate_udf_resolvers()
-    non_udf_resolvers = non_udf_resolvers if non_udf_resolvers is not None else STANDARD_RESOLVER
+    non_udf_resolvers = non_udf_resolvers if non_udf_resolvers is not None else UDF_BASE_RESOLVER
```
richard-rogers, line 349: Let's make the default resolver list configurable 🥑

jamie256: Maybe we can do that in a follow up, this could still be the default default for udfs.

jamie256 (commented):
@richard-rogers If we can merge this into your branch and then test there I think we have something that works with our test participants and should give us better udfs defaults, and segmentation support. We can do another round of review/testing on your branch post merge.

python/whylogs/core/resolvers.py
```diff
@@ -197,6 +197,36 @@ def resolve(self, name: str, why_type: DataType, column_schema: ColumnSchema) ->
     ResolverSpec(column_type=AnyType, metrics=COLUMN_METRICS),
 ]
 
+UDF_BASE_RESOLVER = [
```
richard-rogers, line 200: `s/UDF_BASE_RESOLVER/PRIVACY_PRESERVING_RESOLVER/`
or `NO_FI_RESOLVER`
or `SIMPLE_RESOLVER`
...

richard-rogers (approved):

PR #1246, opened 2023-05-20
Configurable MetricConfig defaults

richard-rogers:
## Description

Lets users change globals in `whylogs/core/configs.py` to set the default values for `MetricConfig`.

[clickup](https://app.clickup.com/t/866ab6rhg)

## Changes

- `MetricConfig::__inti__()` now takes all `Optional` arguments and uses `whylogs.core.configs` variables for `None` passed in. That was tricky, since it's a frozen data class...


python/whylogs/experimental/core/metrics/udf_metric.py
```diff
@@ -76,9 +76,45 @@ class UdfMetricConfig(MetricConfig):
...
+            track_unicode_ranges,
+            large_kll_k,
+            kll_k_large,
+            unicode_ranges,
+            lower_case,
+            normalize,
+            max_frequent_item_size,
+            identity_column,
+        )
+        self.__dict__["udfs"] = udfs or dict()
```
richard-rogers, line 117:
```suggestion
        self.__dict__["udfs"] = dict(udfs) if udfs is not None else dict()
```

python/whylogs/core/metrics/metrics.py
```diff
@@ -37,28 +37,50 @@
...
+        self.__dict__["hll_lg_k"] = hll_lg_k if hll_lg_k is not None else conf.hll_lg_k
+        self.__dict__["kll_k"] = kll_k if kll_k is not None else conf.kll_k
+        self.__dict__["fi_lg_max_k"] = fi_lg_max_k if fi_lg_max_k is not None else conf.fi_lg_max_k
+        self.__dict__["fi_disabled"] = fi_disabled if fi_disabled is not None else conf.fi_disabled
+        self.__dict__["track_unicode_ranges"] = (
+            track_unicode_ranges if track_unicode_ranges is not None else conf.track_unicode_ranges
+        )
+        self.__dict__["large_kll_k"] = large_kll_k if large_kll_k is not None else conf.large_kll_k
+        self.__dict__["kll_k_large"] = kll_k_large if kll_k_large is not None else conf.kll_k_large
+        self.__dict__["unicode_ranges"] = unicode_ranges if unicode_ranges is not None else dict(conf.unicode_ranges)
```
jamie256, line 77: Question is this ok with respect to instancing the dictionary, do you want to copy?

python/whylogs/core/resolvers.py
```diff
@@ -48,7 +48,10 @@ def resolve(self, name: str, why_type: DataType, column_schema: ColumnSchema) ->
             metrics.append(StandardMetric.frequent_items)
 
         if column_schema.cfg.fi_disabled:
-            metrics.remove(StandardMetric.frequent_items)
+            try:
+                metrics.remove(StandardMetric.frequent_items)
```
jamie256, line 53: Consider: moving the modification of the metrics to operate on the results dictionary? Could be easier to use get() than a try catch, and slightly more efficient to use the dictionary key.

python/whylogs/extras/image_metric.py
```diff
@@ -150,6 +150,40 @@ class ImageMetricConfig(MetricConfig):
...
+            large_kll_k,
+            kll_k_large,
+            unicode_ranges,
+            lower_case,
+            normalize,
+            max_frequent_item_size,
+            identity_column,
+        )
+        self.__dict__["allowed_exif_tags"] = allowed_exif_tags or set()
+        self.__dict__["forbidden_exif_tags"] = forbidden_exif_tags or set()
```
jamie256, line 185: See if we need to define this init at all? Maybe the base class gives us the behavior already?

python/whylogs/experimental/core/metrics/udf_metric.py
jamie256: can we move that to a field(default_factory=...) rather than having to override init?

jamie256 (commented):
Consider if we can simplify to not have to override the init on the dataclasses?

richard-rogers:
#1257 is better

PR #1247, opened 2023-05-23
Allow hiding condition count metric from whylabs profile

richard-rogers:
## Description

Allows optional hiding condition of count metrics in the profile sent to WhyLabs with the goal of deprecating validators.

## Changes

- `Metric` has an `exclude_from_profile` property
- Excluded metrics are not included in the protobuf produced from the profile
- `ConditionCountMetric` instances can be configured to exclude themselves

## Related

Alternative approach for #1238 


python/tests/core/metrics/test_condition_count.py
```diff
@@ -253,6 +254,31 @@ def resolve(self, name: str, why_type: DataType, column_schema: ColumnSchema) ->
...
+
+    conditions = {
+        "alpha": Condition(X.matches("[a-zA-Z]+")),
+        "digit": Condition(X.matches("[0-9]+")),
+    }
+    config = ConditionCountConfig(conditions=conditions, exclude_from_profile=hide)
+    resolver = TestResolver()
+    schema = DatasetSchema(default_configs=config, resolvers=resolver)
+
+    row = {"col1": ["abc", "123"]}
```
FelipeAdachi: minor thing, but the conditions are being checked against a list.

FelipeAdachi (dismissed):
LGTM!

this is one step towards converging CCM w/ Validators. But to reach complete parity, we need to move Reservoir Sampling /  Row_ID to CCMs as well (in a follow-up)

python/whylogs/core/metrics/condition_count_metric.py
```diff
@@ -67,13 +67,19 @@ class Condition:
 @dataclass(frozen=True)
 class ConditionCountConfig(MetricConfig):
     conditions: Dict[str, Condition] = field(default_factory=dict)
+    exclude_from_profile: bool = False
```
jamie256: Suggest: rename to something like "exclude_from_serialization"

consider moving to base class metric config.

python/whylogs/core/metrics/condition_count_metric.py
```diff
@@ -67,13 +67,19 @@ class Condition:
...
 
 
 @dataclass(frozen=True)
 class ConditionCountMetric(Metric):
     conditions: Dict[str, Condition]
     total: IntegralComponent
     matches: Dict[str, IntegralComponent] = field(default_factory=dict)
+    hide_from_profile: bool = False
```
jamie256: hide_from_serialization?

jamie256 (dismissed):
LGTM, consider some of the renames so its not "profile" that the metric is excluded from.

jamie256 (approved):
LGTM!

PR #1248, opened 2023-05-23
Upload segmented reference profile API

FelipeAdachi:
## Description

Enables uploading segmented reference profiles.
Usage: just as you'd upload regular reference profiles, but instead, pass a SegmentedResultSet to it (see `test_whylabs.py`)

Since, to generate the URLs, we need the complete ResultSet, a SegmentedResultSet is passed to `write`. In the whylabs writer, if a `SegmentedResultSet` is passed along with a `reference_profile_name`, `the write_segmented_reference_profile is called`.

### DIFF
--- a/python/whylogs/api/writer/whylabs.py
+++ b/python/whylogs/api/writer/whylabs.py
@@ -491,7 +491,18 @@ class WhyLabsWriter(Writer):
             return feature_weights
         return None
 
-    def write_segmented_reference_result_set(self, file: SegmentedResultSet, **kwargs: Any):
+    def write_segmented_reference_result_set(self, file: SegmentedResultSet, **kwargs: Any) -> Tuple[bool, str]:
+        """Put segmented reference result set for the specified dataset.
+
+        Parameters
+        ----------
+        file : SegmentedResultSet
+            SegmentedResultSet object representing the segmented reference result set for the specified dataset
+
+        Returns
+        -------
+        Tuple[bool, str]
+        """
         utc_now = datetime.datetime.now(datetime.timezone.utc)
 
         files = file.get_writables()
@@ -515,10 +526,9 @@ class WhyLabsWriter(Writer):
             whylabs_tags.append(view_tags)
         stamp = dataset_timestamp.timestamp()
         dataset_timestamp_epoch = int(stamp * 1000)
-        res = self._get_upload_urls_segmented_reference(whylabs_tags, dataset_timestamp_epoch)
-        profile_id = res["id"]
+        profile_id, upload_urls = self._get_upload_urls_segmented_reference(whylabs_tags, dataset_timestamp_epoch)
         upload_statuses = list()
-        for view, url in zip(files, res["upload_urls"]):
+        for view, url in zip(files, upload_urls):
             with tempfile.NamedTemporaryFile() as tmp_file:
                 if kwargs.get("use_v0") is None or kwargs.get("use_v0"):
                     view.write(file=tmp_file, use_v0=True)
@@ -610,12 +620,8 @@ class WhyLabsWriter(Writer):
                 )
 
             dataset_timestamp_epoch = int(stamp * 1000)
-            logger.debug("Generating the upload URL")
-            upload_url, profile_id = self._get_upload_url(dataset_timestamp=dataset_timestamp_epoch)
             response = self._do_upload(
                 dataset_timestamp=dataset_timestamp_epoch,
-                upload_url=upload_url,
-                profile_id=profile_id,
                 profile_file=tmp_file,
             )
         # TODO: retry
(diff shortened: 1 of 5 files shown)


python/whylogs/api/writer/whylabs.py
```diff
@@ -476,20 +484,79 @@ def get_feature_weights(self, **kwargs: Any) -> Optional[FeatureWeights]:
...
         result = self._do_get_feature_weights()
-        feature_weights_set = result.get("segmentWeights")
+        feature_weights_set = result.get("segment_weights")
         metadata = result.get("metadata")
         if feature_weights_set and isinstance(feature_weights_set, list):
             feature_weights = FeatureWeights(weights=feature_weights_set[0]["weights"], metadata=metadata)
             return feature_weights
         return None
 
+    def write_segmented_reference_result_set(self, file: SegmentedResultSet, **kwargs: Any) -> Tuple[bool, str]:
```
jamie256: prefix with underscore is not a new public API we want customers to use.

python/whylogs/api/writer/whylabs.py
```diff
@@ -476,20 +484,79 @@ def get_feature_weights(self, **kwargs: Any) -> Optional[FeatureWeights]:
...
+            return False, "Failed to upload all segments"
+
     @deprecated_alias(profile="file")
     def write(self, file: Writable, **kwargs: Any) -> Tuple[bool, str]:
         if isinstance(file, FeatureWeights):
             return self.write_feature_weights(file, **kwargs)
         elif isinstance(file, EstimationResult):
             return self.write_estimation_result(file, **kwargs)
-
+        elif isinstance(file, SegmentedResultSet) and self._reference_profile_name is not None:
```
jamie256, line 558: Consider making SegmentedResultSet a Writeable (can be done in follow up). Looks like we are missing this base class on some of the other things we write.

python/whylogs/api/writer/whylabs.py
```diff
@@ -476,20 +484,79 @@ def get_feature_weights(self, **kwargs: Any) -> Optional[FeatureWeights]:
...
+                    profile_file=tmp_file,
+                )
+                upload_statuses.append(upload_res)
+        if all([status[0] for status in upload_statuses]):
+            return upload_statuses[0]
+        else:
+            return False, "Failed to upload all segments"
+
     @deprecated_alias(profile="file")
     def write(self, file: Writable, **kwargs: Any) -> Tuple[bool, str]:
```
jamie256, line 553: We need to follow up (not in this PR) and make other ResultSet's something this method supports, so its not just segmented results.

python/whylogs/api/writer/whylabs.py
```diff
@@ -476,20 +484,79 @@ def get_feature_weights(self, **kwargs: Any) -> Optional[FeatureWeights]:
...
+            dataset_timestamp = view.dataset_timestamp or utc_now
+            if view.partition.id != partition.id:
+                continue
+            _, segment_tags, _ = _generate_segment_tags_metadata(view.segment, view.partition)
+            for segment_tag in segment_tags:
+                tag_key = segment_tag.key.replace("whylogs.tag.", "")
+                tag_value = segment_tag.value
+                view_tags.append({"key": tag_key, "value": tag_value})
+            whylabs_tags.append(view_tags)
+        stamp = dataset_timestamp.timestamp()
```
jamie256, line 527: I think we have a utility that does this? Lets factor out in follow up.

python/whylogs/api/writer/whylabs.py
```diff
@@ -476,20 +484,79 @@ def get_feature_weights(self, **kwargs: Any) -> Optional[FeatureWeights]:
...
+                tmp_file.flush()
+                tmp_file.seek(0)
+
+                upload_res = self._do_upload(
+                    dataset_timestamp=dataset_timestamp_epoch,
+                    upload_url=url,
+                    profile_id=profile_id,
+                    profile_file=tmp_file,
+                )
+                upload_statuses.append(upload_res)
```
jamie256, line 546: Can we move this to a function in a follow up.

jamie256 (changes requested):
Can you prefix the new method with underscore so we aren't adding new public API with this change. Can you file issues for the follow up items, then I think we can merge this to unblock people, with the right API shape and fix the internals separately.

python/whylogs/api/writer/whylabs.py
FelipeAdachi: created issue for it https://github.com/whylabs/whylogs/issues/1250

python/whylogs/api/writer/whylabs.py
FelipeAdachi: https://github.com/whylabs/whylogs/issues/1249

python/whylogs/api/writer/whylabs.py
FelipeAdachi: https://github.com/whylabs/whylogs/issues/1251

python/whylogs/api/writer/whylabs.py
FelipeAdachi: https://github.com/whylabs/whylogs/issues/1251

python/whylogs/api/writer/whylabs.py
FelipeAdachi: done in latest commit

jamie256 (approved):
LGTM!

PR #1258, opened 2023-05-26
make resultsets/estimation result writables

FelipeAdachi:
## Description

Currently, non-Writable objects were being passed to WhyLabs Writers, which expected a ]Writable object. This PR:

- Makes all ResultSets Writables
- Makes Performance Estimation Result a Writable
- Added checks for write results in some of the writers
- Fixed signature and return for HTML Report

Even though Estimation Result and Segmented Result Set are Writables, the `write` method is not implemented and will return a write status with a message suggesting the use of supported writers.

### DIFF
--- a/python/whylogs/api/logger/result_set.py
+++ b/python/whylogs/api/logger/result_set.py
@@ -1,4 +1,4 @@
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from datetime import datetime
 from logging import getLogger
 from typing import Any, Dict, List, Optional, Tuple, Union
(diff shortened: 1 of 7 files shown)


python/whylogs/api/logger/result_set.py
```diff
@@ -161,7 +161,7 @@ def read(self, **kwargs: Any) -> "ResultSet":
         return self._reader.read(**kwargs)
 
 
-class ResultSet(ABC):
+class ResultSet(Writable):
```
richard-rogers, line 164:
```suggestion
class ResultSet(ABC, Writable):
```

python/whylogs/api/logger/result_set.py
```diff
@@ -1,4 +1,4 @@
-from abc import ABC, abstractmethod
+from abc import abstractmethod
```
richard-rogers, line 1:
```suggestion
from abc import ABC, abstractmethod
```

PR #1267, opened 2023-06-02
Add new apis to enable guest sessions from whylogs

naddeoa:
This change introduces the concept of a session that can be initialized in order to enable a new UX that integrates with WhyLabs, without requiring any initial user sign up. The new apis work as follows

```python
import whylogs as why

why.init(session_type='whylabs_anonymous')

result_set = why.log(df, name="foo")

result_set = why.log(multiple={"foo": df1, "bar": df2})
```

The `init` api is optional. Existing users can continue to do what they're doing today. It can be called with
`session_type='whylabs_anonymous'` to create an anonymous session. A session is mostly an internal detail. It's something that is available to other parts of whylogs when we want to know the user's authentication/configuration status. For example, `log` will check to see if there is a session and if so, it will use that session to upload profiles instead of just returning the profiles locally, also printing out the whylabs url that they can be viewed at. At the moment, `log` is the only logging api that is aware of sessions. We can follow up with other apis.

This version is intended to log reference profiles, as opposed to normal batch profiles that have timestamps. They require an alias along with their data. The only real distinction between batch profiles and reference profiles is the downstream use in whylabs and the absence of time in reference profiles, but they're identical structurally.

## Description

This repository needs 

## Changes

## Related


python/examples/integrations/Fugue_Profiling.ipynb
jamie256, line 1: If we don't need edits to the fugue example can we remove the changes to this file?

python/whylogs/api/whylabs/session/session_manager.py
```diff
@@ -0,0 +1,285 @@
...
+    except Exception as e:
+        logger.warning("Could not initialize session", e)
+
+
+def get_current_session() -> Optional[Session]:
+    manager = SessionManager.get_instance()
+    if manager is not None:
+        return manager.session
+
+    logger.warning("No session is initialized. Call whylogs.init(anonymous=True) to create a guest session.")
```
jamie256, line 284: If I have a WhyLabs account and I'm logging production data will this create a warning message on each log call?

If I have whylogs running in a headless env and don't want a session will this also emit a warning?

python/whylogs/api/whylabs/session/notebook_check.py
```diff
@@ -0,0 +1,28 @@
...
+def is_jupyter_notebook() -> bool:
+    """
+    Detects whether the current environment is a Jupyter notebook or not.
+    """
+    try:
+        shell = get_ipython().__class__.__name__  # type: ignore
+        if shell == "ZMQInteractiveShell":
+            return True  # Jupyter notebook or qtconsole
+        elif shell == "TerminalInteractiveShell":
+            return False  # Terminal running IPython
```
jamie256, line 14: the else block below already covers this, don't need to test for terminal string specifically.

python/whylogs/api/whylabs/session/notebook_check.py
```diff
@@ -0,0 +1,28 @@
+def is_notebook() -> bool:
+    return is_jupyter_notebook() or is_colab_notebook()
+
+
+def is_jupyter_notebook() -> bool:
+    """
+    Detects whether the current environment is a Jupyter notebook or not.
+    """
+    try:
+        shell = get_ipython().__class__.__name__  # type: ignore
```
jamie256, line 10: consider testing if get_ipython is defined rather than trying to call it and catching the NameError, e.g.
`if "get_ipython" is in dir()`

probably this could be simplified a bit.

python/whylogs/api/whylabs/session/notebook_check.py
```diff
@@ -0,0 +1,28 @@
...
+            return True  # Jupyter notebook or qtconsole
+        elif shell == "TerminalInteractiveShell":
+            return False  # Terminal running IPython
+        else:
+            return False
+    except NameError:
+        return False  # Probably standard Python interpreter
+
+
+def is_colab_notebook() -> bool:
```
jamie256, line 21: this raises uncaught ModuleNotFoundError

python/whylogs/api/whylabs/session/session_manager.py
```diff
@@ -0,0 +1,285 @@
...
+                result_set = profile_aliases[ref.alias]
+                writables = result_set.get_writables()
+
+                if writables is None:
+                    logger.warning(f"ResultSet for alias {ref.alias} has no profiles. Skipping upload.")
+                    continue
+
+                if len(writables) == 1:
+                    view = cast(DatasetProfileView, writables[0])  # TODO Jamie, is this right?
+                    r.put(ref.upload_url, data=view.serialize())
```
jamie256, line 169: in whylabs writer we use a tmp file, and have a pattern like:
```
for view in writeables:
    view.write(file=tmp_file)
```

see : https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/api/writer/whylabs.py#L590

then the upload is copying the tmp file. DatasetProfileView's write method allows you to specify v0 format as an optional kwarg: use_v0=True

What you have theoretically should work for most basic cases, but you are skipping checks to see if WhyLab's supports v1 metrics such as condition counts and multi-metrics. We could just call _uncompound on the anonymous session's profiles?

python/whylogs/api/logger/__init__.py
```diff
@@ -33,8 +40,22 @@ def log(
...
+            result_set = log(obj=d, schema=schema)
+            result_sets[alias] = result_set
+
+        # Return one result set with everything in it since we have to return result_sets
+        result_set = reduce(lambda r1, r2: r1.merge(r2), result_sets.values())
+        notebook_session_log_comparison(multiple, result_sets)
+        return result_set
+    else:
+        result_set = TransientLogger(schema=schema).log(obj, pandas=pandas, row=row, name=name)
+        notebook_session_log(result_set, obj, pandas=pandas, row=row, name=name)
```
jamie256, line 57: I wonder if there is an easy high level flag or something we can pull up for a check for which path to execute? I'm looking for a very easy to reason about structure on if the code is executing notebook onboarding flow vs production logging. The `mulitple is not None` is good in that way, but I have to step through notebook_session_log to know if its early exiting or doing something interesting?

python/whylogs/api/whylabs/session/session_manager.py
```diff
@@ -0,0 +1,285 @@
+import logging
+import time
+import uuid
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from enum import Enum
+from typing import Dict, List, Optional, Union, cast
+
+import requests as r
```
jamie256, line 9: requests is not an existing whylogs core dependency, we will need to add this to pyproject.toml, can do with `poetry add requests` and then need to update the poetry lock file.

jamie256 (changes requested):
High level API changes and approach look ok. I want to test this a bit more and can help make updates to this branch.

Can you try and get the CI tests running and help with some of the following:
* `poetry add requests` -> unblock the CI tests failing on not having that package, get the CI tests running and mostly passing.
* remove the edits to the fugue example notebook

python/whylogs/api/whylabs/session/session_manager.py
naddeoa: The code for production logging isn't here yet, but if you were to do it with something like :

```
why.init(api_ke='...', ...)
```

Then you would end up seeing this message (which is currently tailored towards guest sessions since the production init doesn't exist yet). I think the assumption there is that you ran init with the intent of having it actually work. The message has to change after I add the ability to actually do api key init of course.

You would see this though even if you didn't use a session, which I agree isn't right. I updated it to use the `log_if_notebook` util so that it's only shown in a notebook environment, and I added a flag so that it only happens once.

python/whylogs/api/whylabs/session/notebook_check.py
naddeoa: Seems to work. Just switched to it.

python/whylogs/api/whylabs/session/session_manager.py
naddeoa: - `view.write(file=tmp_file, use_v0=True)` It looks like you can use an instance of DatasetProfileView to .write something besides itself by specifying a file? Should that just be a static function?
- Are temporary files required? Wouldn't it be better to just do it all in memory?
- I'm explicitly testing for len==1 when writing to make sure there are no segments in my result set. Is there a better way to ensure that I'm just getting a single dataset profile from logging?

python/whylogs/api/logger/__init__.py
naddeoa: These are two separate things I think. This function doesn't have any notebook onboarding logic, it's just determining if its going to log a single thing or multiple things.

The notebook_session_log_comparison is doing some logic to determine how much "extra" to do, which is determined by the session type (if there is one). So both paths just do logging, and both paths have extra session logic they execute if there is one.

I guess in that way it might be better to call these just `session_log` or something. They do the profile uploading as well as the notebook logging if you're in a notebook.

python/whylogs/api/whylabs/session/session_manager.py
jamie256: Answers to the bullet point questions:
* DatasetProfileView's `write` can take an existing open file to write to or other optional kwargs, it always write's itself as the content, but the path/file tells where and how to write the content.

* Ideally tmp files wouldn't be required, but its somewhat baked into the flow because we need to know file offsets for the headers since the binary file format is chunked. The serialization uses a tmp file to write the content to and then update the headers with the actual offsets based on what we wrote to the tmp file. Additionally having a tmp file makes retry of upload more reliable, so the serialization happens first, then upload.

* You could check the DatasetSchema to see if it has segmentation defined, then you should expect segmented result sets to be returned from log statements using that schema. Its valid to have a single segment profile since you might have sparse data, in any case if all rows have the same value for the segmentation column it will result in a single segmented profile. Alternatively to checking the DatasetSchema you can check the type of the result set, segmented profiles will be wrapped in a SegmentedResultSet (or you can check if the result_set has a segments() method on it without needing to check the class type).

python/whylogs/api/logger/__init__.py
jamie256: makes sense, maybe an updated name helps clarify.

I was looking for an easy way to avoid a potential future edit to `notebook_session_log` interfering with a production use case's why.log when there is no init called. I was considering putting some flag variable there and wrapping this all in an if block something like:
`if session_is_defined():`

jamie256:
changes pulled into: https://github.com/whylabs/whylogs/pull/1285

PR #1270, opened 2023-06-05
dont add UDF columns to input dataframe

richard-rogers:
## Description

Construct a new dataframe with the UDF outputs rather than adding them to the input dataframe.


bernease (approved):
This looks fine and you hit the major points that I'm aware of as far as performance (e.g., use concat, column insertions okay but never rows).

Issue #1266, opened 2023-06-01
Scaling issues with large numbers of columns in pyspark


jamie256:
## Description

~mapInPandas udf does not support vectors natively, whylogs pyspark tries to convert these to then profile them, but~ whylogs pyspark profiling trying to process all columns at once even when there are thousands of columns, this can cause performance and stability issues on larger datasets. See https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/api/pyspark/experimental/profiler.py#L65

## Suggestions

~Let's limit the transformation of vector columns and instead skip profiling them by default so we don't block profiling larger datasets that happen to have sparse vectors. The vector processing is important for embeddings use cases but we can require this scenario be opt in with config on the interested columns.~
* Add a default batch size of the number of columns to process at a time, we can use a test dataset to look for a reasonable value using data frames with 10,000 columns on cluster size of 12 instances of c6.xlarge (8 GB, 4 CPU per instance).

- [x] I have reviewed the [Guidelines for Contributing](https://github.com/whylabs/whylogs/blob/mainline/.github/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/whylabs/whylogs/blob/mainline/.github/CODE_OF_CONDUCT.md).

PR #1273, opened 2023-06-07
pyspark batched column profiling

jamie256:
## Description

When there are larger numbers of columns the total size of the column profiles can be large enough to exert memory pressure on executor nodes, so this change introduces a default limit to the number of columns to profile at once. Some tests show that on a DataFrame containing 10k columns that batch processing up to 1024 columns at a time speeds up the overall execution on cluster size of 48 vCPUs, and should improve overall stability. The batch size is configurable.

## Changes

- Add new column_batch_size config, default value of 1024
- Add new helper method to profile columns in batches in pyspark

## Related
- Candidate fix for #1266 


python/whylogs/api/pyspark/experimental/profiler.py
```diff
@@ -75,6 +77,51 @@ def collect_column_profile_views(
...
+        combined_column_views.update(column_views_dict)
+
+    return combined_column_views
+
+
+def _get_column_batch_size(schema: Optional[DatasetSchema] = None) -> Optional[int]:
+    if schema and schema.default_configs:
+        batch_size = schema.default_configs.column_batch_size
+    else:
+        default_config = MetricConfig()
```
richard-rogers: This doesn't make sense to me as part of the `MetricConfig`. Maybe a parameter on the schema?

jamie256: moved out of MetricConfig

FelipeAdachi (approved):
LGTM!

PR #1277, opened 2023-06-09
Fix: add support for pandas 2.x

murilommen:
## Description

`collect_dataset_profile_view` is broken for users with pandas>=2.0

`pandas>=2.0` no longer supports `DataFrame.append()` method, as it does not happen inplace, so it's not the same as Python's list.append changing pyspark's api call to `pd.concat` makes it compatible with 1.x and 2.x

https://stackoverflow.com/questions/75956209/dataframe-object-has-no-attribute-append


jamie256 (approved):
LGTM!

PR #1283, opened 2023-06-20
change usage stats connection error to info log

### LINKED ISSUE #1281: Usage stats error is logged as a warning which is confusing
## Description
Usage stats errors are logged as warning which makes it seem like something with profiling is failing for the user. Since this error impacts not the user directly logging data but rather whylogs api usage statistics this should not be logged as a warning.

See: https://github.com/whylabs/whylogs/blame/mainline/python/whylogs/api/usage_stats/__init__.py#L161
```
  except:  # noqa
        logger.warning("Connection error. Skip stats collection.")
        _TELEMETRY_DISABLED = True
```

## Suggestions

Change to INFO level logging.

jamie256:
## Description
 The whylogs usage stats emits a log statement at the warning level if it cannot log the usage statistic, and the message is generic enough it sounds like an error profiling data. This is confusing at best and not necessary to log as a warning.

* change the log level to info for this issue, no need for user to take actions in this case.
* update message to be less confusing

## Related

Fixes #1281


FelipeAdachi (approved):
LGTM!

PR #1286, opened 2023-06-22
Support condition validators for rows

FelipeAdachi:
## Description

Currently, condition validators is implemented only for Pandas dfs. In this PR:

- Condition Validators now expect a PreProcessedColumn instead of a Pandas Series
- columnar_validate is called from both Column Profile's `track_column` and  `track_datum` 
- updated test

### DIFF
--- a/python/whylogs/core/column_profile.py
+++ b/python/whylogs/core/column_profile.py
@@ -56,17 +56,25 @@ class ColumnProfile(object):
             self._success_count += res.successes
 
     def track_column(self, series: Any, identity_values: Any = None) -> None:
-        for validator in self._column_validators:
-            validator.columnar_validate(series, identity_values=identity_values)
         ex_col = PreprocessedColumn.apply(series)
+        id_col = None
+        if identity_values is not None:
+            id_col = PreprocessedColumn.apply(identity_values)
+        for validator in self._column_validators:
+            validator.columnar_validate(ex_col, id_col)
         self._process_extracted_column(ex_col)
 
     def _track_homogeneous_column(self, series: Any) -> None:
         ex_col = PreprocessedColumn._process_homogeneous_column(series)
         self._process_extracted_column(ex_col)
 
-    def _track_datum(self, value: Any) -> None:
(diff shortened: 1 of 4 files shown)


python/whylogs/core/column_profile.py
```diff
@@ -56,17 +56,25 @@ def _process_extracted_column(self, extracted_column: PreprocessedColumn) -> Non
             self._success_count += res.successes
 
     def track_column(self, series: Any, identity_values: Any = None) -> None:
-        for validator in self._column_validators:
-            validator.columnar_validate(series, identity_values=identity_values)
```
jamie256: This looks like it changes what the data validators see from raw inputs to an extracted column, can we preserve the old behavior for the non-dictionary case? Its an edge case but I think you need to avoid the re-ordering that can happen in preprocessed columns when there are mixed types since we batch the types, otherwise the identity column could be off.

jamie256 (approved):
LGTM!

python/whylogs/core/dataset_profile.py
```diff
@@ -140,13 +140,11 @@ def _do_track(
...
 
         if row is not None:
-            if col_id is not None:
-                logger.warning(
-                    "validation with identity column is not supported for row level tracking.\
-                    Validation will be performed without identity column"
-                )
             for k in row.keys():
-                self._columns[k]._track_datum(row[k])
+                if col_id:
```
richard-rogers, line 144: could do `for key, value in row.items():` so you could replace `row[k]` with `value` in the loop body

python/whylogs/core/validators/condition_validator.py
```diff
@@ -52,12 +53,17 @@ def columnar_validate(self, data: Any, identity_values: Any = None) -> None:
         validate_with_row_id = False
         if self.enable_sampling and self._sampler is None:
             self._sampler = ds.var_opt_sketch(k=self.sample_size)
-        if pd.Series is not None and isinstance(identity_values, pd.Series):
```
richard-rogers, line 55: Could we tighten the type hints on `columnar_validate()`?

PR #1290, opened 2023-06-27
Fix whylabs segment writer method

murilommen:
the verification for v0 profiles was redundant 

details:
It's always converting to be a v0 profile, so it needs a small fix.

### DIFF
--- a/python/whylogs/api/writer/whylabs.py
+++ b/python/whylogs/api/writer/whylabs.py
@@ -530,7 +530,7 @@ class WhyLabsWriter(Writer):
         upload_statuses = list()
         for view, url in zip(files, upload_urls):
             with tempfile.NamedTemporaryFile() as tmp_file:
-                if kwargs.get("use_v0") is None or kwargs.get("use_v0"):
+                if kwargs.get("use_v0") is not None or kwargs.get("use_v0"):
                     view.write(file=tmp_file, use_v0=True)
                 else:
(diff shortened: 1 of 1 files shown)


python/whylogs/api/writer/whylabs.py
```diff
@@ -530,7 +530,7 @@ def _write_segmented_reference_result_set(self, file: SegmentedResultSet, **kwar
         upload_statuses = list()
         for view, url in zip(files, upload_urls):
             with tempfile.NamedTemporaryFile() as tmp_file:
-                if kwargs.get("use_v0") is None or kwargs.get("use_v0"):
+                if kwargs.get("use_v0") is not None or kwargs.get("use_v0"):
```
richard-rogers, line 533: I think the original code might be correct if `use_v0 = True` is meant to be the default. With the proposed change, calling `foo._write_segmented_reference_result_set(file, use_v0=False)` would act the same as calling `foo._write_segmented_reference_result_set(file, use_v0=True)` because `kwargs.get("use_v0") is not None` would be `True` in either case.

richard-rogers (changes requested):
I think original code is ok?

python/whylogs/api/writer/whylabs.py
jamie256: correct, but we might be able to remove this special handling for v0 format if we don't need this anymore?

PR #1294, opened 2023-06-29
Dataset-ish UDFs by column type

richard-rogers:
## Description

Attach UDFs by column type. Output column name is `{input_column_name}.{UDF name}`.

Such UDFs get the value/column as input rather than a dictionary/data frame, more like a metric UDF.

```
@register_type_udf(Fractional)
def square_type(x) -> float:
    return x * x

def test_type_udf_row() -> None:
    schema = udf_schema()
    data = {"col1": 3.14}
    results = why.log(row=data, schema=schema).view()
    assert "col1.square_type" in results.get_columns().keys()
```

## Changes

- `UdfSpec` can optionally match by type
- adds `@register_type_udf()`  (better name suggestions?)

### DIFF
--- a/python/tests/experimental/core/test_udf_schema.py
+++ b/python/tests/experimental/core/test_udf_schema.py
@@ -11,6 +11,7 @@ from whylogs.experimental.core.udf_schema import (
     UdfSchema,
     UdfSpec,
     register_dataset_udf,
+    register_type_udf,
     udf_schema,
 )
 
@@ -386,3 +387,26 @@ def test_direct_udfs() -> None:
     result = why.log(more_data, schema=schema).view()
     more_columns = set(result.get_columns())
     assert more_columns == profile_columns
+
+
+@register_type_udf(Fractional)
+def square_type(x) -> float:
+    return x * x
+
+
+def test_type_udf_row() -> None:
+    schema = udf_schema()
+    data = {"col1": 3.14}
+    print(schema.apply_udfs(row=data))
+    results = why.log(row=data, schema=schema).view()
+    print(results._columns.keys())
+    assert "col1.square_type" in results.get_columns().keys()
+
+
+def test_type_udf_dataframe() -> None:
+    schema = udf_schema()
+    data = pd.DataFrame({"col1": [3.14, 42.0]})
+    print(schema.apply_udfs(pandas=data))
+    results = why.log(data, schema=schema).view()
+    print(results._columns.keys())
+    assert "col1.square_type" in results.get_columns().keys()
--- a/python/tests/experimental/core/test_udf_schema.py
+++ b/python/tests/experimental/core/test_udf_schema.py
@@ -397,16 +397,12 @@ def square_type(x) -> float:
 def test_type_udf_row() -> None:
     schema = udf_schema()
     data = {"col1": 3.14}
-    print(schema.apply_udfs(row=data))
     results = why.log(row=data, schema=schema).view()
-    print(results._columns.keys())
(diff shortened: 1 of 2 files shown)


python/whylogs/experimental/core/udf_schema.py
```diff
@@ -198,6 +229,31 @@ def decorator_register(func):
...
+    schema_name: str = "",
+    anti_metrics: Optional[List[Metric]] = None,
+) -> Callable[[Any], Any]:
+    def decorator_register(func):
+        global _multicolumn_udfs, _resolver_specs
+        name = udf_name or func.__name__
+        name = f"{namespace}.{name}" if namespace else name
+        _multicolumn_udfs[schema_name].append(UdfSpec(None, {name: func}, col_type))
+        if metrics:
+            _resolver_specs[schema_name].append(ResolverSpec(None, col_type, deepcopy(metrics)))
```
richard-rogers: I think this will leak [anti-]metrics out onto other columns of the matched type. Since we don't know the input or output column names at registration time, we may not be able to support this...

python/tests/experimental/core/test_udf_schema.py
```diff
@@ -386,3 +387,22 @@ def test_direct_udfs() -> None:
...
+@register_type_udf(Fractional)
+def square_type(x) -> float:
+    return x * x
+
+
+def test_type_udf_row() -> None:
+    schema = udf_schema()
+    data = {"col1": 3.14}
+    results = why.log(row=data, schema=schema).view()
+    assert "col1.square_type" in results.get_columns().keys()
```
jamie256, line 401: might be worth checking if there are any metrics on this (distribution maybe?)

python/whylogs/experimental/core/udf_schema.py
jamie256: Is the same true for the metrics then?

do we need the anti_metrics on this decorator? Can't they also do that type->metric binding in the schema? I suppose that means generating the schema is trickier but it might be an acceptable escape hatch for this more advanced scenario.

jamie256 (dismissed):
Looks good! I left some minor comments/questions.

python/whylogs/experimental/core/udf_schema.py
richard-rogers: > Is the same true for the metrics then?

Assuming you mean metric UDFs, no. The [anti-]metrics registered there apply only to the submetric created by the UDF. Regular dataset UDFs are also OK, since the output column name is known at registration time. 

> do we need the anti_metrics on this decorator? 

I'm going to remove them. For this sort of UDF, you'll have to explicitly manage your resolver specs if you want something other than the unadulterated schema.

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -32,20 +32,26 @@
 @dataclass
 class UdfSpec:
     """
-    Defines UDFs to apply to matching input columns. Tthe UDF is
+    Defines UDFs to apply to matching input columns. The UDF is
```
FelipeAdachi: this docstring should be updated to include matching by column type

jamie256 (dismissed):
LGTM!

python/whylogs/experimental/core/udf_schema.py
richard-rogers: Yes, I will do the docstrings last once I'm sure the API is fully baked :)

jamie256 (dismissed):

jamie256 (approved):

PR #1298, opened 2023-07-06
whylabs writer - tag column as custom performance metric

FelipeAdachi:
## Description

adds a method to WhyLabsWriter to tag a column as custom performance metrics to be shown in the Performance tab in WhyLabs UI

### DIFF
--- a/python/whylogs/api/writer/whylabs.py
+++ b/python/whylogs/api/writer/whylabs.py
@@ -25,6 +25,7 @@ from whylabs_client.model.column_schema import ColumnSchema
 from whylabs_client.model.create_reference_profile_request import (
     CreateReferenceProfileRequest,
 )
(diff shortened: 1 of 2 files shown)


python/whylogs/api/writer/whylabs.py
```diff
@@ -446,6 +447,40 @@ def tag_input_columns(self, columns: List[str]) -> Tuple[bool, str]:
 
         return self._tag_columns(columns, "input")
 
+    def tag_custom_performance_column(self, column: str, label: str, default_metric: str) -> Tuple[bool, str]:
```
jamie256: Can we make the label and default_metric optional with a default set here?

jamie256 (dismissed):

jamie256 (approved):
LGTM!

PR #1307, opened 2023-07-13
UDF signature update & Python data types

richard-rogers:
## Description

Changes dataset and type UDF signatures so that they can be coded to ignore whether they're called with a row or `DataFrame`. Also allows type-matching `ResolverSpec`s to use Python types rather than whylogs `DataType`.

## Changes

- Dataset UDF signature is `udf(x: Union[Dict[str, List], pd.DataFrame]) -> Union[List, pd.Series]`
- Type UDF signature is `udf(x: Union[List, pd.Series]) -> Union[List, pd.Series]`
- UNCHANGED: metric UDF signature remains `udf(x: Any) -> Any`, where the Anys are single values in a row.
- `UdfSpec` and `ResolverSpec` matching by type can now use native Python types via a `TypeMapper` in addition to the whylogs `DataType`

### DIFF
--- a/python/tests/experimental/core/test_udf_schema.py
+++ b/python/tests/experimental/core/test_udf_schema.py
@@ -82,7 +82,7 @@ def test_decorator_row() -> None:
 
 @register_dataset_udf(["col1"], "annihilate_me", anti_metrics=[CardinalityMetric, DistributionMetric])
 def plus1(x: Union[Dict[str, List], pd.DataFrame]) -> Union[List, pd.Series]:
-    return x["col1"] + 1 if isinstance(x, pd.DataFrame) else [xx + 1 for xx in x["col1"]]
+    return x["col1"] + 1 if isinstance(x, pd.DataFrame) else map(lambda i: i + 1, x["col1"])
 
 
 def test_anti_resolver() -> None:
--- a/python/tests/experimental/core/test_udf_schema.py
+++ b/python/tests/experimental/core/test_udf_schema.py
@@ -46,11 +46,11 @@ def test_udf_pandas() -> None:
 
 @register_dataset_udf(["col1"])
 def add5(x):
-    return x["col1"] + 5
+    return [xx + 5 for xx in x["col1"]]
 
 
 def square(x):
-    return x["col1"] * x["col1"]
+    return x["col1"] * x["col1"] if isinstance(x, (pd.Series, pd.DataFrame)) else [xx * xx for xx in x["col1"]]
 
 
 def test_decorator_pandas() -> None:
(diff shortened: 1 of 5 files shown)


python/whylogs/experimental/core/udf_schema.py
```diff
@@ -263,6 +268,9 @@ def register_type_udf(
     it will be registered to the defualt schema.
 
     """
+    if not issubclass(col_type, DataType):
+        type_mapper = type_mapper or StandardTypeMapper()
+        col_type = type(type_mapper(col_type))
```
richard-rogers: Needs error checking.

python/tests/experimental/core/test_udf_schema.py
```diff
@@ -46,11 +46,11 @@ def test_udf_pandas() -> None:
 
 @register_dataset_udf(["col1"])
 def add5(x):
```
richard-rogers:
```suggestion
def add5(x: Union[Dict[str, List], pd.DataFrame]) -> Union[List, pd.Series]:
```
Type hint UDF signatures

python/tests/experimental/core/test_udf_schema.py
```diff
@@ -390,8 +390,8 @@ def test_direct_udfs() -> None:
 
 
 @register_type_udf(Fractional)
-def square_type(x) -> float:
-    return x * x
+def square_type(x):
```
richard-rogers:
```suggestion
def square_type(x: Union[List, pd.Series]) -> Union[List, pd.Series]:
```

python/tests/experimental/core/test_udf_schema.py
```diff
@@ -80,7 +80,7 @@ def test_decorator_row() -> None:
 
 @register_dataset_udf(["col1"], "annihilate_me", anti_metrics=[CardinalityMetric, DistributionMetric])
 def plus1(x):
-    return x["col1"] + 1
+    return x["col1"] + 1 if isinstance(x, pd.DataFrame) else [xx + 1 for xx in x["col1"]]
```
richard-rogers: `map()` the list?

python/whylogs/core/resolvers.py
```diff
@@ -110,17 +118,21 @@ class ResolverSpec:
     """
 
     column_name: Optional[str] = None  # TODO: maybe make this a regex
-    column_type: Optional[DataType] = None
+    column_type: Optional[Union[DataType, Type]] = None
```
richard-rogers:
```suggestion
    column_type: Optional[Type] = None
```
Though technically `DataType <: Type`, the union might be helpful as a reminder that `DataType` is a special case.

richard-rogers: Actually, `DataType` would imply an instance of the `DataType` class, but we mean a subclass of `DataType`. I think that's problematic across the range of Python versions we support, so perhaps `Type` after all...

FelipeAdachi (approved):
I didn't take a deep look into the code itself, but tested the branch out with some scenarios on both rows and dfs, and didn't hit into any errors.

jamie256 (approved):
LGTM!

PR #1311, opened 2023-07-14
Several changes to support why.init for authenticated users

naddeoa:
- Add org id and default dataset id support to sessions.
- Make sessions depend on the new api key format and update the api key validation to allow for longer api keys that include `:orgId`.
- Remove explicit session type configuration and define a fallback system for selecting session type based on `allow*` flags, config, and env state.
- Allow specifying default dataset directly to the session config as an alternative to env variables, which is how sessions treat api keys.
- Add interactive prompt mode for notebook environments that gets api key and related config via `input()` and saves it to the new whylogs ini config.
- Add two new files as pseudo CLIs for setting api key/default model id in the whylogs config. Without this, the only way to set the config is via interactive prompt. Supplying that info via env vars doesn't result in persisting that to the config which would be a strange side effect at best since the env is presumably still around and definitely taking priority over the config value.
- Enable interactive log output and url generation for batch profiles, with or without authentication. This output is disabled in local mode since no uploading happens there. The output only happens when using the top level transient logger in `why.log`. The loggers themselves don't have any output logic at all. It wouldn't make much sense to add output of this sort to the rolling logger running in the background.
- Add lots of docs to `why.init` since that's the main interface to this system for users. The goal with these docs is to make it clear what kind of session you'll be getting since the logic is a little more complicated than before. The upside to this approach is that most notebooks, scripts, and production jobs can just have `why.init()` and not worry about what arguments to supply, so it can truly be ubiquitous.

## Reviewing

To get a sense of the UX introduced here, I recommend running the Guest Session example notebook, as well as running `python -m whylogs.api.whylabs.session.why_init` from the command line. The first one should be pretty representative of a typical user experience after this change and the second one is a psuedo-cli that users can use to repeat the first UX without having to spin up a notebook. The command is similar to running `aws sso configure`, it just gets everything it needs from a user and stores it locally in the new whylogs.ini config.


python/whylogs/api/whylabs/session/config.py
```diff
@@ -29,19 +47,48 @@ class ConfigVariableName(Enum):
...
     WHYLABS_API_ENDPOINT = "whylabs_api_endpoint"
     USER_GUID = "user_guid"
+    DEFAULT_DATASET_ID = "default_dataset_id"
+    WHYLABS_REFERENCE_PROFILE_NAME = "whylabs_reference_profile_name"
+    WHYLABS_PRIVATE_API_ENDPOINT = "whylabs_private_api_endpoint"
+    WHYLABS_PRIVATE_S3_ENDPOINT = "whylabs_private_s3_endpoint"
+
+
+@dataclass
+class InitConfig:
```
andyndang, line 58: can we add a documentation here about how these variables interact?

naddeoa: The documentation is on the `init` function in `session.py`. That's the interface that people will actually see. This thing is just an intermediate object passed into the config based on the arguments to `init`. No one will ever actually see this config type. I have a fallback list in the `init` docs so people know what to expect.

naddeoa:
I'm currently working on a doc page for this new auth system on docs.whylabs.ai. I link to it in several places since the logic can be pretty verbose. It will be here when I'm done and we should hold off on merging until I finish `https://docs.whylabs.ai/docs/whylabs-whylogs-init`

python/whylogs/api/whylabs/session/config.py
```diff
@@ -3,24 +3,43 @@
...
+from whylogs.api.whylabs.session.session_types import InteractiveLogger as il
+from whylogs.api.whylabs.session.session_types import (
+    SessionType,
+    parse_api_key,
+    parse_api_key_v2,
+)
 
 _DEFAULT_WHYLABS_HOST = "https://api.whylabsapp.com"
 _CONFIG_APP_NAME = "whylogs"
+_INIT_DOCS = "https://docs.whylabs.ai/docs/whylabs-whylogs-init"
```
naddeoa, line 30: I'm currently working on a doc page for this new auth system on docs.whylabs.ai. I link to it in several places since the logic can be pretty verbose. It will be here when I'm done and we should hold off on merging until I finish `https://docs.whylabs.ai/docs/whylabs-whylogs-init`

naddeoa: Feel free to comment on the url and side bar location 
image

naddeoa: Draft of the new doc page: https://64b5bca8f6fdda0d2fff5639--whylabs-docs-dev.netlify.app/docs/whylabs-whylogs-init

naddeoa:
@jamie256 @andyndang Looks like there are 36 different example notebooks that included WHYLABS_API_KEY. All of those will have to be updated. I'm going to hold off on touching those until I know this PR and the new doc page look good though. I have to take an inventory of the things that need to change on the doc site as well.

FelipeAdachi (commented):
I just ran the `Writing_Regression_Performance_Metrics_to_WhyLabs` without any changes on this branch, and some things to note:

- There are some rather spammy messages, when logging:
   - `😭 No active session found because why.init() wasn't used. Skipping reference profile upload. Start a session with why.init(session_type='whylabs_anonymous')
writing profiles to whylabs...`
- and also when calling `results.writer("whylabs").write()`:
   - `Initialing session with config /home/felipeadachi/.config/whylogs/config.ini
Initialing session with config /home/felipeadachi/.config/whylogs/config.ini
No session is initialized. Call whylogs.init(anonymous=True) to create a guest session.
No session is initialized. Call whylogs.init(anonymous=True) to create a guest session.`

This feels like something wrong is happening, but I'm just logging the usual way by setting `os.environ` (and it works normally)

Maybe we could set this as info instead of warning?
Or is this now an unrecommended approach, and we always want to call `why.init()`?

richard-rogers (commented):
LGTM

python/examples/integrations/Guest Session.ipynb
FelipeAdachi, line 1: delete first block with commented dev endpoint?

naddeoa:
> I just ran the `Writing_Regression_Performance_Metrics_to_WhyLabs` without any changes on this branch, and some things to note:
> 
>     * There are some rather spammy messages, when logging:
>       
>       * `😭 No active session found because why.init() wasn't used. Skipping reference profile upload. Start a session with why.init(session_type='whylabs_anonymous') writing profiles to whylabs...`
> 
>     * and also when calling `results.writer("whylabs").write()`:
>       
>       * `Initialing session with config /home/felipeadachi/.config/whylogs/config.ini Initialing session with config /home/felipeadachi/.config/whylogs/config.ini No session is initialized. Call whylogs.init(anonymous=True) to create a guest session. No session is initialized. Call whylogs.init(anonymous=True) to create a guest session.`
> 
> 
> This feels like something wrong is happening, but I'm just logging the usual way by setting `os.environ` (and it works normally)
> 
> Maybe we could set this as info instead of warning? Or is this now an unrecommended approach, and we always want to call `why.init()`?

This has been fixed. You'll now only see the warning one time and you won't see any other output unless you actually have a session, so if people run that notebook before we update it with our latest whylogs then they'll just see a warning that they don't have a session set up.

python/whylogs/api/whylabs/session/config.py
```diff
@@ -29,19 +48,48 @@ class ConfigVariableName(Enum):
...
+class InitConfig:
+    whylabs_api_key: Optional[str] = None
+    allow_anonymous: bool = True
+    allow_local: bool = False
+    default_dataset_id: Optional[str] = None
 
 
 class SessionConfig:
-    def __init__(self) -> None:
+    def __init__(self, init_config: InitConfig = InitConfig()) -> None:
```
jamie256: Is it possible to have more than one session in a process or switch sessions for authenticated users?

You might want to change the default init_config value to None, and then in the body check for None and create an instance of InitConfig() inside the __init__ method if so to keep separate instances of the InitConfig.

python/whylogs/api/whylabs/session/session_manager.py
```diff
@@ -259,31 +61,83 @@ def is_active() -> bool:
...
 
 
-def init(session_type: Union[SessionType, str] = SessionType.LOCAL, reinit: bool = False) -> None:
+def init(
+    reinit: bool = False,
+    allow_anonymous: bool = True,
+    allow_local: bool = False,
+    whylabs_api_key: Optional[str] = None,
+    default_dataset_id: Optional[str] = None,
+) -> None:
```
jamie256, line 70: This is a breaking change in the signature of why.init

Is it difficult to maintain the old session_type parameter so we don't break the previous callers that had this:
`why.init(session_type='whylabs_anonymous')`

A known impact would be in the LangKit example notebooks, e.g. we would need to update the init if we update the whylogs dependency. See: https://github.com/whylabs/langkit/blob/main/langkit/examples/Intro_to_Langkit.ipynb

jamie256 (commented):
Looks close, left two questions.

python/whylogs/api/whylabs/session/session_manager.py
naddeoa: This might be a bit of a pickle. The original idea for getting rid of this came from @andyndang. This was the only way to make it so that you can almost always just use `why.init()` and the right thing will happen automatically.

I can add some behavior that tells the user that the argument no longer exists and what to do instead. Is that what you were getting at? We would also have to add the llm docs to the very long list of docs to update after this PR.

python/whylogs/api/whylabs/session/config.py
naddeoa: The default InitConfig is there really just to establish good default values for `why.init()` or for manually creating the SessionConfig (like I'm doing in the whylabs writer). Are you worried about someone mutating the default InitConfig object after its created? I can convert this into a field with a factory default instead.

python/whylogs/api/whylabs/session/session_manager.py
jamie256: LangKit pins the whylogs dependency, so we can update the init in the notebook examples as we update the dependency on whylogs.

I wanted to call attention to the breaking change so it was understood and intentional and we could reason about the risks and mitigations.

jamie256 (approved):
LGTM!

PR #1312, opened 2023-07-17
WhyLabs Writer - tag known custom performance metrics

FelipeAdachi:
## Description

With this change, any profile being written to WhyLabs will check for the existence of columns that match known custom performance metrics and, if present, will tag them as custom performance columns.


python/whylogs/api/writer/whylabs.py
```diff
@@ -68,6 +68,16 @@
 _S3_PUBLIC_DOMAIN = os.environ.get("_WHYLABS_PRIVATE_S3_DOMAIN") or _US_WEST2_DOMAIN
 _WHYLABS_SKIP_CONFIG_READ = os.environ.get("_WHYLABS_SKIP_CONFIG_READ") or False
 
+KNOWN_CUSTOM_PERFORMANCE_METRICS = {
```
jamie256, line 71: we might call these known ranking metric names, but lets follow up and make these read form profile metadata, and mark the metadata when the ranking metrics are computed?

jamie256 (approved):
LGTM!

PR #1315, opened 2023-07-18
only pass necessary columns to row UDF

richard-rogers:
Should only use the columns referenced by the UDF rather than all columns.


FelipeAdachi (approved):
LGTM!

PR #1318, opened 2023-07-21
type handling - add np.integer to int types

FelipeAdachi:
## Description

The motivation for this change is that the following code snippet would output empty distribution metrics for `udf/add5`:

```python
@register_metric_udf(col_name="col1")
def add5(x):
    return x + 5

df = pd.DataFrame({"col1": [1, 2, 3, 4, 5]})

profile = why.log(df,schema=udf_schema()).view()
```

because the values were of type `np.int64`, which evaluated to false when doing `isinstance(value,int)`

This pr:

- adds `np.integer` to the integer check when processing the scalar values
- creates stub for `np.integer`
- adds test to check for this case


jamie256 (approved):
LGTM!

PR #1323, opened 2023-07-28
Add new process logger and thread logger

naddeoa:
This PR builds off of the thread based logger that was already in the experimental name space, which was based on the logger that I had added to the whylogs container. It essentially ports the container logger out for use in whylogs as a normal library, and refactored the common parts out of the thread/process version so they share the same, actor based, super class.

They process and thread version behave slightly differently and they have slightly different use cases.

## Process Logger:
- Capable of handling profile data for multiple datasets and multiple timestamps.
- Runs as a separate process and defers to instances of the thread based logger to handle all of the actual logging.
- Enforces a notion of dataset id. Can almost be ignored if the whylabs writer isn't the preferred writer, but some id will have to be used.
- Most behavior is inherited from the thread logger as far as logging is concerned since it defers entirely to the thread logger.
- Supports synchronous logging if needed, but it has to be explicitly enabled since it has more overhead than thread synchronous logging because it depends on IPC. It might also interfere with batch accumulation which hits performance. This is mostly provided for debugging and testing.

## Thread Logger:
- Capable of handling multiple timestamps, but only for a single dataset id. It doesn't enforce any notion of dataset id.
- Can automatically flush internal state to known writers on a timer.
- Supports synchronous or asynchronous logging, but you'll probably want asynchronous.

## vs Current Rolling Logger
Using either of them has a few advantages over the current rolling logger.
- Serialized actions via the actor interface. This means that there won't be any concurrency issues that lead to data frame tearing while the logger is flushing/writing while the caller is sending messages, and it won't block the caller unless the message queue is full.
- Support for historical data. log() takes a timestamp now and the loggers will manage the profiles for each timestamp, which includes timestamp truncation and other tedious things.
- Bulk processing of messages to minimize internal track() calls. whylogs performs way better when you call it fewer times with more data. The process logger will handle automatically grouping data and merging them into large data frames for more efficient profiling. This behavior isn't currently in the thread logger but it can be refactored. I held off because the thread logger's performance is inherently capped by the gil anyway so the process logger will be inevitable for anyone who needs performance.

I think we should retire the current Logger interface now as well. It's already insufficient for the logging options that we have because the interface was updated to include properties from each of the loggers in order to make the types work out. It also uses kwargs for everything so its very difficult to determine what options even exist without looking at the source. We should just let people create the logger that they want with normal python.

This PR can't be pushed until the new session logic is pushed because it depends on the session to correctly configure the whylabs writer. We're also going to have a fair bit of documentation to update. Once this is approved.


python/examples/basic/Getting_Started.ipynb
FelipeAdachi, line 1: - why do we need `pmdarima` for this?
- let's clear the output from pip install, and maybe add -q?
- there are a bunch of comments that need to be removed
- the output of `summary_drift_html` is not rendered correctly, I think.

naddeoa: Those changes were just left over from our hackathon actually, I didn't mean to include them. They're out now.

python/whylogs/api/fugue/__init__.py
```diff
@@ -1,6 +1,4 @@
 # flake8: noqa
 from whylogs.api.usage_stats import emit_usage
 
-from .profiler import fugue_profile
```
murilommen: why did you have to delete this? can't it break fugue usage?

python/whylogs/api/whylabs/session/session_manager.py
```diff
@@ -211,6 +226,9 @@ def __init__(self, config: SessionConfig) -> None:
         if is_notebook():
             log_if_notebook(f"Initialized whylabs session with for org {self.org_id}")
 
+    def get_config(self) -> SessionConfig:
+        return self._config
```
murilommen: if all child classes from `Session` implement it on the same way, wouldn't it be best to leave it as a default implementation of `Session` and not repeat the code?

python/whylogs/api/logger/experimental/logger/actor/thread_rolling_logger.py
```diff
@@ -237,22 +242,26 @@ def __init__(
...
+        elif batch_type == CloseMessage:
+            self._process_close_messages(cast(List[CloseMessage], batch))
+        elif batch_type == StatusMessage:
+            self._process_status_messages(cast(List[StatusMessage], batch))
+        elif batch_type == GetResultsMessage:
+            self._process_get_results_messages(cast(List[GetResultsMessage], batch))
         else:
             # Safe guard for forgetting to handle a message in development
-            raise Exception(f"Don't know how to handle message {message}")
+            raise Exception(f"Don't know how to handle message {batch_type}")
```
murilommen, line 260: wouldn't it be good to add a `logger.error` here too?

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+class DataDict(TypedDict):
+    columns: List[str]
+    data: List[List[DataTypes]]
+
+
+def data_dict_from_pandas(df: pd.DataFrame) -> DataDict:
+    multiple = df.to_dict(orient="split")
+    del multiple["index"]  # get rid of this to min/max payload size
```
murilommen, line 56: i think it's better to `multiple.drop("index", axis="columns")`

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
```diff
@@ -0,0 +1,283 @@
...
+    return {
+        "columns": multiple["columns"],
+        "data": multiple["data"],
+    }
+
+
+class LogRequestDict(TypedDict):
+    datasetId: str
```
murilommen, line 64: any reason not to use snake_case here?

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
```diff
@@ -0,0 +1,283 @@
...
+    timestamp: Optional[int]
+    multiple: DataDict
+
+
+class LogEmbeddingRequestDict(TypedDict):
+    datasetId: str
```
murilommen, line 70: same

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+@dataclass
+class RawLogMessage:
+    request: bytes
+    """
+    Bytes that represent json stringified LogRequestDict
+    """
+    request_time: int
+    id: str = "None"
```
murilommen: why not `id = field(default=None)` + `__post_init__`? little weird to have a string "None" in all of them

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
```diff
@@ -0,0 +1,283 @@
...
+        return data
+
+
+class PubSubEmbeddingDict(TypedDict):
+    subscription: str
+    message: PubSubMessage
+    log_embedding_request: LogEmbeddingRequestDict
+
+
+def _dencode_pubsub_data(data: Union[PubSubEmbeddingDict, PubSubDict]) -> Optional[bytes]:
```
murilommen: typo? decode

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
```diff
@@ -0,0 +1,283 @@
...
+
+        acc["embeddings"][col].extend(embeddings)
+
+    return acc
+
+
+def determine_dataset_timestamp(
+    cadence: TimeGranularity, request: Union[LogRequestDict, LogEmbeddingRequestDict]
+) -> Optional[int]:
+    ts = request["timestamp"]
```
murilommen, line 278: it will break if "timestamp" is actually `None`. better with `request.get("timestamp")`

murilommen: from what I saw, you can define a `LogRequestDict` without a timestamp, since it's optional

murilommen (commented):
overall this is a great addition to whylogs. my comments were more focused on documentation and small optional changes

python/whylogs/api/logger/experimental/logger/actor/actor.py
```diff
@@ -0,0 +1,210 @@
...
+
+class CloseMessage:
+    pass
+
+
+@dataclass
+class QueueConfig:
+    max_batch_size: int = 50_000
+    """
+    Maximum number of messages to process in a single batch.
```
murilommen: this should follow a [sphynx-like documentation](https://github.com/whylabs/whylogs/blob/mainline/python/whylogs/api/writer/whylabs.py#L169), as these strings won't appear on a `help(QueueConfig)` properly for example. i.e.:

```python
class QueueConfig:
    """
    This is the interface for (...) and contains (...)

    :param max_batch_size: Maximum number of messages to process in a single batch, defaults to 50_000
    :type max_batch_size: int

    :raises [ErrorType]: [ErrorDescription]

    :return: [ReturnDescription]
    :rtype: [ReturnType]
    """

    max_batch_size: int = 50_000
    (...)
```

python/whylogs/api/logger/experimental/logger/actor/data_logger.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @abstractmethod
+    def log(
+        self,
+        data: TrackData,
+        timestamp_ms: Optional[int] = None,  # The timestamp that the data happened at
+        sync: bool = False,
+    ) -> None:
+        """
+        Log some data.
```
murilommen, line 34: nice

python/whylogs/api/logger/experimental/logger/actor/data_logger.py
```diff
@@ -0,0 +1,58 @@
...
+        Flush the internal state, causing everything to be written using the configured writers.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def close(self) -> None:
+        """
+        Close the logger, causing all resources to be released.
+        """
+        raise NotImplementedError()
```
murilommen, line 58: always add a line-break on the end of files. `fix-format` should take care of it though

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger.py
```diff
@@ -0,0 +1,473 @@
...
+        self.current_time_ms = current_time_fn or current_time_ms
+        self.loggers: Dict[str, ThreadRollingLogger] = {}
+        self.debug_writers = debug_writers
+        self.write_schedule = write_schedule
+        self.schema = schema
+        self.aggregate_by = aggregate_by
+        self._pipe_signaler: Optional[PipeSignaler] = PipeSignaler() if sync_enabled else None
+        self._session = get_current_session()
+
+    def create_writers(self, dataset_id: str) -> List[Writer]:
```
murilommen: shouldn't it be possible to create_writers to other locations? and have WhyLabs as default

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger.py
```diff
@@ -0,0 +1,473 @@
...
+        self._logger.info(f"Created logger for {dataset_id}")
+        return logger
+
+    def _get_logger(self, dataset_id: str) -> ThreadRollingLogger:
+        if dataset_id not in self.loggers:
+            self.loggers[dataset_id] = self._create_logger(dataset_id)
+        return self.loggers[dataset_id]
+
+    def process_batch(self, batch: List[MessageType], batch_type: Type) -> None:
+        if batch_type == FlushMessage:
```
murilommen, line 157: it could be simpler to infer isinstance(batch[0]) instead of parsing and passing the type, but maybe this performs better? curious on the decision here only

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger.py
```diff
@@ -0,0 +1,473 @@
...
+        for datasetId, logger in self.loggers.items():
+            self._logger.info(f"Closing whylogs logger for {datasetId}")
+            logger.close()
+
+        if self._pipe_signaler is not None:
+            self._pipe_signaler.close_child()
+
+    def process_pubsub(self, messages: List[RawPubSubMessage]) -> None:
+        self._logger.info("Processing pubsub message")
+        msgs = [msg["log_request"] for msg in [it.to_pubsub_message() for it in messages] if msg is not None]
```
murilommen, line 188: if msg is not None is there a guarantee that it will contain the `log_request` field? it appears that it will only be created after `to_pubsub_message()` is called. trying to think if there is a case where it won't get called

`.get()` could be less restrictive and exception prone, but could also lead to an empty msg item being passed over to `process_log_dicts`

python/whylogs/api/logger/experimental/logger/actor/thread_actor.py
```diff
@@ -0,0 +1,91 @@
...
+                else:
+                    break
+
+        return messages
+
+    def size(self) -> int:
+        return self._queue.qsize()
+
+
+ThreadMessageType = TypeVar("ThreadMessageType")
```
murilommen, line 57: i would move this up top

python/whylogs/api/fugue/__init__.py
naddeoa: Removed this. I forgot what I was doing that resulted in this.

python/whylogs/api/logger/experimental/logger/actor/thread_rolling_logger.py
naddeoa: That's done in the actor class where this thing is called. It would just be double logged if I did that here.

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
naddeoa: Better in what way?

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
naddeoa: These *request classes were originally created in the container to serve as the container (lul) for the data that is sent in over http as json, which is done in camel case. That's actually still what these are kind of used for even though there's no rest server here. The messages from the main process are serialized to json via orjson, and eventually they'll be hooked up to rest requests as well, so might as well just keep them camel case.

Users won't actually use this directly anyway, they only interact with the log() api on the process/thread logger.

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
naddeoa: default wouldn't have worked because it makes the type optional, which I can't have. But there's also a default_factory which was exactly what I needed.

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
naddeoa: It doesn't break! It would only break if it were an untyped dict and the key didn't exist.  Added a test to confirm. A None value is still associated with the key.

python/whylogs/api/logger/experimental/logger/actor/actor.py
naddeoa: python...

python/whylogs/api/logger/experimental/logger/actor/data_logger.py
naddeoa: You can thank copilot for that one

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger.py
naddeoa: You can via subclassing (how I do it in the tests). But I just replaced that with a WriterFactory pattern because it lets me get rid of the debug_writer thing as well as specify w/e writer you might need. So now you just pass in a factory and the default factory is WhyLabs.

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger.py
naddeoa: This has to do with the batching logic. There isn't a great way to represent this in the type system though.

The batching logic in the super class guarantees that you'll only get subtypes of a particular message and it explicitly tells the subclass which type it is via the batch_type. I felt that it would be a little to squishy/implicit to just assume that in the sub types and check an arbitrary element's type. There are a few cases that wouldn't cover, like the empty list (no current use case) and cases where messages are subtypes of each other.

And yeah there is no runtime typecheck here from a value, just an equality check, but I wasn't super concerned about perf even if that wasn't true.

This is kind of ugly though.

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger.py
naddeoa: Similar answer here to the previous one about get(). I'm using typed dicts for all of these types and the contract is that the type definitions are true, which I can enforce myself since these are all internal types that I make for myself.

python/whylogs/api/logger/experimental/logger/actor/thread_actor.py
naddeoa: Shouldn't it be right next to the place its actually used? Nothing else would really care about this generic class token right? If this were a real language it wouldn't even exist, it would just be `T` in the class def.

python/whylogs/api/whylabs/session/session_manager.py
naddeoa: Going to leave issues about the session stuff for the other pr. This was actually just a stub I put in to make the types work before I have the authenticated sessions merged. I need to rebase off of that PR and update the types to the latest way I do things there before I can merge this.

python/pyproject.toml
```diff
@@ -57,6 +57,10 @@ google-cloud-storage = {version = "^2.5.0", optional = true}
 pyarrow = {version = ">=8.0.0, <13", optional = true}
 pyspark = {version = "^3.0.0", optional = true}
 
+# process logging dependencies
+faster-fifo = {version = "^1.4.5", optional = true}
```
jamie256, line 61: Can we file a github issue in whylogs for following up on this dependency to see if we can get something better than a source only distribution?

python/whylogs/api/logger/experimental/logger/actor/faster_fifo_queue_wrapper.py
```diff
@@ -0,0 +1,38 @@
+try:
+    from faster_fifo import Queue as FasterQueue  # type: ignore
+except ImportError:
+    raise ImportError("Install whylogs with extra [proc] for process based logging: pip install whylogs[proc]")
```
jamie256: Can we hint about the alternative here? I expect some users will be unable to install whylogs[proc] even if they want to because of platform issues or inability to build the source distribution.

python/whylogs/api/logger/experimental/logger/actor/faster_fifo_queue_wrapper.py
```diff
@@ -0,0 +1,38 @@
...
+from typing import Generic, List, Optional, TypeVar
+
+from whylogs.api.logger.experimental.logger.actor.actor import QueueWrapper
+
+_DEFAULT_QUEUE_SiZE = 1000 * 1000 * 1000
```
jamie256, line 17: Any comments on the choice of default queue size? what happens if we hit the limit?

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
```diff
@@ -0,0 +1,282 @@
...
+from typing import Dict, List, Optional, Tuple, Union, cast
+
+from whylogs.api.logger.experimental.logger.actor.thread_rolling_logger import (
+    LoggerStatus,
+)
+
+try:
+    import orjson
+except ImportError:
+    raise ImportError("Install whylogs with extra [proc] for process based logging: pip install whylogs[proc]")
```
jamie256, line 32: or whylogs[proc-mp]

python/whylogs/api/writer/whylabs.py
```diff
@@ -107,7 +107,7 @@ def _check_whylabs_condition_count_uncompound() -> bool:
     return True
 
 
-# TODO update this avliator for new format
+# TODO update this valiator for new format
```
jamie256: almost
```suggestion
# TODO update this validator for new format
```

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
```diff
@@ -0,0 +1,282 @@
...
+
+import numpy as np
+import pandas as pd
+
+from whylogs.api.logger.experimental.logger.actor.time_util import (
+    TimeGranularity,
+    truncate_time_ms,
+)
+
+DataTypes = Union[str, int, float, bool, List[float], List[int], List[str]]
```
jamie256, line 44: What happens when a pandas column is of object type? e.g. Decimal values are supported by whylogs profiling and are dtype==object

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+class PubSubDict(TypedDict):
+    subscription: str
+    message: PubSubMessage
+    log_request: LogRequestDict
+
+
+class DebugMessage:
+    pass
```
jamie256, line 89: Is this a placeholder for a future update that will add some more fields?

python/whylogs/api/logger/experimental/logger/actor/thread_rolling_logger.py
```diff
@@ -350,25 +373,31 @@ def _process_track_message(self, message: TrackMessage) -> None:
...
             profile_container.track(data)
             if message.result is not None:
                 message.result.set_result(None)
         except Exception as e:
             if message.result is not None:
                 message.result.set_exception(e)
 
-    def _status(self) -> LoggerStatus:
-        result: Future[LoggerStatus] = Future()
+    def status(self, timeout: Optional[float] = 1.0) -> LoggerStatus:
```
jamie256: consider defining a few constants for the default timeouts so its easier to change these later if we have to, I see there are a few other places that use 0.1 for example.

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
```diff
@@ -0,0 +1,282 @@
...
+@dataclass
+class ProcessLoggerStatus:
+    dataset_id: str
+    status: LoggerStatus
+
+
+@dataclass
+class ProcessLoggerStatusMessage:
+    id: str = field(default_factory=lambda: str(uuid.uuid4()))
+
```
jamie256, line 105: I'm not sure I follow the semantics of the different message types and how they relate, can you explain this one a bit?

python/whylogs/api/logger/experimental/logger/actor/mp_quueue_wrapper.py
jamie256, line 1: nit: filename typo?
mp_quueue_wrapper.py -> mp_queue_wrapper.py

python/whylogs/api/logger/experimental/logger/actor/actor.py
```diff
@@ -0,0 +1,232 @@
...
+            self.set_closed()
+
+        done = False
+        while not done:
+            try:
+                self._queue.send(message, timeout=self._queue_config.message_poll_wait)
+                done = True
+            except queue.Full:
+                current_size = self._queue.size()
+                if current_size == 0:
```
jamie256, line 144: If the queue is full, I would expect that the size() would return something above 0, is this code correct?

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
```diff
@@ -0,0 +1,282 @@
...
+    row: Dict[str, np.ndarray] = {}
+    row_count = 0
+    for col, embeddings in request["embeddings"].items():
+        row[col] = np.array(embeddings)
+        row_count += len(embeddings)
+
+    return row, row_count
+
+
+def reduce_log_requests(acc: LogRequestDict, cur: LogRequestDict) -> LogRequestDict:
```
jamie256, line 256: With respect to function name: this is more like a 'concatenation' (or collect, or extend) than a 'reduce'.

Where is this function used? I'm trying to reason about the disjoint columns comment but I'm not sure I understand the scenario when this would be called.

jamie256 (commented):
Still looking and trying to evaluate this, but here are some first pass comments.

python/pyproject.toml
naddeoa: https://github.com/whylabs/whylogs/issues/1340

python/whylogs/api/logger/experimental/logger/actor/faster_fifo_queue_wrapper.py
naddeoa: It was pretty arbitrary. I have todos to make it configurable. I can easily add an env var if you think that isn't premature. I just picked a size that seemed reasonable in most cases (from my container testing). When the queue fills up then things start to back up. If the queue is full then the call to `log()` (and `send()` underneath the hood) will end up blocking for the queue to have enough room again.

The bigger the size the more throughput is required to max it out. I went  bigger rather than small to minimize the main  process blocking.

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
naddeoa: The message types are what determine what happens once they're delivered to the actor. The actor checks the message's type and then does _something_ with it. In the case of the `ProcessLoggerStatusMessage`, the actor will lookup the current internal state. For logging, the RawLogMessage has all of the data that has to be logged and the actor will use it to do actual logging.

They're mostly details of message passing. The user doesn't have to care about these. They can interract with the logger though the DataLogger interface which does the heavy lifting of sending the message and coordinating across processes. The process actor is both an Actor and a DataLogger, so we can expose it casted as a DataLogger if you want to hide all of this stuff. And its an actor of  course, so it processes messages serially from its queue.

python/whylogs/api/logger/experimental/logger/actor/actor.py
naddeoa: The else branch is doing what you expect. This case only happens when the message is both full and effectively empty, which means it was configured with a max buffer size that is smaller than the messages that its trying to receive. This is how that shows up, at least in the faster-fifo lib. I've never actually encountered this outside of testing for it though. I added a bit more to the message to reflect that its hopeless.

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
naddeoa: Two slightly separate things going on here.

First thing. This function is actually called all the time when logging data inside of the ProcessRollingLogger. Check `process_log_dicts`. Since the logger handles batches of data, the batches are "reduced" into a single data frame with `reduce_log_requests` literally as the reducer in the built in `reduce()` fn. That's what's going on in `_process_dicts` in the ProcessRollingLogger:

```python
                    giga_message = reduce(reducer, sub_group)  # Called giga because its big
```

Second thing. The disjoint column thing is referring to the following scenario.

```python
logger.log({'a': 1, 'b': 2})
logger.log({'b': 4, 'c': 3})
```

This is something that can happen within a single dataset. Some messages might only contain some columns of the dataset. That's fine normally, and I could probably even pass it to whylogs for logging too, but I can't always assume that I can "reduce" those two log messages down into a single row/data frame. Because of that, when I'm batching these things up, I make sub batches based on the things that I know can be merged into larger data frames before feeding them into whylogs. Those dimensions include time, column names, and dataset id.

All of that stuff is happening in ProcessRollingLogger's `_process_dicts` fn, which is basically where all the batching/logging happens. So, the messages are consumed in bulk from the actor's queue, and then they're batched up into the message type (logging message, status message, etc), and handled in bulk. That means doing data merging in the case of the log messages so minimize the whylogs calls/maximize the data per whylogs call.

jamie256 (dismissed):
LGTM!

jamie256 (approved):

PR #1326, opened 2023-07-31
make HTTP_PROXY env var accessible in WhyLabsWriter

### LINKED ISSUE #1325: Need to support HTTP_PROXY and/or HTTPS_PROXY
https://github.com/whylabs/whylogs/blob/b7d348c00e7e1c966a5a80d83d7c8cec3ae572e2/python/whylogs/api/writer/whylabs.py#L264C20-L264C31

FelipeAdachi:
## Description

This PR adds a ProxyManager for the s3 pool manager if http/https proxy env vars are available, and also forwards proxy URL and authorization headers to whylabs client.

### DIFF
--- a/python/whylogs/api/writer/whylabs.py
+++ b/python/whylogs/api/writer/whylabs.py
@@ -320,11 +320,9 @@ class WhyLabsWriter(Writer):
         cache_key += str(hash(self._key_refresher))
         if self._proxy:
             config.proxy = self._proxy
-            cache_key += str(hash(self._proxy))
             default_header = _get_auth_headers(self._proxy)
             if default_header:
                 config.proxy_headers = default_header
-                cache_key += str(hash(str(default_header)))
         config.discard_unknown_keys = True
         # Disable client side validation and trust the server
         config.client_side_validation = False
--- a/python/whylogs/api/writer/whylabs.py
+++ b/python/whylogs/api/writer/whylabs.py
@@ -118,6 +118,15 @@ def _validate_api_key(api_key: Optional[str]) -> str:
         raise ValueError("Invalid format. Expecting a dot at an index 10")
     return api_key[:10]
 
+def _get_auth_headers(proxy_url: str) -> Dict[str, str]:
+    parsed_url = urlparse(proxy_url)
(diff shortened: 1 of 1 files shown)


andyndang (dismissed):
LGTM

python/whylogs/api/writer/whylabs.py
```diff
@@ -261,12 +263,21 @@ def __init__(
...
+                )
+            else:
+                pool = PoolManager(
+                    num_pools=4,
+                    maxsize=10,
+                    assert_hostname=self._s3_endpoint_subject,
+                    server_hostname=self._s3_endpoint_subject,
+                )
             self._s3_pool = pool
             _UPLOAD_POOLER_CACHE[pooler_cache_key] = pool
```
naddeoa: Do you know how the pooling is supposed to work? It looks like every instance of the whylabs writer has its own pool and there don't seem to be any concurrent requests (that are obvious). Are we actually getting benefit from the pool or are we just treating it like a single connection?

FelipeAdachi: I don't know what was the original motivation to use pooling. The scope of this PR was only to adapt the original code to accept proxy configurations through ProxyManager.

andyndang: We can easily saturate an endpoint by creating too many connections. The TLS handshake kills our upload FYI. This pooling enable us to re-use the TCP connection and don't have to repeatedly handshake with the endpoint

python/whylogs/api/writer/whylabs.py
```diff
@@ -293,7 +318,13 @@ def _refresh_client(self) -> None:
         config.api_key = {"ApiKeyAuth": ""}
         config.refresh_api_key_hook = self._key_refresher
         cache_key += str(hash(self._key_refresher))
-
+        if self._proxy:
+            config.proxy = self._proxy
+            cache_key += str(hash(self._proxy))
```
vsuen: We don't need to update the cache_key for the proxy and proxy_headers fields. The config itself is already included in the cache_key

```
cache_key += str(hash(config))
```

vsuen (approved):

PR #1329, opened 2023-08-02
fix serializing confusion matrix from older whylogs

cswarth:
## Description

Support backwards compatibility.
Decoding a KllFloatsSketch when expecting a KllDoublesSketch can result in a few different types of exceptions.
This change broadens the class of exceptions that can be caught before trying to decode the sketch as KllFloatsSketch. 
Adds unit test to make sure `IndexError` is caught.


jamie256 (dismissed):
Looks good, seems ok to try the fallback on a wider scope.

jamie256 (approved):

PR #1330, opened 2023-08-02
register validator udf

### LINKED ISSUE #1331: schema copy fails when var_opt_sketch is in schema
## Description

In a schema with ConditionValidators, the test `test_udf_schema::test_schema_copy` fails, because now we have a sampler of type `var_opt_sketch` that cannot be pickled.

FelipeAdachi:
## Description

This PR allows for registering condition validators with `register_validator_udf`, inspired by `register_dataset_udf`, etc.

The goal is to simplify the creation and assignment of Condition Validators, so one could do:

```python
def do_something_important(validator_name, condition_name: str, value: Any):
    global action_counter
    print("Validator: {}\n    Condition name {} failed for value {}".format(validator_name, condition_name, value))
    return

@register_validator(["col1"],condition_name="less_than_four",actions=[do_something_important])
def lt_4(x):
    return x < 4

data = pd.DataFrame({"col1": [1,3,7]})
profile = why.log(data, schema=udf_schema()).view()

```

Incidentally, this PR also fixes a bug where a schema with validators with var_opt sketches samplers would error when being copied with `schema.copy()`


python/whylogs/experimental/core/udf_schema.py
```diff
@@ -202,6 +204,30 @@ def _reset_udfs(reset_metric_udfs: bool = True) -> None:
     _resolver_specs = defaultdict(list)
 
 
+def register_validator_udf(
```
richard-rogers: This should probably be in `whylogs/experimental/core/validators/validator.py`, and move to `whylogs/core/validators/validator.py` when it's no longer experimental.

richard-rogers: I'd just make the decorator name `@validator`. In hindsight, I'd like to shorten the UDF decorators too...

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -202,6 +204,30 @@ def _reset_udfs(reset_metric_udfs: bool = True) -> None:
...
+) -> Callable[[Any], Any]:
+    def decorator_register(func):
+        global _validator_udfs
+        name = condition_name or func.__name__
+        name = f"{namespace}.{name}" if namespace else name
+        for col in col_names:
+            validator = ConditionValidator(
+                name=name,
+                conditions={func.__name__: func},
+                actions=actions,
```
richard-rogers: perhaps `copy(actions)`?

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -202,6 +204,30 @@ def _reset_udfs(reset_metric_udfs: bool = True) -> None:
...
+    namespace: Optional[str] = None,
+    schema_name: str = "",
+) -> Callable[[Any], Any]:
+    def decorator_register(func):
+        global _validator_udfs
+        name = condition_name or func.__name__
+        name = f"{namespace}.{name}" if namespace else name
+        for col in col_names:
+            validator = ConditionValidator(
+                name=name,
```
richard-rogers:
```suggestion
                name=col,
```

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -202,6 +204,30 @@ def _reset_udfs(reset_metric_udfs: bool = True) -> None:
...
+                conditions={func.__name__: func},
```
richard-rogers:
```suggestion
                conditions={name: func},
```

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -202,6 +204,30 @@ def _reset_udfs(reset_metric_udfs: bool = True) -> None:
     _resolver_specs = defaultdict(list)
 
 
+def register_validator_udf(
+    col_names: List[str],
+    condition_name: Optional[str] = None,
+    actions: Optional[List[Callable[[str, str, Any, Any], None]]] = None,
```
richard-rogers: It looks like `actions` is not optional in the `ConditionValidator` constructor. That might be a bug in the constructor, but if not, we should `actions = actions or []`, or just make `[]` the default.

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -202,6 +204,30 @@ def _reset_udfs(reset_metric_udfs: bool = True) -> None:
...
+    namespace: Optional[str] = None,
+    schema_name: str = "",
```
richard-rogers: I'd add `enable_sampling: bool = True` here, with a comment below about why it's ignored.

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -202,6 +204,30 @@ def _reset_udfs(reset_metric_udfs: bool = True) -> None:
...
+    def decorator_register(func):
+        global _validator_udfs
+        name = condition_name or func.__name__
+        name = f"{namespace}.{name}" if namespace else name
+        for col in col_names:
+            validator = ConditionValidator(
+                name=name,
+                conditions={func.__name__: func},
+                actions=actions,
+                enable_sampling=False,
```
richard-rogers:
```suggestion
                enable_sampling=False,  # should be = enable_sampling, but bug...
```

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -348,6 +384,8 @@ def udf_schema(
 
     for name in schema_names:
         resolver_specs += _resolver_specs[name]
+        for validators_udf in _validator_udfs[name]:
+            validators = _update_validators(validators, validators_udf)
```
richard-rogers:
```suggestion
```

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -348,6 +384,8 @@ def udf_schema(
 
     for name in schema_names:
         resolver_specs += _resolver_specs[name]
+        for validators_udf in _validator_udfs[name]:
```
richard-rogers:
```suggestion
```

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -324,6 +350,16 @@ def add5(x):
...
+def _update_validators(validators, validators_udf):
+    if not validators:
+        validators = {}
+    for col, validator in validators_udf.items():
+        if col not in validators:
+            validators[col] = []
+        validators[col].extend(validator)
+    return validators
+
+
```
richard-rogers:
```suggestion
def generate_validators(
    initial_validators: Optional[Dict[str, List[Validator]]],
    schema_name: Union[str, List[str]],
    include_defualt_schema: bool = True
) -> Dict[str, List[Validator]]:
    """Merge registered validators for requested schemas"""
    schema_names = schema_name if isinstance(schema_name, list) else [schema_name]
    if include_default_schema and "" not in schema_names:
        schema_names = [""] + schema_names

    result: Dict[str, List[Validator]] = defaultdict(List)
    if initial_validators is not None:
        for column, validators in initial_validators.items():
            result[column] += validators

    for name in schema_names:
        for registration in validator_udfs[name]:
            for column, validators in registration.items():
                result[column] += validators
    return result
```

python/whylogs/experimental/core/udf_schema.py
richard-rogers: make 399 `validators = generate_validators(validators, schema_name, include_default_schema)`

andrewelizondo:
This is a much easier approach from a user perspective and makes a ton of sense. Is the intent here to remove the need for the Predicate class entirely?

andrewelizondo (dismissed):
lgtm 👍

FelipeAdachi:
> This is a much easier approach from a user perspective and makes a ton of sense. Is the intent here to remove the need for the Predicate class entirely?

Not exactly, the predicate is still there to be used when needed - it can be used for condition validators, condition count metrics, or metric constraints. I think the main motivation for predicates is for serialization purposes.

richard-rogers (changes requested):

python/whylogs/experimental/core/udf_schema.py
FelipeAdachi: I think validator would be confusing with the Validator class we already have. Maybe @register_validator ?

python/whylogs/experimental/core/udf_schema.py
FelipeAdachi: The condition validator should not be named after the column: the condition validators are later mapped to columns in a dict.

python/whylogs/experimental/core/udf_schema.py
FelipeAdachi: This is actually very confusing, because we have:
- condition name
- condition validator name

Which are not necessarily the same.  One condition validator can have multiple conditions. By using the register decorator, it happens to be 1 condition to 1 validator, so maybe in this case we could have the same name for both?

All to say that I agree with the proposal.

python/whylogs/experimental/core/udf_schema.py
FelipeAdachi: comment is no longer needed, the bug was solved

python/whylogs/experimental/core/udf_schema.py
richard-rogers: I still vote for just `@validator`. The syntax and context are different enough that I don't think there will be confusion.

python/whylogs/experimental/core/udf_schema.py
richard-rogers: The `name` argument to the `ConditionValidator` constructor appears to refer to the column name rather than the validator name?

python/tests/experimental/core/test_udf_schema.py
```diff
@@ -55,6 +61,42 @@ def square(x: Union[Dict[str, List], pd.DataFrame]) -> Union[List, pd.Series]:
     return x["col1"] * x["col1"] if isinstance(x, (pd.Series, pd.DataFrame)) else [xx * xx for xx in x["col1"]]
 
 
+action_list = []
+
+
+def do_something_important(validator_name, condition_name: str, value: Any, column_id=None):
+    global action_counter
```
richard-rogers: `action_counter` is undefined?

python/tests/experimental/core/test_udf_schema.py
```diff
@@ -55,6 +61,42 @@ def square(x: Union[Dict[str, List], pd.DataFrame]) -> Union[List, pd.Series]:
...
+    print("Validator: {}\n    Condition name {} failed for value {}".format(validator_name, condition_name, value))
+    action_list.append(value)
+    if column_id:
```
richard-rogers, line 71: This confuses me

python/whylogs/experimental/core/validators/condition_validator.py
```diff
@@ -0,0 +1,53 @@
...
+from copy import copy
+from typing import Any, Callable, Dict, List, Optional, Union
+
+from whylogs.core.validators import ConditionValidator
+
+_validator_udfs: Dict[str, List[Dict[str, List[ConditionValidator]]]] = defaultdict(list)
+
+
+def register_validator(
+    col_names: List[str],
```
richard-rogers:
```suggestion
    col_names: Union[str, List[str]],
```

python/whylogs/experimental/core/validators/condition_validator.py
```diff
@@ -0,0 +1,53 @@
...
+    condition_name: Optional[str] = None,
+    actions: List[Callable[[str, str, Any, Any], None]] = [],
+    namespace: Optional[str] = None,
+    schema_name: str = "",
+    enable_sampling: bool = True,
+) -> Callable[[Any], Any]:
```
richard-rogers, line 17:
```suggestion
) -> Callable[[Any], Any]:
    col_names = col_names if isinstance(col_names, list) else [col_names]
```

python/whylogs/experimental/core/validators/condition_validator.py
```diff
@@ -0,0 +1,53 @@
+from collections import defaultdict
+from copy import copy
+from typing import Any, Callable, Dict, List, Optional, Union
+
+from whylogs.core.validators import ConditionValidator
+
+_validator_udfs: Dict[str, List[Dict[str, List[ConditionValidator]]]] = defaultdict(list)
```
richard-rogers: Should we move the registration machinery to `validator.py` so other types of validators can use it too?

python/whylogs/experimental/core/validators/condition_validator.py
```diff
@@ -0,0 +1,53 @@
...
+    schema_name: str = "",
+    enable_sampling: bool = True,
+) -> Callable[[Any], Any]:
+    def decorator_register(func):
+        global _validator_udfs
+        name = condition_name or func.__name__
+        name = f"{namespace}.{name}" if namespace else name
+        for col in col_names:
+            validator = ConditionValidator(
+                name=name,
```
richard-rogers, line 26: I still think this should be the column name?

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -26,6 +26,9 @@
     _reset_metric_udfs,
     generate_udf_resolvers,
 )
+from whylogs.experimental.core.validators.condition_validator import (
+    _generate_validators,
```
richard-rogers: `s/_generate_validators/generate_validators/`

python/whylogs/experimental/core/udf_schema.py
FelipeAdachi: If you're referring to the docstring in the ConditionValidator:
`name (str): The name of the column to validate`

That appears to be a mistake. I will correct it. If you were referring to anything else, please let me know!

richard-rogers (changes requested):

python/whylogs/experimental/core/udf_schema.py
richard-rogers: Trying to track down where in the system it determines the set of validators to apply to a specific column was challenging, but I think you're right... the comment is lying. Let's fix it.

python/whylogs/core/validators/condition_validator.py
```diff
@@ -18,15 +18,15 @@ class ConditionValidator(Validator):
     A validator that checks if a column satisfies a condition.
 
     Args:
-        name (str): The name of the column to validate
+        name (str): The name of the ConditionValidator.
```
richard-rogers, line 21: We should include descriptions of the action arguments here.

python/whylogs/core/validators/condition_validator.py
```diff
@@ -18,15 +18,15 @@ class ConditionValidator(Validator):
...
         conditions (Dict[str, Union[Condition, Callable[[Any], bool]]]): A dictionary of conditions to check.
         actions (List[Callable[[str, str, Any, Any], None]]): A list of actions to take when a condition fails.
         enable_sampling (bool): Whether to enable sampling of failed values. Defaults to True.
 
     """
 
     conditions: Dict[str, Union[Condition, Callable[[Any], bool]]]
-    actions: List[Callable[[str, str, Any, Any], None]]
+    actions: List[Union[Callable[[str, str, Any], None], Callable[[str, str, Any, Optional[Any]], None]]]
```
richard-rogers: I take it the `Union` is for back compatability?

We need to do some cleanup in the `track()` calls in `DatasetProfile`

python/whylogs/core/validators/validator.py
```diff
@@ -1,19 +1,40 @@
...
 
 class Validator(ABC):
     name: str
     conditions: Dict[str, Condition]
-    actions: List[Callable[[str, str, Any], None]]
+    actions: List[Union[Callable[[str, str, Any], None], Callable[[str, str, Any, Optional[Any]], None]]]
 
     @abstractmethod
-    def columnar_validate(self, data: PreprocessedColumn):
+    def columnar_validate(self, data: Any, identity_values: Any = None) -> None:
```
richard-rogers:
```suggestion
    def columnar_validate(self, data: Any, identity_values: Optional[Any] = None) -> None:
```
Why is `data` no longer a `PreprocessedColumn`?

python/whylogs/experimental/core/validators/condition_validator.py
richard-rogers: Nope :)

python/whylogs/experimental/core/validators/condition_validator.py
```diff
@@ -0,0 +1,53 @@
...
+
+from whylogs.core.validators import ConditionValidator
+
+_validator_udfs: Dict[str, List[Dict[str, List[ConditionValidator]]]] = defaultdict(list)
+
+
+def register_validator(
+    col_names: List[str],
+    condition_name: Optional[str] = None,
+    actions: List[Callable[[str, str, Any, Any], None]] = [],
```
richard-rogers:
```suggestion
    actions: List[Callable[[str, str, Any, Optional[Any]], None]] = [],
```

richard-rogers (changes requested):

python/tests/experimental/core/test_udf_schema.py
FelipeAdachi: this list is just to verify that the action was called with the correct column id.

There are more elegant ways to do that, but I was hitting concurrency errors when running all the tests, and that was the first way I found to bypass that.

I can add a comment explaining that or find another approach, if you suggest any

python/whylogs/experimental/core/validators/condition_validator.py
FelipeAdachi: Considering the registration assumes the existence of Conditions, which is specific to Condition Validators, I think this wouldn't be usable to other types of validators. I'm not even sure what other types of validators could be. maybe we should make this more specific, like @condition_validator ?

python/whylogs/core/validators/validator.py
FelipeAdachi: A PreprocessedColumn won't preserve the original order, and that's a problem when we want to correlate the values with the identity PreprocessedColumn, so it needs to be done on the raw data. This is not so much of a problem for the track_datum case, but it is for dataframes.

python/whylogs/core/validators/condition_validator.py
FelipeAdachi: Yes, for back compat - before the identity values addition, action would take only 3 arguments. With identity values, it's now 4.

python/tests/experimental/core/test_udf_schema.py
```diff
@@ -55,6 +61,42 @@ def square(x: Union[Dict[str, List], pd.DataFrame]) -> Union[List, pd.Series]:
...
+
+def test_validator_udf_pandas() -> None:
+    global action_list
+    data = pd.DataFrame({"col1": [1, 3, 7]})
+    schema = udf_schema()
+    why.log(data, schema=schema).view()
+    assert 7 in action_list
+
+
+def test_validator_udf_row_with_id() -> None:
```
richard-rogers, line 90: It would be good to cover all the code paths... row, dataframe, homogeneous dataframe

python/tests/experimental/core/test_udf_schema.py
richard-rogers: I think I just missed the test that was looking for the row id.

python/whylogs/experimental/core/validators/condition_validator.py
```diff
@@ -0,0 +1,55 @@
...
+
+from whylogs.core.validators import ConditionValidator
+
+_validator_udfs: Dict[str, List[Dict[str, List[ConditionValidator]]]] = defaultdict(list)
+
+
+def condition_validator(
+    col_names: Union[str, List[str]],
+    condition_name: Optional[str] = None,
+    actions: List[Callable[[str, str, Any, Optional[Any]], None]] = [],
```
richard-rogers:
```suggestion
    actions: Union[List[Callable[[str, str, Any, Optional[Any]], None]], Callable[[str, str, Any, Optional[Any]], None]]] = [],
```
Along with `actions = actions if isinstance(actions, list) else [actions]`

python/whylogs/experimental/core/validators/condition_validator.py
richard-rogers: I think the `ConditionValidator` class should not repeat the members from the base class?

The `Validator` currently takes `Condition` as conditions. I'd be OK with letting it also take `Callable` so we could register any `Validator` subclass. You'd only need a specialized decorator if you need to fill in fields not present in the base class, and we could use kwargs for that?

python/whylogs/core/validators/validator.py
richard-rogers: I still think `identity_values` needs to be `Optional`

python/whylogs/core/validators/condition_validator.py
```diff
@@ -18,16 +18,15 @@ class ConditionValidator(Validator):
     A validator that checks if a column satisfies a condition.
 
     Args:
-        name (str): The name of the column to validate
+        name (str): The name of the ConditionValidator.
         conditions (Dict[str, Union[Condition, Callable[[Any], bool]]]): A dictionary of conditions to check.
-        actions (List[Callable[[str, str, Any, Any], None]]): A list of actions to take when a condition fails.
+        actions (List[Union[Callable[[str, str, Any], None], Callable[[str, str, Any, Optional[Any]], None]]]):
+            A list of actions to take when a condition fails. The action arguments represent, respectively:
+            condition validator name, condition name, validated value, identity value (if any).
```
richard-rogers:
```suggestion
            condition validator name, condition name, invalid value, row identity value (if any).
```

python/whylogs/core/validators/validator.py
FelipeAdachi: right, missed that

python/whylogs/core/validators/condition_validator.py
```diff
@@ -18,16 +18,15 @@ class ConditionValidator(Validator):
     A validator that checks if a column satisfies a condition.
 
     Args:
-        name (str): The name of the column to validate
+        name (str): The name of the ConditionValidator.
         conditions (Dict[str, Union[Condition, Callable[[Any], bool]]]): A dictionary of conditions to check.
-        actions (List[Callable[[str, str, Any, Any], None]]): A list of actions to take when a condition fails.
+        actions (List[Union[Callable[[str, str, Any], None], Callable[[str, str, Any, Optional[Any]], None]]]):
+            A list of actions to take when a condition fails. The action arguments represent, respectively:
+            condition validator name, condition name, validated value, row identity value (if any).
```
richard-rogers:
```suggestion
            condition validator name, condition name, invalid value, row identity value (if any).
```

python/whylogs/core/validators/validator.py
```diff
@@ -1,19 +1,42 @@
...
 from whylogs.core.metrics.condition_count_metric import Condition
-from whylogs.core.preprocessing import PreprocessedColumn
 
 
+@dataclass
 class Validator(ABC):
     name: str
-    conditions: Dict[str, Condition]
-    actions: List[Callable[[str, str, Any], None]]
+    conditions: Dict[str, Union[Condition, Callable]]
```
richard-rogers:
```suggestion
    conditions: Dict[str, Union[Condition, Callable[[Any], bool]]]
```
I think :)

python/whylogs/core/validators/validator.py
```diff
@@ -1,19 +1,42 @@
...
+    actions: List[Union[Callable[[str, str, Any], None], Callable[[str, str, Any, Optional[Any]], None]]]
```
richard-rogers, line 13: Having the relation and actions both on the `Validator` and `Condition` is redundant, but there's probably nothing we can do about it at this point :(

FelipeAdachi: agreed, unfortunately

richard-rogers (approved):

PR #1334, opened 2023-08-08
Update schema documenation

richard-rogers:
## Description

Covers the new features supported by schemas.

### DIFF
--- a/python/examples/basic/Schema_Configuration.ipynb
+++ b/python/examples/basic/Schema_Configuration.ipynb
@@ -6,7 +6,7 @@
         "id": "_klgn5JO0oqh"
       },
       "source": [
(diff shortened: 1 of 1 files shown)


python/examples/basic/Schema_Configuration.ipynb
jamie256, line 1: This looks like whitespace changes only, can you update this PR?

richard-rogers: It looks like I uploaded an old version of the notebook instead of my updated one. I will look a bit longer for my updated version, but if I don't find it I'll have to start over :(

richard-rogers:
replaced with #1348

PR #1337, opened 2023-08-09
Respect the allow_anonymous flag even if you've previously been anonymous

naddeoa:
Previously, you would do an anonymous upload despite allow_anonymous being False if you already had an anonymous session id stored in your local config file. Now the flag is required in addition.


murilommen (dismissed):

jamie256 (approved):
LGTM!

PR #1339, opened 2023-08-11
Factor out the whylabs client cache so sessions can use it

naddeoa:
The whylabs writer had a ton of complex caching code embedded in it that depended heavily on self mutations and implicit updates. I tried to simplify the cache logic and make everything as explicit as I could so that the client cache would be reused from other places in the application. This moves the http proxy env var logic into the session config as well, where all of the config in whylogs now lives.

There were a lot of circular dependencies and initialization race conditions that made me introduce some lazy initialization. This also makes the new Session concept mandatory for using whylabs, which should have minimal impact because we can call it implicitly for people who don't call it themselves whenever the whylabs writer is used.

This doesn't allow the anonymous sessions to use the private s3 functionality, which seems ok for the moment. The anonymous sessions don't use the whylabs writer under the hood so that didn't come for free.

In general, we should always get the latest session with _default_init() instead of saving a reference to a session because the session can be reinitialized.

### WhyLabs Writer
- Sessions are now required by the whylabs writer, but the whylabs writer implicitly initializes the session if the user didn't. The implicit init doesn't allow for anonymous usage though, just to avoid any surprises.
- If the writer is passed api key/org id/dataset id explicitly then it won't depend on the session.
- If the session is an anonymous session then the writer will throw if its created because it won't have any way of getting the credentials for uploading to WhyLabs. User can still directly supply api key to the writer if they want to use it directly for whatever reason. This behavior is similar to how it was before when the writer couldn't find anything in the env, but it goes through the session now.

### Client Cache
- Cache logic is no longer directly embedded in the whylabs writer as a series of mutations/side effects, it stands alone and anything can get cached instances of clients.
- Cache key is explicitly modeled instead of being a series of string concats with hashes.
- Cache key is immutable (frozen via dataclass)
- The session has to be the one to initialize the client cache, so the session is required now. This is because the session depends on the client cache to do uploads, while the cache depends on the session to get various env variable states. TLDR, either the user calls `why.init()` themselves or we call `_default_init()` as a fallback if we need to make sure the session exists.

### Sessions
- Various parts of the sessions are now lazy. Specifically the GuestSession fetching session ids. It used to happen in the __init__ but because of circular dependencies it now happens whenever its needed (profile upload time). This means that we can't tell the user their session id right after a session is made but it improves the overall experience because it happens on demand whenever the session is "bad" for any reason (like being claimed). Users used to have to reset the notebook kernel to generate a new session id and now it just works automatically (with a warning message when it happens).
- Nothing should ever hang onto the session reference because the user can reinitialize the session at any point. Everything should get a hold of the current session by doing one of these things:
    - `get_current_session()` will give you an optional Session object.
    - `_default_init()` will implicitly initialize a session if the user doesn't, or get the current one. This can be used if a session must exist for whatever reason. The implicitly initialized session is a local session, which means nothing will happen automatically.

### DIFF
--- a/python/examples/integrations/Guest Session.ipynb
+++ b/python/examples/integrations/Guest Session.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -28,6 +28,38 @@
   },
   {
    "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Initializing session with config /home/anthony/.config/whylogs/config.ini\n",
+      "\n",
+      "✅ Using session type: WHYLABS_ANONYMOUS\n",
+      " ⤷ session id: session-6LpLjnAE\n"
+     ]
+    },
+    {
+     "data": {
+      "text/plain": [
+       "<whylogs.api.whylabs.session.session.GuestSession at 0x7fc674559c70>"
+      ]
+     },
+     "execution_count": 2,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "import whylogs as why\n",
+    "\n",
+    "why.init()"
+   ]
+  },
+  {
+   "cell_type": "code",
    "execution_count": 3,
    "metadata": {},
    "outputs": [
@@ -35,16 +67,16 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Initialing session with config /home/anthony/.config/whylogs/config.ini\n",
-      "✅ Using session type: WHYLABS_ANONYMOUS\n",
-      " ⤷ session id: session-XfQUnBhJ\n"
+      "\n",
+      "✅ Aggregated 48842 rows into profile foo\n",
+      "\n",
+      "Visualize and explore this profile with one-click\n",
+      "🔍 https://hub.whylabsapp.com/resources/model-1/profiles?profile=ref-zv5Qm5zwJw0XEzpo&sessionToken=session-6LpLjnAE\n"
      ]
     }
    ],
    "source": [
-    "import whylogs as why\n",
-    "\n",
-    "why.init()"
+    "profile = why.log(df, name=\"foo\")"
    ]
   },
   {
@@ -56,15 +88,17 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "✅ Aggregated 48842 rows into profile foo\n",
+      "\n",
+      "✅ Aggregated 48842 rows into profile \n",
       "\n",
       "Visualize and explore this profile with one-click\n",
-      "🔍 https://hub.whylabsapp.com/resources/model-1/profiles?profile=ref-547v42LMV5rKJi4R&sessionToken=session-XfQUnBhJ\n"
+      "🔍 https://hub.whylabsapp.com/resources/model-1/profiles?profile=1691712000000&sessionToken=session-6LpLjnAE\n"
      ]
     }
    ],
    "source": [
-    "profile = why.log(df, name=\"foo\")"
+    "# Upload the same data as a batch profile by leaving out the name\n",
+    "profile = why.log(df)"
    ]
   },
   {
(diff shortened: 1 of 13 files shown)


FelipeAdachi (changes requested):
Throughout our examples, a frequent suggested pattern is `log()`->`profile.set_dataset_timestamp()`->`write()`

Some examples that would need to be changed: 
- Writing Reference Profiles to WhyLabs
- Writing Classification Performance Metrics to WhyLabs
-  Writing to WhyLabs

This PR changes this behavior in the current state, considering `log()` now uploads directly to whylabs. We can upload the notebooks to reflect this, but most concerning is that users who based their code on our examples will get a breaking change when uploading whylogs (uploading data twice, with possible diferring timestamps)

We need to either change the code to ensure back compatibility or discuss if we want to include this breaking change and bump the minor version/other measures

FelipeAdachi (approved):

python/tests/api/writer/test_whylabs.py
```diff
@@ -1,39 +1,181 @@
...
+        session = init(whylabs_api_key=api_key, default_dataset_id=dataset_id)
+
+        assert session.get_type() == SessionType.WHYLABS
+
+        writer = WhyLabsWriter()
+        assert writer.key_id == key_id
+        assert writer._org_id == org_id
+        assert writer._dataset_id == dataset_id
+
+    def test_writer_uses_session_for_creds_implicitly(self) -> None:
```
FelipeAdachi, line 93: I tested some notebooks without modification, and now they are working as expected. The only issue now are the error messages - if I log 7 days worth of data with the following code:
```python
    profile = why.log(df).profile()
    profile.set_dataset_timestamp(timestamp)
    writer.write(profile.view())
```
I get one initial message:

`✅ Using session type: LOCAL. Profiles won't be uploaded or written anywhere automatically.`

And a series of messages (one for each day):

`Skipping uploading profiles to WhyLabs because this session is configured to be local only.`

I understand that this is related to the `log()` statement, but for the user this gives the impression that no data is being uploaded at all, and that there is something wrong.

Maybe we should rewrite the message so it is clearer that this is related only to the log statement? and maybe reducing the number of messages or changing the type of message?

FelipeAdachi (changes requested):

python/tests/api/writer/test_whylabs.py
naddeoa: I updated the message to this

```
⚠️ Skipping automatic upload because the session type is LOCAL. Uploads have to be done manually.
```

And I made it only happen once as well. Now the output of the classification notebook looks like this

```
⚠️ No session found. Call whylogs.init() to initialize a session and authenticate. See https://docs.whylabs.ai/docs/whylabs-whylogs-init for more information.
writing profiles to whylabs...
⚠️ Initializing default session because no session was found.
Initializing session with config /home/anthony/.config/whylogs/config.ini

✅ Using session type: LOCAL. Profiles won't be uploaded or written anywhere automatically.
logging data for date 2023-08-10 00:00:00+00:00
⚠️ Skipping automatic upload because the session type is LOCAL. Uploads have to be done manually.
writing profiles to whylabs...
logging data for date 2023-08-11 00:00:00+00:00
writing profiles to whylabs...
logging data for date 2023-08-12 00:00:00+00:00
writing profiles to whylabs...
logging data for date 2023-08-13 00:00:00+00:00
writing profiles to whylabs...
logging data for date 2023-08-14 00:00:00+00:00
writing profiles to whylabs...
logging data for date 2023-08-15 00:00:00+00:00
writing profiles to whylabs...
```

It is a little verbose but each one is only printed once and they all happen for different things. One happens when something tries to get a session. One happens when something forces a default init session. And one happens when one of the top level log apis are used in local mode.

python/examples/integrations/Guest Session.ipynb
FelipeAdachi, line 1: there's an api key in this example. let's remove before merging

FelipeAdachi (changes requested):

python/examples/integrations/Guest Session.ipynb
FelipeAdachi: or is this intended?

naddeoa: Yeah I added the api key (and then expired it from my demo org) so it's more realistic. It won't actually work for anything.

FelipeAdachi (approved):

python/examples/integrations/Guest Session.ipynb
jamie256: Wouldn't a more realistic way be to not have the API key in plaintext in the code at all and use the env variable for the API Key? I wouldn't want to encourage users to follow this way of specifying the API key for a production authenticated session.

This part of the example looks like a mistake because I can't as a user run this code, and there are no comments calling out that these are example values you need to change, it will look broken when I run it.

Can we replace the API key and dataset_id values with placeholder strings:
'<your dataset_id here, e.g. model-62>'

The API key parameter to init seems like we don't want to have there at all, but if there is a compelling argument to having it here, I would replace that string with '<replace with an API key you created at https://hub.whylabsapp.com/settings/access-tokens>'

PR #1341, opened 2023-08-15
Allow session_type argument to init() for back-compatability

richard-rogers:
## Description

Allows `session_type` as a deprecated parameter argument to `init()`

### DIFF
--- a/python/whylogs/api/whylabs/session/session_manager.py
+++ b/python/whylogs/api/whylabs/session/session_manager.py
@@ -11,7 +11,7 @@ from whylogs.api.whylabs.session.session import (
     Session,
 )
 from whylogs.api.whylabs.session.session_types import SessionType
-from whylogs.core.utils.utils import deprecated, deprecated_alias
+from whylogs.core.utils.utils import deprecated_alias
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


jamie256 (dismissed):
LGTM!

python/whylogs/api/whylabs/session/session_manager.py
```diff
@@ -11,6 +11,7 @@
     Session,
 )
 from whylogs.api.whylabs.session.session_types import SessionType
+from whylogs.core.utils.utils import deprecated, deprecated_alias
```
richard-rogers:
```suggestion
from whylogs.core.utils.utils import deprecated_alias
```

python/whylogs/api/whylabs/session/session_manager.py
```diff
@@ -61,7 +62,9 @@ def is_active() -> bool:
         return SessionManager.get_instance() is not None
 
 
+@deprecated_alias(session_type="allow_anonymous")
```
richard-rogers, line 65: This isn't perfect as it doesn't check/message about `allow_local`. But allegedly we only care about this case :)

jamie256: Right, I think thats fine.

jamie256 (approved):

PR #1357, opened 2023-09-05
Add trace_id to log method

jamie256:
* add trace_id as optional parameter to log method
* generate a trace_id from uuid4 if not specified
* add metadata property to profiles and result sets
* add trace_id to the metadata
* set trace_id when uploading results
* Add some unit tests to cover new functionality
* Add result sets unit test.

### DIFF
--- a/python/tests/api/logger/test_segments.py
+++ b/python/tests/api/logger/test_segments.py
@@ -94,6 +94,43 @@ def test_single_column_segment() -> None:
     assert cardinality == 1.0
 
 
+def test_single_column_segment_with_trace_id() -> None:
+    input_rows = 100
+    segment_column = "col3"
+    number_of_segments = 5
+    trace_id = "123-456"
+    d = {
+        "col1": [i for i in range(input_rows)],
+        "col2": [i * i * 1.1 for i in range(input_rows)],
+        segment_column: [f"x{str(i%number_of_segments)}" for i in range(input_rows)],
+    }
+
+    df = pd.DataFrame(data=d)
+    test_segments = segment_on_column("col3")
+    results: SegmentedResultSet = why.log(df, schema=DatasetSchema(segments=test_segments), trace_id=trace_id)
+    assert results.count == number_of_segments
+    partitions = results.partitions
+    assert len(partitions) == 1
+    partition = partitions[0]
+    segments = results.segments_in_partition(partition)
+    assert len(segments) == number_of_segments
+
+    first_segment = next(iter(segments))
+    first_segment_profile = results.profile(first_segment)
+    assert first_segment_profile is not None
+    assert first_segment_profile._columns["col1"]._schema.dtype == np.int64
+    assert first_segment_profile._columns["col2"]._schema.dtype == np.float64
+    assert first_segment_profile._columns["col3"]._schema.dtype.name == "object"
+    segment_cardinality: CardinalityMetric = (
+        first_segment_profile.view().get_column(segment_column).get_metric("cardinality")
+    )
+    cardinality = segment_cardinality.estimate
+    assert cardinality is not None
+    assert cardinality == 1.0
+    assert results.metadata is not None
+    assert results.metadata["whylabs.traceId"] == trace_id
+
+
 def test_single_integer_column_segment() -> None:
     input_rows = 100
     segment_column = "col3"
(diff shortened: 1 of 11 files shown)


python/tests/api/logger/test_segments.py
```diff
@@ -94,6 +94,43 @@ def test_single_column_segment() -> None:
...
+    assert first_segment_profile is not None
+    assert first_segment_profile._columns["col1"]._schema.dtype == np.int64
+    assert first_segment_profile._columns["col2"]._schema.dtype == np.float64
+    assert first_segment_profile._columns["col3"]._schema.dtype.name == "object"
+    segment_cardinality: CardinalityMetric = (
+        first_segment_profile.view().get_column(segment_column).get_metric("cardinality")
+    )
+    cardinality = segment_cardinality.estimate
+    assert cardinality is not None
+    assert cardinality == 1.0
```
richard-rogers, line 131: Took me a second to figure out why this was 1.0. Perhaps a comment? Could just be the jet lag :)

python/whylogs/api/logger/logger.py
```diff
@@ -98,13 +102,22 @@ def log(
 
         # If segments are defined use segment_processing to return a SegmentedResultSet
         if active_schema and active_schema.segments:
-            return segment_processing(active_schema, obj, pandas, row, self._segment_cache)
+            segmented_results = segment_processing(active_schema, obj, pandas, row, self._segment_cache)
+            _extract_common_profile_metadata(
```
richard-rogers: I think I'd be happier if the return value was assigned to something.

python/whylogs/api/logger/logger.py
```diff
@@ -98,13 +102,22 @@ def log(
...
+                segmented_results.metadata, trace_id=trace_id, tags=tags, segment_key_values=segment_key_values
+            )
+            return segmented_results
 
         profiles = self._get_matching_profiles(obj, pandas=pandas, row=row, schema=active_schema)
 
         for prof in profiles:
             prof.track(obj, pandas=pandas, row=row, execute_udfs=False)
-            prof._metadata
```
richard-rogers, line 107: Was old line 107 a bug?

python/whylogs/api/logger/segment_processing.py
```diff
@@ -126,6 +126,7 @@ def segment_processing(
     pandas: Optional[pd.DataFrame] = None,
     row: Optional[Dict[str, Any]] = None,
     segment_cache: Optional[SegmentCache] = None,
+    trace_id: Optional[str] = None,
```
richard-rogers: Should we do something with this argument?

python/whylogs/core/metadata.py
```diff
@@ -0,0 +1,58 @@
...
+diagnostic_logger = logging.getLogger(__name__)
+
+WHYLABS_TRACE_ID_KEY = "whylabs.traceId"
+CREATION_TIMESTAMP_KEY = "whylogs.creationTimestamp"
+DATASET_TIMESTAMP_KEY = "whylogs.datasetTimestamp"
+USER_TAGS_KEY = "whylogs.user.tags"
+NAME_KEY = "whylogs.name"
+
+
+def _extract_common_profile_metadata(
```
richard-rogers: `s/extract/populate/`

python/whylogs/api/logger/segment_processing.py
jamie256: Good catch, removing.

python/whylogs/core/metadata.py
jamie256: sure.

python/whylogs/api/logger/logger.py
jamie256: yes! A casualty of splitting previous PRs from larger changes with the _metadata.

python/whylogs/api/logger/logger.py
jamie256: This updates the dictionary. I added a comment in the code.

result_set has a metadata property but not a metadata setter. The idea is that we are aggregating information but not creating new instances of the collection on every modification.

python/tests/api/logger/test_segments.py
jamie256: added a comment.

richard-rogers (approved):

python/whylogs/api/logger/logger.py
richard-rogers: I might still do `_ = _populate_common_profile_metadate(...)  # I like side-effects :)`

PR #1362, opened 2023-09-07
Update init to allow suppression

naddeoa:
The new init system can make for confusing demos for existing users and new use cases. Making this suppression option to allow us to avoid output when we don't need to see it.

This is also the first example that was updated to use init, though its a kind of weird case since we don't actually want to display the output in this context.

### DIFF
--- a/python/test_notebooks/notebook_tests.py
+++ b/python/test_notebooks/notebook_tests.py
@@ -8,6 +8,7 @@ PARENT_DIR = os.path.join(TEST_DIR, os.pardir, os.pardir)
 OUTPUT_NOTEBOOK = "output.ipynb"
 skip_notebooks = [
     "Guest Session.ipynb",
+    "Single_Image_Tracing_Profile_to_WhyLabs.ipynb",
     "Pyspark_Profiling.ipynb",
     "Kafka_Example.ipynb",
     "Writing_to_WhyLabs.ipynb",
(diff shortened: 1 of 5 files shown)


python/examples/integrations/writers/Single_Image_Tracing_Profile_to_WhyLabs.ipynb
FelipeAdachi, line 1: Is feast a dependency for the required extras?
if not, maybe it's from your environment, @naddeoa ?

In any case, maybe we could clear the output from install whylogs

jamie256, line 1: This is useful to demonstrate the changes, but this example was a side branch I don't think we want to check this file in as a main example.

jamie256 (changes requested):
Improvement over our current output, thanks!

Let's remove that test example notebook from the PR, otherwise looks good!

python/examples/integrations/writers/Single_Image_Tracing_Profile_to_WhyLabs.ipynb
naddeoa: This one isn't the big doc update. This pr was for a one off to get Felipe's example working with init. The other PR shows init better: https://github.com/whylabs/whylogs/pull/1365

python/test_notebooks/notebook_tests.py
```diff
@@ -8,6 +8,7 @@
 OUTPUT_NOTEBOOK = "output.ipynb"
 skip_notebooks = [
     "Guest Session.ipynb",
+    "Single_Image_Tracing_Profile_to_WhyLabs.ipynb",
```
jamie256, line 11: don't need this now that the notebook file is removed.

jamie256 (approved):
LGTM!

PR #1365, opened 2023-09-08
Update Getting Started notebook to use init

naddeoa:
Just a small sample of how the docs/ux will look after the update to include init(). We can use this to make tweaks or back compat fixes before switching more things over.

Gaps:
- There isn't an easy way to make segmented classification/regression metrics work because those don't use why.log under the hood, so I'll have to make a custom integration for them at some point. Just left those examples the same (no init).

### DIFF
--- a/python/examples/integrations/writers/Getting_Started_with_WhyLabsV1.ipynb
+++ b/python/examples/integrations/writers/Getting_Started_with_WhyLabsV1.ipynb
@@ -117,7 +117,7 @@
     },
     {
       "cell_type": "code",
-      "execution_count": 7,
+      "execution_count": 2,
       "id": "b78028ea-c7cb-494f-a303-071f1c345dfc",
       "metadata": {
         "colab": {
@@ -154,7 +154,7 @@
     },
     {
       "cell_type": "code",
-      "execution_count": 9,
+      "execution_count": 3,
       "id": "67b81ab4-a456-4d2d-9547-ad0d772e0aaa",
       "metadata": {
         "colab": {
@@ -471,7 +471,7 @@
               "[8 rows x 126 columns]"
             ]
           },
-          "execution_count": 9,
+          "execution_count": 3,
           "metadata": {},
           "output_type": "execute_result"
         }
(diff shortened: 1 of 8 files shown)


python/examples/integrations/writers/Writing_Reference_Profiles_to_WhyLabs.ipynb
```diff
@@ -307,101 +309,30 @@
...
-    "writer = WhyLabsWriter().option(reference_profile_name=\"my_reference_profile\")\n",
-    "writer.write(file=profile.view())"
-   ]
-  },
-  {
-   "attachments": {},
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "> Another way of setting `reference_profile_name` is to define the environment variable `WHYLABS_REFERENCE_PROFILE_NAME`\n",
```
naddeoa, line 363: @jamie256 Is this true?? I can't find any reference to that name/feature.

jamie256: https://github.com/search?q=repo%3Awhylabs%2Fwhylogs+get_whylabs_refernce_profile_name&type=code

The environment variable reference profile name functionality still exists.

murilommen (dismissed):
apart from the pre-commit errors and updating the branch with the latest changes, this looks like what we need for the examples with `why.init`

python/examples/integrations/writers/Writing_Reference_Profiles_to_WhyLabs.ipynb
naddeoa: lol I think I added that for feature parity, but I had no idea if it existed before. It's fine I guess, it's there now.

python/examples/integrations/writers/Getting_Started_with_WhyLabsV1.ipynb
FelipeAdachi, line 1: - Should we add `reinit=True` to init just to make sure? I ask because I'm testing this, and I got old orgs and model ids that are unrelated to the current model I was planning to upload to, and that confused me.

- On a related note, `reinit=True` is not working for me. Even with reinit, it's automatically picking the old org and model ids that I don't want to be picked.

python/whylogs/api/whylabs/session/prompts.py
```diff
@@ -1,5 +1,6 @@
 import sys
 from typing import List, Optional
+import getpass
```
jamie256: isort?

can you run this locally:
`make pre-commit`

python/examples/integrations/writers/Getting_Started_with_WhyLabsV1.ipynb
FelipeAdachi: It looks like reinit doesn't work as I was thinking - I thoush reinit would reprompt me the variables, but instead you need to set the env vars yourself and then call reinit to update the config.

Maybe this is fine, but it's not clear that this is the case. Maybe it's just a matter of clearly explaining how this works? like, on the comments, stating that "If you want to change the variables in your config file, set them with os....etc and then call why.init(reinit=True)

I think that would give enough info to the user that falls in this scenario as I did.

naddeoa: Rather than clutter the profiling code here with init explanations, we could give a very brief description and then link to the more complete doc that explains these nuances?

FelipeAdachi (approved):
LGTM!

PR #1372, opened 2023-09-13
Fix typo in whylabs session config error message

richard-rogers:
s/if/is/

Also corrects `writer::check_interval()` doctoring


FelipeAdachi (approved):
LGTM!

PR #1376, opened 2023-09-19
Add support for numpy images

naddeoa:
This supports both numpy images and regular nested lists that represent images in the image metric by converting that image data into PIL format and then doing exactly what was happening before.

One thing I need feedback on: is there any loss of information/functionality from logging images in this way? I assume PIL might have modeled some image metadata or manually assigned tags and that wouldn't come through on the np/list log variants of course.


python/whylogs/extras/image_metric.py
```diff
@@ -244,6 +249,24 @@ def zero(cls, config: Optional[MetricConfig] = None) -> "ImageMetric":
         )
 
 
+def init_image_schema(column_prefix: str = "image") -> DatasetSchema:
```
richard-rogers, line 252:
```suggestion
def init_image_schema(column_prefix: str = "image") -> DatasetSchema:
  return DeclarativeSchema().add_resolver_spec(column_name=column_prefix, [MetricSpec(ImageMetric)])
```

python/whylogs/extras/image_metric.py
```diff
@@ -211,6 +213,9 @@ def _update_relevant_submetrics(self, name: str, data: PreprocessedColumn) -> No
     def columnar_update(self, view: PreprocessedColumn) -> OperationResult:
         count = 0
         for image in list(chain.from_iterable(view.raw_iterator())):
+            if isinstance(image, np.ndarray):
+                image = Image.fromarray(image, mode="RGB")
```
jamie256: what about other formats with different numbers of channels? We'll need to document the assumptions here and interpretation as RGB.

We also see RGBA, and L in some of our demos for example.

andyndang: `mode` seems to be an important param. Should it be part of the Schema and is overridable?

jamie256: PIL's fromarray will produce an image we can profile without the mode parameter for a variety of shapes and formats, I think we can leave `mode` out, at least initially.

jamie256:
```suggestion
                image = Image.fromarray(image)
```

naddeoa: Just did some manual testing and the image.mode param is correctly set based on the numpy shape for RGB, RGBA, and L, so I think we're covered here.

python/whylogs/extras/image_metric.py
naddeoa: @richard-rogers tried to take your changes but the image metrics were missing in the dataset profile after that, which I was luckily testing for.

python/whylogs/extras/image_metric.py
```diff
@@ -211,6 +213,9 @@ def _update_relevant_submetrics(self, name: str, data: PreprocessedColumn) -> No
     def columnar_update(self, view: PreprocessedColumn) -> OperationResult:
         count = 0
         for image in list(chain.from_iterable(view.raw_iterator())):
+            if isinstance(image, np.ndarray):
+                image = Image.fromarray(image.astype(np.uint8))
```
jamie256, line 217: do you need the cast there? uint8 is common, but there is also 'F' and 'I'* modes, see https://pillow.readthedocs.io/en/latest/handbook/concepts.html#modes. If we restrict the support to uint8 modes that might be ok, but we would need to document that limitation and help warn when people send in other kinds of inputs.

For example you can send pixels as 16 bit integer values in "I;16" in which case the expected range of values could go into the thousands.

naddeoa: It won't work without the cast.

```python
import numpy as np
img = Image.open("/home/anthony/Downloads/volcano.jpg")
img2 = np.array(img).tolist()
Image.fromarray(np.array(img2))
```

That will  throw with `TypeError: Cannot handle this data type: (1, 1, 3), <i8`. That only happens when you're starting from a nested python list. We would have to provide some config in the schema or something to let people do anything else.

jamie256: Got it, lets scope this to uint8 image data initially, then what you have should be reasonable.

jamie256 (approved):
Let's document the kinds of images this supports, but its a nice start!

python/whylogs/extras/image_metric.py
```diff
@@ -29,6 +29,8 @@
 logger = logging.getLogger(__name__)
 
 try:
+    import numpy as np
```
jamie256: We need to add numpy as dependency of the image extra, or switch this to the NumpyStub

jamie256 (changes requested):
With the change to reference numpy in the image metric, we need to add numpy as a dependency of the whylogs image extra in the pyproject.toml file

the changes are otherwise fine.

jamie256 (approved):
LGTM, thanks!

python/whylogs/extras/image_metric.py
```diff
@@ -244,6 +249,24 @@ def zero(cls, config: Optional[MetricConfig] = None) -> "ImageMetric":
...
+        column_prefix (str): The prefix that appears in the dataset profiles along with all of the
+        image features. If the prefix is "image", then you'll log image with why.log({image: image_data}).
+    """
+
+    class ImageResolver(Resolver):
+        def resolve(self, name: str, why_type: DataType, column_schema: ColumnSchema) -> Dict[str, Metric]:
+            return {ImageMetric.get_namespace(): ImageMetric.zero(column_schema.cfg)}
+
+    return DatasetSchema(
+        types={column_prefix: Image.Image}, default_configs=ImageMetricConfig(), resolvers=ImageResolver()
```
jamie256, line 266: `Image.Image` is already imported Image as ImageType near line 34 in this file (line 32 in mainline) can we not use that?

PR #1379, opened 2023-09-21
Lazier whylabs writer client initialization

### LINKED ISSUE #1371: Can't instantiate whylabs writer without environment variables set
## Description

We used to be able to create a whylabs writer and set its API key, org id, dataset id via `.option()` calls. Now it throws a value error unless those are present in environment variables at construction time.

## Suggestions

Make the checking lazy so they're only needed when `write()` is called. It shouldn't care where they're sourced from (env, config file, etc.)

richard-rogers:
## Description

`WhyLabsWriter()` currently fails if the auth environment variables aren't set. But there are other ways to specifying them (lure config file, `.option()` calls, constructor arguments), so the absence of the env variables should not be fatal.

## Changes

- Initialize API client just before it's actually needed to give other mechanisms a change to specify auth
- Adds integration tests for `WhyLabsWriter`
- Minor docstring/comment/log message corrections
- `WhyLabsWriter::write()` supports `ResultSet`

## Related

Relates to whylabs/whylogs#1371

Closes whylabs/whylogs#1371


python/whylogs/api/writer/whylabs.py
```diff
@@ -558,8 +554,13 @@ def write(self, file: Writable, **kwargs: Any) -> Tuple[bool, str]:
             return self.write_feature_weights(file, **kwargs)
         elif isinstance(file, EstimationResult):
             return self.write_estimation_result(file, **kwargs)
-        elif isinstance(file, SegmentedResultSet) and self._reference_profile_name is not None:
-            return self._write_segmented_reference_result_set(file, **kwargs)
+        elif isinstance(file, ResultSet):
+            if isinstance(file, SegmentedResultSet) and self._reference_profile_name is not None:
+                return self._write_segmented_reference_result_set(file, **kwargs)
+                # TODO: why does segmented reference profile need special handling, but
```
jamie256: Its because segmented reference profile has a special case API server side, so that it can treat N files as a single profile (where N is the number of individual segments), and returns back a single reference profile-id for all the segments uploaded in this single call.

Non-reference segmented profiles upload each segment as a separate file with a separate upload by design, there is no notion of completeness and the segments are not referenceable as a group by profile-id, you can add a new segment later, and the combined merge of all segments has no profile-id. These are platform specific considerations for being able to define a segmented reference to use as a baseline for monitoring.

python/tests/api/writer/test_whylabs_integration.py
```diff
@@ -0,0 +1,98 @@
...
+
+import whylogs as why
+from whylogs.api.writer.whylabs import WhyLabsWriter
+from whylogs.core import DatasetProfileView
+from whylogs.core.feature_weights import FeatureWeights
+from whylogs.core.schema import DatasetSchema
+
+# TODO: These won't work well if multiple tests run concurrently
+
+# TODO: Aim this at an org with granular profiles
```
richard-rogers:
```suggestion
# TODO: Aim this at an org with granular profiles
# API key needs to come from the environment
```

python/tests/api/writer/test_whylabs_integration.py
```diff
@@ -0,0 +1,100 @@
...
+MODEL_ID = "model-2250"
+
+os.environ["WHYLOGS_NO_ANALYTICS"] = "True"
+os.environ["WHYLABS_API_ENDPOINT"] = "https://songbird.development.whylabsdev.com"
+os.environ["WHYLABS_DEFAULT_ORG_ID"] = ORG_ID
+os.environ["WHYLABS_DEFAULT_DATASET_ID"] = MODEL_ID
+# WHYLABS_API_KEY needs to come from the environment
+
+
+@pytest.mark.load
```
richard-rogers, line 27: Perhaps `s/load/integration/`?

python/whylogs/api/writer/whylabs.py
```diff
@@ -557,8 +550,13 @@ def write(self, file: Writable, **kwargs: Any) -> Tuple[bool, str]:
...
+                # segmented non-reference profile and unsegmented reference profile just work?
+            file = file.profile()
```
richard-rogers, line 592: This throws on `SegmentedResultSet` :(  Probably need to loop over segments like we do for segmented reference profiles.

python/tests/api/writer/test_whylabs_integration.py
jamie256: We could add a new pytest mark value for "integration". But let's do some of that in a follow up. Better to get the initial test capabilities and fixes in first.

jamie256 (dismissed):
LGTM!

python/whylogs/api/writer/whylabs.py
jamie256: The result sets have a get_writeables method, and this logic exists already in the result set writer.

python/tests/api/writer/test_whylabs.py
```diff
@@ -231,6 +237,7 @@ def test_upload_reference_request(self, results):
             )
             assert response[0] is True
 
+    @pytest.mark.skip("Skip for now. Will need less mocking")
```
richard-rogers, line 251: There are a few of these tests that mock the code under test and thus don't test anything. I've left them in rather than just deleting them on the theory that we might come back later and mock the back end rather than mocking the entire interaction.

python/tests/api/writer/test_whylabs.py
```diff
@@ -269,9 +278,9 @@ def test_put_feature_weight(self):
         result = writer.write(feature_weights)
 
         writer.write.assert_called_with(feature_weights)
-        assert isinstance(feature_weights, FeatureWeights)
```
richard-rogers, line 272: I deleted a couple of these
```
feature_weights = FeatureWeights(...)
...
assert isinstance(feature_weights, FeatureWeights)
```

python/tests/api/writer/test_whylabs.py
```diff
@@ -51,14 +51,20 @@ def teardown_method(self) -> None:
         SessionManager.reset()
         del os.environ[EnvVariableName.WHYLOGS_CONFIG_PATH.value]
 
-    def test_writer_throws_for_anon_sessions(self) -> None:
+    @pytest.fixture
+    def results(self, pandas_dataframe):
+        return why.log(pandas=pandas_dataframe)
+
+    def test_writer_throws_for_anon_sessions(self, results) -> None:
+        # NOTE: Assumes WHYLABS_API_KEY environment variable is not set
```
jamie256: Can you add an assert to check that assumption?

python/whylogs/api/writer/whylabs.py
```diff
@@ -544,6 +537,36 @@ def _write_segmented_reference_result_set(self, file: SegmentedResultSet, **kwar
...
+
+        logger.debug(f"About to write {len(files)} files:")
+        # TODO: special handling of large number of files, handle throttling
+        for view in files:
+            bool_status, status = self.write(file=view, **kwargs)
+            and_status = and_status and bool_status
+            statuses.append(status)
+        logger.debug(f"Completed writing {len(files)} files!")
+
+        return bool_status, ";".join(statuses)
```
jamie256: `bool_status` looks like it reflects only the status of the last write, this is surprising, I think you intended to return `and_status`?

python/whylogs/api/writer/whylabs.py
```diff
@@ -544,6 +537,36 @@ def _write_segmented_reference_result_set(self, file: SegmentedResultSet, **kwar
...
+        statuses: List[str] = list()
+        and_status: bool = True
+        if not files:
+            logger.warning("Attempt to write a result set with no writables, nothing written!")
+            return True, ""
+
+        logger.debug(f"About to write {len(files)} files:")
+        # TODO: special handling of large number of files, handle throttling
+        for view in files:
+            bool_status, status = self.write(file=view, **kwargs)
```
jamie256: the tuple is a status, containing a bool and a message which makes your `status` variable name confusing, suggest renaming to something like `message` if you unpack the tuple like this.

jamie256 (dismissed):
Left some minor suggestions, overall still looks good.

python/tests/api/writer/test_whylabs.py
richard-rogers: I added code to unset it.

python/whylogs/api/whylabs/session/session_manager.py
```diff
@@ -155,6 +155,7 @@ def _default_init() -> Session:
     if manager is None:
         il.warning_once("Initializing default session because no session was found.", logger.warning)
 
+        # TODO: s/or local// ?
         # To be safe, don't allow default session to be anonymous or local if this is happening as a side effect
```
richard-rogers:
```suggestion
        # To be safe, don't allow default session to be anonymous if this is happening as a side effect
```

python/whylogs/api/whylabs/session/session_manager.py
```diff
@@ -155,6 +155,7 @@ def _default_init() -> Session:
     if manager is None:
         il.warning_once("Initializing default session because no session was found.", logger.warning)
 
+        # TODO: s/or local// ?
```
richard-rogers:
```suggestion
```

python/tests/api/writer/test_whylabs_integration.py
```diff
@@ -0,0 +1,163 @@
...
+from whylogs.core import DatasetProfileView
+from whylogs.core.feature_weights import FeatureWeights
+from whylogs.core.schema import DatasetSchema
+from whylogs.core.segmentation_partition import segment_on_column
+
+# TODO: These won't work well if multiple tests run concurrently
+
+
+ORG_ID = "org-0"
+MODEL_ID = "model-2250"
```
jamie256: Can we require these be set in env variables or some other config?

jamie256 (dismissed):

python/tests/api/writer/test_whylabs_integration.py
```diff
@@ -0,0 +1,162 @@
...
+from whylogs.core.segmentation_partition import segment_on_column
+
+# TODO: These won't work well if multiple tests run concurrently
+
+os.environ["WHYLOGS_NO_ANALYTICS"] = "True"
+# WHYLABS_API_ENDPOINT, WHYLABS_API_KEY, WHYLABS_DEFAULT_ORG_ID, and
+# WHYLABS_DEFAULT_DATASET_ID need to come from the environment
+
+ORG_ID = os.environ["WHYLABS_DEFAULT_ORG_ID"]
+MODEL_ID = os.environ["WHYLABS_DEFAULT_DATASET_ID"]
```
jamie256:
```suggestion
ORG_ID = os.environ.get("WHYLABS_DEFAULT_ORG_ID")
MODEL_ID = os.environ.get("WHYLABS_DEFAULT_DATASET_ID")
```

richard-rogers: It's probably easier to diagnose the missing key exception than the collateral damage of having these end up as `None`?

jamie256: The writers have ok exceptions, we just need to avoid reading the values outside of the tests in a way that throws.

jamie256: If you want you can move the read of the org/dataset ids into the test methods?

jamie256 (approved):
LGTM!

PR #1380, opened 2023-09-21
Allow UdfSchema chaining

richard-rogers:
## Description

`UdfSchema` takes an optional `next_schema` argument. After applying a schema's UDFs, it will apply the next schema's UDFs to the current phase's output. `udf_schema()` takes an optional list of schema names and constructs the corresponding chain of `UdfSchema`.


jamie256 (commented):
This is interesting. We might need to be careful of long chains adding surprising latency?

It would help to have a short example notebook to get wider feedback.

python/whylogs/experimental/core/udf_schema.py
```diff
@@ -182,6 +184,9 @@ def _run_udfs(
             self._run_udfs_on_dataframe(pandas, new_df, pandas.keys())
             new_df = pd.concat([pandas, new_df], axis=1)
 
+        if self.next_schema is not None:
+            new_df, new_columns = self.next_schema._run_udfs(new_df, new_columns)
+
```
richard-rogers, line 189: In its current form, this only chains the UDFs themselves--not the metrics, validators, etc. That's probably bad. All the schema bits should be merged, or we should split out a mechanism to just chain the UDFs alone.

richard-rogers:
Don't do this :)

PR #1388, opened 2023-09-29
handle np.nans and Nones on ConfusionMatrix calculation

### LINKED ISSUE #1373: log_classification_metric calculates wrong set of labels when np.nan present in columns
## Description

The following code:

```python
import pandas as pd
import whylogs as why
import numpy as np
df = pd.DataFrame({"label_output":[0.0,0.0,1.0,1.0,np.nan]*10000,"pred_output":[0.0,1.0,1.0,0.0,np.nan]*10000,"pd_output":[0.2,0.3,0.4,0.5,np.nan]*10000})

why.log_classification_metrics(df,
                            target_column="label_output",
                            prediction_column="pred_output",
                            score_column="pd_output",
                            log_full_data=True
                            )

```

Will hit the following error:

> ValueError: The initialized confusion matrix has 20002 labels and the resulting confusion matrix will be larger than is supported by whylogs current representation of the model metric for a confusion matrix of this size, selectively log the most important labels or configure the threshold of {MODEL_METRICS_MAX_LABELS} higher by setting MODEL_METRICS_MAX_LABELS.

This happens because when calculating the labels on `model_performance_metrics.py` with `labels = sorted(list(set(targets + predictions)))`  the NaN values are included multiple times in the resulting list of labels.

FelipeAdachi:
## Description

This PR replaces np.nans and Nones with string "None" in labels, predictions, and targets used in ConfusionMatrix. This is done to prevent the bug reported in https://github.com/whylabs/whylogs/issues/1373 .

When a "None" is present on the labels, it will always be the last element in the list.

With this change, in an example where binary classification results are passed with None or np.nan, the Confusion Matrix will be calculated as a three-class confusion matrix, with "None" being the third class.


python/whylogs/core/model_performance_metrics/confusion_matrix.py
```diff
@@ -60,8 +82,6 @@ def __init__(
                     " selectively log the most important labels or configure the threshold of"
                     " {MODEL_METRICS_MAX_LABELS} higher by setting MODEL_METRICS_MAX_LABELS."
                 )
-
-            self.labels = sorted(labels)
```
jamie256, line 64: why is it safe to remove the sort here?

FelipeAdachi: because the sorted is done above in the if/else (lines 66 and 69)

when None/nans are replaced with string "None", sorted() will fail on numerical labels, so for that special case I'm removing "None", sorting it, and appending "None" back to the end of the list.

jamie256: I see it now, thanks!

jamie256 (approved):
LGTM!

PR #1397, opened 2023-10-04
Add prefer sync env variable and option for WhyLabsWriter

### LINKED ISSUE #1396: writing reference profiles hits errors in an AWS Lambda
## Description

Issue is related to lack of support for async calls which reference profile uploads use.

## Suggestions

Add env variable or option to override and use sync calls.

jamie256:
## Description

Adds env variable and option to the WhyLabsWriter to allow writing reference profiles with a sync call to upload the profile to avoid issues with AWS Lambda environment execution.

## Changes

- Adds a WHYLOGS_PREFER_SYNC_KEY env key
- adds _prefer_sync property to the WhyLabsWriter
- Allows env variable to override default value of prefering async where supported by the server side API.
- Add utility helper method read_bool_env_var to parse environ values into bool

Example new environ vars to make AWS Lambda work:
```python
os.environ["WHYLOGS_PREFER_SYNC"] = "1"
os.environ["WHYLOGS_CONFIG_PATH"] = "/tmp/whylogs_config.ini"
```

OR
```python
import whylogs as why

results = why.log({"a": 1})
status = results.writer("whylabs").option(reference_profile_name="ref-a", prefer_sync=True).write()
print(status)
```

## Related

Fixes #1396 

dev build can be evaluated with this PR here:
```
pip install whylogs==1.3.9.dev0
```

### DIFF
--- a/python/whylogs/api/writer/whylabs.py
+++ b/python/whylogs/api/writer/whylabs.py
@@ -38,6 +38,7 @@ from whylogs.api.whylabs.session.whylabs_client_cache import (
 )
 from whylogs.api.writer import Writer
 from whylogs.api.writer.writer import Writable
+from whylogs.context.environ import read_bool_env_var
 from whylogs.core import DatasetProfileView
 from whylogs.core.dataset_profile import DatasetProfile
 from whylogs.core.errors import BadConfigError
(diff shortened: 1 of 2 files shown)


python/whylogs/api/writer/whylabs.py
```diff
@@ -177,6 +179,7 @@ def __init__(
         self._feature_weights = None
         self._reference_profile_name = config.get_whylabs_refernce_profile_name()
         self._api_config: Optional[Configuration] = None
+        self._prefer_sync = read_bool_env_var(WHYLOGS_PREFER_SYNC_KEY, False)
```
andyndang, line 182: Should this be `WHYLOGS_NO_ASYNC`? `prefer` kinda confuses me since it sounds like we might use sync, but we might use async as well.

andyndang (commented):
LGTM but leaving a comment on the naming.

python/whylogs/api/writer/whylabs.py
jamie256: whylogs won't force a mode not exposed by whylabs-client, so the naming was trying to hint that this only affects those places where there is both an async/sync

jamie256: We could add support for WHYLOGS_NO_ASYNC naming and deprecate the "prefer" phrasing.

I'm inclined to keep the current value as we had some community testing with this naming already on the published dev build.

richard-rogers (approved):

PR #1400, opened 2023-10-11
Handle None/NaN with multi-column segmentation

### LINKED ISSUE #1300: Segmenting on multiple columns raises KeyError when there are NaNs
## Description

I've executed the following code on a much larger table, and one of the fields was exactly as `col_nan`, it was supposed to be a boolean, but contained NaNs as well.

```python
import pandas as pd
import whylogs as why
from whylogs.core.schema import DatasetSchema
from whylogs.core.segmentation_partition import ColumnMapperFunction, SegmentationPartition
from whylogs.core.segment import Segment
from whylogs.core.view.segmented_dataset_profile_view import SegmentedDatasetProfileView

df = pd.DataFrame({"col_1":[1,2,3,4], "col_nan": [True, True, None, None]})

segmentation_partition = SegmentationPartition(
    name="col_1,col_nan", mapper=ColumnMapperFunction(col_names=["col_1", "col_nan"])
)
multi_column_segments = {segmentation_partition.name: segmentation_partition}
schema = DatasetSchema(segments=multi_column_segments)

why.log(df, schema=schema)
```

And that gets me a KeyError:

```
KeyError                                  Traceback (most recent call last)
Cell In[5], line 1
----> 1 why.log(df, schema=schema)

File ~/Documents/repos/testing_weird_whylogs/.venv/lib/python3.8/site-packages/whylogs/api/logger/__init__.py:58, in log(obj, pandas, row, schema, name, multiple)
     55     notebook_session_log_comparison(multiple, result_sets)
     56     return result_set
---> 58 result_set = TransientLogger(schema=schema).log(obj, pandas=pandas, row=row, name=name)
     59 notebook_session_log(result_set, obj, pandas=pandas, row=row, name=name)
     60 return result_set

File ~/Documents/repos/testing_weird_whylogs/.venv/lib/python3.8/site-packages/whylogs/api/logger/logger.py:101, in Logger.log(self, obj, pandas, row, schema, timestamp_ms, name)
     99 # If segments are defined use segment_processing to return a SegmentedResultSet
    100 if active_schema and active_schema.segments:
--> 101     return segment_processing(active_schema, obj, pandas, row, self._segment_cache)
    103 profiles = self._get_matching_profiles(obj, pandas=pandas, row=row, schema=active_schema)
    105 for prof in profiles:

File ~/Documents/repos/testing_weird_whylogs/.venv/lib/python3.8/site-packages/whylogs/api/logger/segment_processing.py:149, in segment_processing(schema, obj, pandas, row, segment_cache)
    145 if segment_partition.mapper:
    146     logger.debug(
    147         f"{partition_name}: defines mapper on colums ({segment_partition.mapper.col_names}) and id ({segment_partition.mapper.id})"
    148     )
--> 149 partition_segments = _log_segment(segment_partition, schema, obj, pandas, row, segment_cache)
    150 segmented_profiles[segment_partition.id] = partition_segments
    151 segment_partitions.append(segment_partition)

File ~/Documents/repos/testing_weird_whylogs/.venv/lib/python3.8/site-packages/whylogs/api/logger/segment_processing.py:117, in _log_segment(partition, schema, obj, pandas, row, segment_cache)
    115     columns = partition.mapper.col_names if partition.mapper else None
    116     if columns:
--> 117         _process_simple_partition(partition.id, schema, segments, columns, pandas, row, segment_cache)
    118 else:
    119     raise NotImplementedError("custom mapped segments not yet implemented")

File ~/Documents/repos/testing_weird_whylogs/.venv/lib/python3.8/site-packages/whylogs/api/logger/segment_processing.py:60, in _process_simple_partition(partition_id, schema, segments, columns, pandas, row, segment_cache)
     58 grouped_data = pandas.groupby(columns)
     59 for group in grouped_data.groups.keys():
---> 60     pandas_segment = grouped_data.get_group(group)
     61     segment_key = _get_segment_from_group_key(group, partition_id)
     62     _process_segment(pandas_segment, segment_key, segments, schema, segment_cache)

File ~/.pyenv/versions/3.8.15/lib/python3.8/site-packages/pandas/core/groupby/groupby.py:811, in BaseGroupBy.get_group(self, name, obj)
    809 inds = self._get_index(name)
    810 if not len(inds):
--> 811     raise KeyError(name)
    813 return obj._take_with_is_copy(inds, axis=self.axis)

KeyError: (3, nan)
```

FelipeAdachi:
## Description

The presence of Nones/NaNs would cause multi-column segmentation to crash.
This PR handles the presence of nans by grouping all classes of nans (None, np.nan, math.nan) into a single segment combination.

closes #1300

## Changes

## Related

### DIFF
--- a/python/tests/api/logger/test_segments.py
+++ b/python/tests/api/logger/test_segments.py
@@ -1,3 +1,4 @@
+import math
 import os
 import pickle
 import tempfile
@@ -426,3 +427,27 @@ def test_segment_merge_different_columns() -> None:
             assert segmented_view._columns["A"] is not None
             assert segmented_view._columns["B"] is not None
         assert segmented_view._columns["A"]._metrics["cardinality"].estimate == pytest.approx(1.0)
+
+
+def test_segment_with_nans() -> None:
+    df = pd.DataFrame({"col_1": [1, 2, 3, 4, 5, 6], "col_nan": [True, True, None, None, np.nan, math.nan]})
+    column_segments = segment_on_column("col_nan")
+    schema = DatasetSchema(segments=column_segments)
+    profile_results = why.log(df, schema=schema)
+    assert profile_results.count == 1  # col_nan = True
+    segment = profile_results.segments()[0]
+    segmented_view = profile_results.profile(segment).view()
+    assert segmented_view.get_column("col_nan").get_metric("counts").to_summary_dict()["n"] == 2
+
+    segmentation_partition = SegmentationPartition(
+        name="col_1,col_nan", mapper=ColumnMapperFunction(col_names=["col_1", "col_nan"])
+    )
+    multi_column_segments = {segmentation_partition.name: segmentation_partition}
+    schema = DatasetSchema(segments=multi_column_segments)
+
+    profile_results = why.log(df, schema=schema)
+    assert profile_results.count == 6  # (1,True), (2,True), (3,nan), (4,nan), (5,nan), (6,nan)
+    for segment in profile_results.segments():
+        segmented_view = profile_results.profile(segment).view()
+        # each segment has n=1
+        assert segmented_view.get_column("col_nan").get_metric("counts").to_summary_dict()["n"] == 1
--- a/python/whylogs/api/logger/segment_processing.py
+++ b/python/whylogs/api/logger/segment_processing.py
@@ -9,6 +9,8 @@ from whylogs.core.input_resolver import _pandas_or_dict
 from whylogs.core.segment import Segment
 from whylogs.core.segmentation_partition import SegmentationPartition, SegmentFilter
 from whylogs.core.stubs import pd
+from functools import reduce
+import math
 
 logger = logging.getLogger(__name__)
 
@@ -57,7 +59,17 @@ def _process_simple_partition(
         # simple means we can segment on column values
         grouped_data = pandas.groupby(columns)
         for group in grouped_data.groups.keys():
-            pandas_segment = grouped_data.get_group(group)
+            if any([math.isnan(x) for x in group]):
+                evaluations = []
+                for val,col in zip(group,columns):
+                    if math.isnan(val):
+                        evaluations.append((pandas[col].isna()))
+                    else:
+                        evaluations.append((pandas[col] == val))
+                mask = reduce(lambda x,y: x&y, evaluations)
+                pandas_segment = pandas[mask]
+            else:
+                pandas_segment = grouped_data.get_group(group)
             segment_key = _get_segment_from_group_key(group, partition_id)
             _process_segment(pandas_segment, segment_key, segments, schema, segment_cache)
     elif row:
--- a/python/whylogs/api/logger/segment_processing.py
+++ b/python/whylogs/api/logger/segment_processing.py
@@ -1,4 +1,6 @@
 import logging
+import math
+from functools import reduce
 from typing import Any, Dict, Iterable, Iterator, List, Mapping, Optional, Tuple
 
 from whylogs.api.logger.result_set import SegmentedResultSet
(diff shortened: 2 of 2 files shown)


python/tests/api/logger/test_segments.py
```diff
@@ -426,3 +427,27 @@ def test_segment_merge_different_columns() -> None:
             assert segmented_view._columns["A"] is not None
             assert segmented_view._columns["B"] is not None
         assert segmented_view._columns["A"]._metrics["cardinality"].estimate == pytest.approx(1.0)
+
+
+def test_segment_with_nans() -> None:
+    df = pd.DataFrame({"col_1": [1, 2, 3, 4, 5, 6], "col_nan": [True, True, None, None, np.nan, math.nan]})
+    column_segments = segment_on_column("col_nan")
```
jamie256, line 434: We also need a test that segments on more than one column to reproduce the multi-column segmentation bug.

python/whylogs/api/logger/segment_processing.py
```diff
@@ -57,7 +66,17 @@ def _process_simple_partition(
         # simple means we can segment on column values
         grouped_data = pandas.groupby(columns)
```
jamie256, line 67: Does this line work in multi-column segmentation where one column has a nan value and the other is non-nan?

python/whylogs/api/logger/segment_processing.py
```diff
@@ -44,6 +46,13 @@ def _get_segment_from_group_key(group_key, partition_id) -> Tuple[str, ...]:
     return Segment(segment_tuple_key, partition_id)
 
 
+def _is_nan(x):
```
jamie256, line 49: if the data being processed is a pandas series, there is a hasnans property that we can use. The isnan check is surprisingly slow when we tested this in the past, so if there is a way to avoid checking each item it will save a lot on compute costs of profiling large inputs.

jamie256 (changes requested):
Left a few comments. I think you are covering the single column scenario in the tests, but we need a multi-column segmentation with mixed nan/not-nan values.

python/tests/api/logger/test_segments.py
FelipeAdachi: this is done  later in this same test, starting in line 442

python/whylogs/api/logger/segment_processing.py
FelipeAdachi: this is not being done on a pandas series. It's only on all the possible combinations of segments, not on the raw data. We can try to look for more efficient ways, but I think it's not such a bottleneck, considering the number of checks is limited by the number of segments? or rather, number of groups times the number of elements in a group.

python/whylogs/api/logger/segment_processing.py
FelipeAdachi: yes, but, for example a value (col1,col2) = (42, None) will get mapped to a group (42, float.nan), so this breaks when we try to do `get_group`, because float.nan is not equal to None. So the idea here is to avoid using get_group by checking if the value in a group is a float.nan and then applying the mask manually with `pandas[col].isna()`

python/tests/api/logger/test_segments.py
jamie256: you are right! sorry missed that.

python/whylogs/api/logger/segment_processing.py
jamie256: Thank you for that clarification, makes a lot more sense now that I understand this is only used on the smaller set of segments.

python/whylogs/api/logger/segment_processing.py
jamie256: Thanks for the clarification!

jamie256 (approved):
I missed the tests already cover the multi-column use case, thanks for the clarifications and correcting my understanding there. Changes look good!

PR #1404, opened 2023-10-12
Add sagemaker example notebook

naddeoa:
This just adds a new example notebook for using Sagemaker with the process based rolling logger.


python/examples/integrations/sagemaker_whylabs_example/WhyLabs_Sagemaker-PyTorch.ipynb
jamie256, line 1: the `row_name` name is a little confusing beause we also have a row parameter in why.log, suggest something like `image_column` or `image_column_name` or image_*?

jamie256 (approved):
Overall looks good, sweet example!

PR #1406, opened 2023-10-16
Update make install-poetry to correct version

richard-rogers:
## Description

`make install-poetry` should install 1.4.1, not 1.2.1


python/Makefile
```diff
@@ -34,7 +34,7 @@ endif
 
 install-poetry:
 	@$(call i, Installing Poetry)
-	curl -sSL https://install.python-poetry.org | python3 - --version 1.2.1
+	curl -sSL https://install.python-poetry.org | python3 - --version 1.4.1
```
richard-rogers:
```suggestion
	curl -sSL https://install.python-poetry.org | python3 - --version 1.6.1
```
It looks like LangKit is using 1.6. They should probably match...

jamie256 (dismissed):
LGTM!

jamie256 (approved):

PR #1416, opened 2023-10-31
Add graceful behavior when processes are killed

naddeoa:
When working with gunicorn I discovered that you'll end up shooting yourself in the foot if you start the process logger at the wrong time because gunicorn will end up killing the logger, which results in apparently nothing happening after you make log calls. This will update the logger to be explicit about how dead it is and why, throwing on log() and close() calls.

### DIFF
--- a/python/tests/api/logger/experimental/logger/actor/test_actor_loggers.py
+++ b/python/tests/api/logger/experimental/logger/actor/test_actor_loggers.py
@@ -2,6 +2,7 @@ import datetime
 import logging
 import multiprocessing as mp
 import os
+import time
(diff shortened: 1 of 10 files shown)


jamie256 (commented):
The tests seem to be taking a long time and not progressing, can you look into the CI issues?

https://github.com/whylabs/whylogs/actions/runs/6713659241/job/18267898666

naddeoa:
@jamie256 Finally got this working again with a few additional process logger changes needed for the container.

jamie256 (approved):
LGTM!

PR #1428, opened 2023-11-22
Tons of bug fixes and updates for pyright

naddeoa:
Started using pyright and it immediately exposed tons of issues that mypy wasn't catching. I updated the actor stuff to reflect it since there's still hope that I can fix all of the errors in the container. There are a bunch of actual errors though, not just type issues.


python/whylogs/api/logger/experimental/logger/actor/data_logger.py
```diff
@@ -1,8 +1,17 @@
 from abc import abstractmethod
 from typing import Any, Dict, Generic, List, Optional, TypeVar, Union
 
-from whylogs.core.stubs import pd
+try:
```
richard-rogers, line 4: Why do we not like stubs?

python/whylogs/api/logger/experimental/logger/actor/list_util.py
```diff
@@ -1,17 +1,27 @@
 from itertools import takewhile
-from typing import Any, Generator, List, Tuple, Type
+from typing import Generator, List, Optional, Tuple, Type, TypeVar
 
+K = TypeVar("K")
 
-def get_like_items(items: List[Any]) -> Tuple[List[Any], Type, List[Any]]:
+
+def get_like_items(items: List[K]) -> Tuple[Optional[List[K]], Optional[Type[K]], List[K]]:
```
richard-rogers, line 7: `List[Any]` seems accurate?

python/whylogs/api/logger/experimental/logger/actor/thread_actor.py
```diff
@@ -57,14 +57,14 @@ def size(self) -> int:
...
-class ThreadActor(Actor, th.Thread, Generic[ThreadMessageType]):
+class ThreadActor(Actor[ThreadMessageType], th.Thread, Generic[ThreadMessageType]):
     """
     Subclass of Actor that uses a thread to process messages.
     """
 
     def __init__(self, queue_config: QueueConfig = QueueConfig()) -> None:
-        # our mypy version has a false positive on this super call
-        super().__init__(ThreadQueueWrapper[ThreadMessageType](), queue_config)  # type: ignore
+        # mypy can't infer but pyright can
```
richard-rogers, line 69: That suggests we don't need the `ignore`?

python/whylogs/api/logger/experimental/logger/actor/thread_rolling_logger.py
```diff
@@ -111,11 +116,9 @@ def to_views(self) -> List[DatasetProfileView]:
             result_set = self._target.get_result_set(dataset_timestamp=self._dataset_timestamp)
             segments = result_set.segments() or []
             return [it for it in [result_set.view(segment) for segment in segments] if it is not None]
-        elif isinstance(self._target, DatasetProfile):
+        else:
             return [self._target.view()]
 
-        raise Exception("Unknown profile type")
```
richard-rogers, line 117: Raising a plain `Exception` is probably  not optimal, but do we not want to throw  here?

python/whylogs/api/writer/whylabs.py
```diff
@@ -115,8 +115,8 @@ class WhyLabsWriter(Writer):
     A WhyLogs writer to upload DatasetProfileView's onto the WhyLabs platform.
 
     >**IMPORTANT**: In order to correctly send your profiles over, make sure you have
-    the following environment variables set: `[WHYLABS_ORG_ID, WHYLABS_API_KEY, WHYLABS_DEFAULT_DATASET_ID]`. You
-    can also follow the authentication instructions for the why.init() method at {_INIT_DOCS}.
+    the following environment variables set: `[WHYLABS_ORG_ID, WHYLABS_API_KEY, WHYLABSdefault_DATASET_ID]`. You
```
richard-rogers:
```suggestion
    the following environment variables set: `[WHYLABS_ORG_ID, WHYLABS_API_KEY, WHYLABS_DEFAULT_DATASET_ID]`. You
```

python/whylogs/api/logger/experimental/logger/actor/data_logger.py
naddeoa: Pyright (apparently rightfully) calls them out as invalid python because they aren't types, they're values, so they can't be used in all of the places that types are used. I had to either silence the pyright errors or ditch the stubs. Something is definitely fishy with the stubs though, I'm not sure this is the best way of getting around a hard dependency on pandas.

python/whylogs/api/logger/experimental/logger/actor/list_util.py
naddeoa: Not exactly. it implies that you could supply a `List[str]` and get back a `List[int]`, which you can't. The generic isn't perfect but its more right than Any would be for the first return value, and actually right for the third return value.

python/whylogs/api/logger/experimental/logger/actor/thread_actor.py
naddeoa: The comment was misleading, I removed it. The real type_ignore happens because mypy doesn't like that the Thread init is invoked with these arguments through multiple inheritance, but the Thread init was designed to only look at certain kwargs and doesn't mind this.

python/whylogs/api/logger/experimental/logger/actor/thread_rolling_logger.py
naddeoa: The exception wasn't needed, the type check is exhaustive, even for mypy.

python/whylogs/api/writer/whylabs.py
naddeoa: Woops. Can you tell I refactored something by doing a search/replace for `_default`...

python/whylogs/api/logger/experimental/logger/actor/list_util.py
richard-rogers: ...verbal discussion...

I don't think it's possible to accurately annotate for both the input and output sides, so annotate how you like and add a comment explaining what's happening.

naddeoa: The comments were actually already in the function above. I noted in this function that it just wraps the other one and to see those docs, with a new example in those docs.

richard-rogers (approved):

PR #1432, opened 2023-11-30
Add profiles to the rolling logger status calls

naddeoa:
This change includes the content of the profiles themselves in the
logger status() calls. This isn't super useful as a primary use case but
its very handy for tests. The use case that triggered this change was
performing integration tests on the whylogs container.


murilommen (approved):

jamie256 (approved):
LGTM!

PR #1434, opened 2023-12-05
Initial work on support for zipped segmented reference profiles

richard-rogers:
Plumbs a `zip` keyword argument through `WhyLabsWriter::write()` to upload zipped segmented reference profiles.  Still needs client support.

### DIFF
--- a/python/whylogs/api/writer/whylabs.py
+++ b/python/whylogs/api/writer/whylabs.py
@@ -4,6 +4,7 @@ import os
 import tempfile
 from typing import IO, Any, Dict, List, Optional, Tuple, Union
 from urllib.parse import urlparse
+from zipfile import ZipFile
 
 import requests  # type: ignore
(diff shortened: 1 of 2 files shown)


python/whylogs/api/writer/whylabs.py
```diff
@@ -911,6 +935,15 @@ def _get_upload_urls_segmented_reference(self, whylabs_tags, dataset_timestamp:
         res = self._post_log_segmented_reference(request=request, dataset_timestamp=dataset_timestamp)
         return res["id"], res["upload_urls"]
 
+    # TODO: add this to client API
```
richard-rogers, line 938: This is a fake implementation. We need a client entry point that returns a URL for a zipped profile. The integration test for this case will fail until this is implemented, but the unzipped cases are working.

jamie256 (approved):
I think this will let us test the integration, lets try it out.

PR #1435, opened 2023-12-05
Fix type hints on pyspark timestamps

### LINKED ISSUE #1381: incorrect type hint on collect_dataset_profile_view's dataset_timestamp parameter, should be Optional[datetime]
## Description

the experimental pyspark profiler has a bad type hint on the signature which is confusing.

need example showing some of these other optional parameters.

jamie256:
## Description

Small change to fix #1381

## Changes

There are incorrect type hints in the experimental pyspark api for optional datetimes for timestamps, switch them to correct type hints.

These parameters don't work with int specified so this is updating the hints to reflect behavior but does not change the run time behavior.


FelipeAdachi (approved):
LGTM!

PR #1437, opened 2023-12-12
viz extra: raise ImportError on pybars instead of warning

### LINKED ISSUE #1427: Better error message for missing dependency - pybars3 for viz
## Description

When whylogs is not installed with [viz], using the viz features leads to a confusing error message. we should handle it better so a clearer message to install the correct extra is displayed.

- System info
  - Operating system and version: 
  - Browser and version: 
- Steps to reproduce:
  - Step
  - Next step

## Related

FelipeAdachi:
## Description

When trying to use the viz extra without having pybars installed, a warning was raised, which allows the code to run and hit an error further down. This leads to an error message that looks unrelated to the root cause (missing pybars).

This change raises an error instead of a warning when pybars cannot be installed.

Closes https://github.com/whylabs/whylogs/issues/1427


python/whylogs/viz/utils/html_template_utils.py
```diff
@@ -18,9 +18,10 @@ def _get_compiled_template(template_name: str) -> "Callable":
     try:
         from pybars import Compiler  # type: ignore
     except ImportError as e:
+        msg = "Unable to load pybars; install pybars3 to load profile directly from the current session "
         logger.debug(e, exc_info=True)
-        logger.warning("Unable to load pybars; install pybars3 to load profile directly from the current session ")
```
jamie256, line 22: can you include the hint:
...or `pip install whylogs[viz]` to install dependencies for this module

jamie256 (approved):
LGTM!

PR #1442, opened 2023-12-20
Add regions support for Whylabs profile upload

FelipeAdachi:
## Description

With this PR, an environment variable `WHYLABS_UPLOAD_REGION` is supported to select S3 region when uploading profiles to whylabs (batch, reference and segmented reference profiles)

Requires whylabs-client 0.5.10


murilommen (approved):
looks ok. only question is: should we prevent users from mistyping a region somehow? when it breaks due to a malformed string, is it - the error message - clear enough?

jamie256 (approved):
LGTM!

Filed https://github.com/whylabs/whylogs/issues/1444 for better error message.

PR #1443, opened 2023-12-21
Refac: Add generic process status check

murilommen:
## Description

This repository needs to make the `status` method that is declared on ProcessRollingLogger to be placed 1 structure above, so that any ProcessActor might benefit from it.


python/whylogs/api/logger/experimental/logger/actor/pipe_signaler.py
```diff
@@ -0,0 +1,92 @@
...
+        )
+        self.futures: Dict[str, "Future[Any]"] = {}
+        self._end_polling = th.Event()
+        self._done = th.Event()
+
+    def signal(self, result: Tuple[str, Optional[Exception], Optional[T]]) -> None:
+        """
+        ... [Your existing docstring] ...
+        """
+        self.queue.send(result)
```
naddeoa: these doc strings were removed by chat gpt. Can you sub them back in from the old code.

python/whylogs/api/logger/experimental/logger/actor/process_actor.py
```diff
@@ -45,6 +57,13 @@ def __init__(
         else:
             raise ValueError(f"Unknown queue type: {queue_type}")
 
+        self._sync_enabled = sync_enabled
+
+        # TODO is Any the right type here? It's the type of the PipeSignaler's message
+        # but BaseProcessRollingLogger appears to need some different things to work
+        # properly
+        self._pipe_signaler: Optional[PipeSignaler[Any]] = PipeSignaler() if self._sync_enabled is True else None
+
```
naddeoa: The problem with the pipe signaler in the context of the base logger is that we use the same signaler to signal multiple types of things IIRC. One for making log synchronous, and another for the status check. The most correct type to use here is probably a union of w/e it is we signal for those two methods.

python/whylogs/api/logger/experimental/logger/actor/process_actor.py
```diff
@@ -87,6 +106,32 @@ def close(self) -> None:
...
 
+    def _signal(self, messages: Sequence[SyncMessage] = [], error: Optional[Exception] = None) -> None:
+        if self._pipe_signaler is None:
+            return
+
+        for message in messages:
+            if message.sync:
+                self._pipe_signaler.signal((message.id, error, None))
+
+    # TODO
```
naddeoa: What's the TODO for? Maybe implementing the timeout? We can probably just remove the timeout for now and kick this can down the road.

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger.py
```diff
@@ -344,28 +323,21 @@ def process_log_messages(self, messages: List[LogMessage]) -> None:
...
+            self._signal(messages, e)
 
     def process_raw_log_dicts(self, messages: List[RawLogMessage]) -> None:
         try:
             self._logger.info("Processing raw log request message")
             log_dicts = [msg for msg in [m.to_log_request_dict() for m in messages] if msg is not None]
             self.process_log_dicts(log_dicts)
-            for message in messages:
-                self._signal(message.id, None)
+            # for message in messages:
```
naddeoa: can delete

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
```diff
@@ -180,7 +194,9 @@ class PubSubEmbeddingDict(TypedDict):
     log_embedding_request: LogEmbeddingRequestDict
 
 
-def _decode_pubsub_data(data: Union[PubSubEmbeddingDict, PubSubDict]) -> Optional[bytes]:
+def _decode_pubsub_data(
+    data: Union[PubSubEmbeddingDict, PubSubDict],
+) -> Optional[bytes]:
```
naddeoa: This didn't actually change right? Looks like it shouldn't have been reformatted either based on the length.

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
```diff
@@ -243,7 +259,9 @@ def log_dict_to_data_frame(request: LogRequestDict) -> Tuple[pd.DataFrame, int]:
     return df, len(df)
 
 
-def log_dict_to_embedding_matrix(request: LogEmbeddingRequestDict) -> Tuple[Dict[str, "np.ndarray[Any, Any]"], int]:
+def log_dict_to_embedding_matrix(
+    request: LogEmbeddingRequestDict,
+) -> Tuple[Dict[str, "np.ndarray[Any, Any]"], int]:
```
naddeoa: Same here

naddeoa (commented):
I'll check things out and inspect them locally soon too to make sure everything looks right

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
murilommen: I think this might be something with whylogs' auto format tools? let me run them again

python/whylogs/api/logger/experimental/logger/actor/process_actor.py
murilommen: I've tried with an union of both StatusType + MessageType, but it leads to errors on BaseLogger since it also has a case for a Future[None] type (L#433 on base_process_rolling_logger.py). then maybe `Optional[Union[ProcessStatus, ProcessMessageType]]` could work? idk, I saw you were using `[Any]` so I've decided to maintain

murilommen: I've tested and even with the Optional type, pyright complains about the Future[None] on BaseProcessingLogger. thoughts?

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger_messages.py
murilommen: PEP 8 suggests lines should be limited to 79 characters by default, and this line had 90 without the indents proposed by `flake8` here. however, our pre-commit hooks suggests it can go up to 160, so I'm not sure what might have triggered it. I can revert.

python/whylogs/api/logger/experimental/logger/actor/process_actor.py
naddeoa: Updated this branch with the fix

naddeoa (approved):

Issue #1449, opened 2024-01-03
Don't throw on errors getting observatory profile link after uploading profiles


jamie256:
## Description

upload_batch_profile() in whylogs/api/whylabs/session/session.py near line 324 can throw if there are server side errors, but this is misleading when the profiles are batch profiles that successfully uploaded.

## Suggestions

Don't throw, but catch these exceptions and log as warning.

- [x] I have reviewed the [Guidelines for Contributing](https://github.com/whylabs/whylogs/blob/mainline/.github/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/whylabs/whylogs/blob/mainline/.github/CODE_OF_CONDUCT.md).

PR #1450, opened 2024-01-03
Don't throw on profile url link generation errors post upload

jamie256:
## Description

wrap the call to get the profile url in try/except and log as warning so we're not raising in the case where profile upload is successful.

## Changes

try/except arount post upload profile url call
default strings for results

## Related

Address #1449


andyndang (approved):
Lgtm

PR #1454, opened 2024-01-10
action must exist on mainline before we can experiment with it on a branch

richard-rogers:
Creates the workflow file so GitHub can run it


.github/workflows/docker-test.yml
jamie256, line 1: Did you intend for this to be empty?

richard-rogers: Yes, the file just needs to exist. It has contents on another PR branch, but I can't test it there until GitHub sees the file on mainline.

jamie256 (approved):
LGTM

PR #1456, opened 2024-01-15
make to_pandas result same between profiles pre/post uncompounding

### LINKED ISSUE #1401: uncompounded profiles break viz
## Description

we uncompound metrics so that they map to feature charts in WhyLabs, but if you download that profile again, its hard to compare these profiles to local profiles in the case of images.

Also if you view the to_pandas() of an uncompounded profile, its very confusing because the table would then look sparse.

## Suggestions

* Be able to detect uncompounded state and do that as needed in notebook profile viz
* make to_pandas() show something reasonable on uncompounded profiles

FelipeAdachi:
## Description

Calling `to_pandas()` on an uncompounded profiles will produce a sparse dataframe. This PR adds a verification step in to_pandas to check whether the column is the result of uncompounding. If it is, it won't be included in the resulting DF, producing the same result as the pre-uncompound profile.

Plus: incidental np.nan bug fix for drift calculation

## Related

closes #1401 


jamie256 (approved):
LGTM!

PR #1464, opened 2024-02-02
Transaction updates

richard-rogers:
* More transaction integration tests
* support for distributed transactions
* Transaction context manager

### DIFF
--- a/python/whylogs/api/writer/whylabs.py
+++ b/python/whylogs/api/writer/whylabs.py
@@ -575,8 +575,6 @@ class WhyLabsWriter(Writer):
             return False, "Failed to upload all segments"
 
     def _flatten_tags(self, tags: Union[List, Dict]) -> List[SegmentTag]:
-        # TODO: this isn't right; just to test datastructure
-        # tags [[{'key': 'Gender', 'value': 'F'}], [{'key': 'Gender', 'value': 'M'}]]
         if type(tags[0]) == list:
             result: List[SegmentTag] = []
             for t in tags:
--- a/python/whylogs/api/writer/whylabs.py
+++ b/python/whylogs/api/writer/whylabs.py
@@ -110,8 +110,9 @@ def _check_whylabs_condition_count_uncompound() -> bool:
             return False
         else:
             logger.info(f"Got response code {response.status_code} but expected 200, so running uncompound")
-    except Exception:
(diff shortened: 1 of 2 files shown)


jamie256 (commented):
Is this ready? I think we're waiting on pending changes.

python/whylogs/api/writer/whylabs.py
```diff
@@ -573,6 +574,67 @@ def _write_segmented_reference_result_set(self, file: SegmentedResultSet, **kwar
         else:
             return False, "Failed to upload all segments"
 
+    def _flatten_tags(self, tags: Union[List, Dict]) -> List[SegmentTag]:
+        # TODO: this isn't right; just to test datastructure
+        # tags [[{'key': 'Gender', 'value': 'F'}], [{'key': 'Gender', 'value': 'M'}]]
```
richard-rogers:
```suggestion
```

python/whylogs/api/writer/whylabs.py
```diff
@@ -111,8 +111,9 @@ def _check_whylabs_condition_count_uncompound() -> bool:
             return False
         else:
             logger.info(f"Got response code {response.status_code} but expected 200, so running uncompound")
-    except Exception:
-        logger.warning("Error trying to read whylabs config, falling back to defaults for uncompounding")
+    finally:
+        pass
```
jamie256: Instead of this, can you change the verbosity of this log statement from warning to info/debug?

You need to add back the:
```python
except Exception: 
```

A `finally` block will attempt to run after an exception but your updated code no longer handles the exception. We don't want this operation to block writing profiles if they can't read this value, we can fallback to the default.

jamie256 (changes requested):
Need to fix the exception handling in the log statement first, otherwise looks ok.

python/whylogs/api/writer/whylabs.py
```diff
@@ -111,8 +111,9 @@ def _check_whylabs_condition_count_uncompound() -> bool:
             return False
         else:
             logger.info(f"Got response code {response.status_code} but expected 200, so running uncompound")
-    except Exception:
-        logger.warning("Error trying to read whylabs config, falling back to defaults for uncompounding")
+    finally:
```
richard-rogers:
```suggestion
    except Exception:
```

jamie256 (approved):
LGTM!

jamie256 (approved):

PR #1465, opened 2024-02-08
Fix bug that prevents prefixes from being omitted in udfs

naddeoa:
The types for a udf schema allow the prefix to be None, but our code doesn't work in that scenario. In addition, the annotation interface for udfs don't even let you have a None prefix. This change fixes both of those issues.


jamie256 (approved):
Looks good, thanks for addressing the issue and adding better tests @naddeoa!

jamie256 (approved):

PR #1469, opened 2024-02-16
Making Segmented Ref Profiles upload as ZIP

murilommen:
- This PR intends to use songbird's latest Header to generate the URL on logAsync, which will be used to write potentially all kinds of profiles in the future
- It also extracts some methods out of WhyLabsWriter.write to make it easier to understand

### DIFF
--- a/python/tests/api/writer/test_whylabs.py
+++ b/python/tests/api/writer/test_whylabs.py
@@ -4,7 +4,7 @@ import os
 import random
 import string
 import tempfile
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 import pytest
 import responses
@@ -395,3 +395,35 @@ class TestWhylabsWriter(object):
             writer.key_id
 
         assert len(WhylabsClientCache.instance()._api_client_cache) == 0
+
+    @patch(
+        "whylogs.api.writer.whylabs.WhyLabsWriter._get_upload_url_segmented_reference",
+        return_value=("profile1", "http://upload.url"),
+    )
+    @patch("whylogs.api.writer.whylabs.WhyLabsWriter._do_upload", return_value=(True, "Success"))
+    @patch("whylogs.api.writer.whylabs.WhyLabsWriter._get_dataset_timestamp", return_value=1234567890)
+    @patch("whylogs.api.writer.whylabs.WhyLabsWriter._upload_zipped_files", return_value=(True, "Success"))
+    def test_write_segmented_reference_result_set(
+        self, mock_do_upload, mock_get_url, mock_get_dataset_timestamp, mock_upload_zipped_files
+    ):
+        mock_file = MagicMock()
+        # Mock view.write() method
+        writable_mock = MagicMock()
+        mock_file.get_writables.return_value = [writable_mock]
+
+        writer = WhyLabsWriter()
+        result = writer._write_segmented_reference_result_set(file=mock_file, zip_file=True)
+        assert result == (True, "Success")
+
+        writer._get_dataset_timestamp.assert_called_once()
+        writer._get_upload_url_segmented_reference.assert_called_once_with([], 1234567890, zip_file=True)
+        writer._upload_zipped_files.assert_called_once_with(
+            files=[writable_mock],
+            dataset_timestamp_epoch=1234567890,
+            upload_url="http://upload.url",
+            profile_id="profile1",
+        )
+        writer._do_upload.assert_not_called()
+
+        result_not_zipped = WhyLabsWriter()._write_segmented_reference_result_set(file=mock_file, zip_file=False)
+        assert result_not_zipped == (True, "Success")
--- a/python/tests/api/writer/test_whylabs.py
+++ b/python/tests/api/writer/test_whylabs.py
@@ -10,6 +10,9 @@ import pytest
 import responses
 from platformdirs import PlatformDirs
 from responses import PUT
+from whylabs_client.model.create_reference_profile_response import (
+    CreateReferenceProfileResponse,
+)
 
 import whylogs as why
 from whylogs.api.logger.result_set import SegmentedResultSet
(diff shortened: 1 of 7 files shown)


python/whylogs/api/writer/whylabs.py
```diff
@@ -469,6 +469,9 @@ def write_feature_weights(self, file: FeatureWeights, **kwargs: Any) -> Tuple[bo
             Tuple with a boolean (1-success, 0-fail) and string with the request's status code.
         """
         self._feature_weights = file.to_dict()
+        # TODO if we want to use a WhyLabsWriter object, the first time I call write_feature_weights, should it really
+        # change the default dataset id? I think this could introduce unexpected behaviors, and should be treated as a
+        # stateless call
```
richard-rogers: It only changes it if you ask it to... I think this is only for edge cases where you want a single writer instance to write multiple datasets.

python/whylogs/api/writer/whylabs.py
```diff
@@ -688,65 +676,95 @@ def write(self, file: Writable, **kwargs: Any) -> Tuple[bool, str]:
...
+        if stamp <= 0:
+            logger.error(
+                f"Profiles should have timestamps greater than 0, but found a timestamp of {stamp}"
+                f" and current timestamp is {utc_now.timestamp()}, this is likely an error."
+            )
+
+        return stamp
+
+    @deprecated_alias(profile="file")
+    def write(self, file: Writable, zip_file: bool = False, **kwargs: Any) -> Tuple[bool, str]:
```
richard-rogers: We can't change this signature because it's part of the `Writer` interface. We can pull `zip_file` from `kwargs` here and then pass it along explicitly to the private methods.

richard-rogers: Actually, I wonder if it would be better to put the zip functionality in the `Writable` subclasses? I guess the `WhyLabsWriter` would still need to know if the `Writable` is zipped, but that could be arranged...

python/whylogs/api/writer/whylabs.py
murilommen: chatting offline about this

python/whylogs/api/logger/result_set.py
```diff
@@ -475,6 +478,27 @@ def get_writables(self) -> Optional[List[Writable]]:
             )
         return results
 
+    def in_memory_zip(self) -> Optional[bytes]:
```
richard-rogers: Might be too big for in memory?

murilommen: True, in case we're dealing with a gigantic SegmentedResultSet, which can contain multiple SegmentedDatasetProfiles with a number of columns.
I'll turn it into a static method on WhyLabsWriter then, so other types of Writables can leverage it

murilommen: it will be an in-memory zipped file, that will be persisted to tmp_dir. same logic as before, only extracted out to its own method

python/tests/api/writer/test_whylabs.py
```diff
@@ -395,3 +398,35 @@ def test_custom_client(self) -> None:
...
         assert len(WhylabsClientCache.instance()._api_client_cache) == 0
+
+    @patch(
+        "whylogs.api.writer.whylabs.WhyLabsWriter._get_upload_url_segmented_reference",
+        return_value=CreateReferenceProfileResponse(id="profile1", upload_urls=["http://upload.url"]),
+    )
+    @patch("whylogs.api.writer.whylabs.WhyLabsWriter._do_upload", return_value=(True, "Success"))
+    @patch("whylogs.api.writer.whylabs.WhyLabsWriter._get_dataset_timestamp", return_value=1234567890)
+    @patch("whylogs.api.writer.whylabs.WhyLabsWriter._upload_zipped_files", return_value=(True, "Success"))
+    def test_write_segmented_reference_result_set(
```
richard-rogers, line 409: I would probably leave this test out... it won't survive the refactoring.

python/tests/api/writer/test_whylabs_integration.py
```diff
@@ -153,11 +153,16 @@ def test_feature_weights():
     retrieved_weights = writer.get_feature_weights()
     assert feature_weights.weights == retrieved_weights.weights
 
-    # swap 'em so we won't accidentally pass from previous state
-    feature_weights = FeatureWeights({"col1": 0.0, "col2": 1.0})
-    writer.write_feature_weights(feature_weights)
+    # making sure that the weights are actually updated
     retrieved_weights = writer.get_feature_weights()
-    assert feature_weights.weights == retrieved_weights.weights
+    assert retrieved_weights.weights == {"col1": 1.0, "col2": 0.0}
```
richard-rogers: This appears redundant?

python/whylogs/api/writer/whylabs.py
```diff
@@ -448,6 +481,15 @@ def tag_custom_performance_column(
             )
             return False, str(e)
 
+    def _tag_custom_perf_metrics(self, view: Union[DatasetProfileView, SegmentedDatasetProfileView]) -> None:
```
richard-rogers, line 483: Perhaps this should handle `SegmentedDatasetProfileView` or not be `Union`?

python/tests/api/writer/test_whylabs.py
murilommen: this "all mock" test has saved me some trouble, especially to understand exactly what are the method's calls. I'd vote for leaving it and just changing the patches -- or deleting it -- after the refactor

python/whylogs/api/writer/whylabs.py
murilommen: it is called within `_get_uncompounded_view`, which is a part of the flow for DatasetProfileView and SegmentedDatasetProfileView. but inside of it, it makes an `isinstance` check for DatasetProfileView only. very confusing, so I didn't want to modify it. out of the scope of this PR

FelipeAdachi:
Closing because an alternative PR was merged

Issue #1463, opened 2024-02-02
Pillow vulnerabilities


matsair:
## Description

There are multiple [osv](https://osv.dev/) vulnerability issues with the Pillow 9.x package dependency. An update to 10.x (https://github.com/whylabs/whylogs/blob/mainline/python/pyproject.toml#L68) is currently not possible.

Some vulnerabilities:
- https://osv.dev/vulnerability/GHSA-3f63-hfp8-52jq
- https://osv.dev/GHSA-56pw-mpj4-fxww
- https://osv.dev/GHSA-8ghj-p4vj-mr35

## Steps to reproduce:
  - `poetry install`
  - `osv-scanner --lockfile poetry.lock` (https://github.com/google/osv-scanner)

![image](https://github.com/whylabs/whylogs/assets/5218859/f9519864-ea7f-4671-82ca-4dfee40af389)

## Suggestions

Could we update Pillow to its latest version?

- [x] I have reviewed the [Guidelines for Contributing](https://github.com/whylabs/whylogs/blob/mainline/.github/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/whylabs/whylogs/blob/mainline/.github/CODE_OF_CONDUCT.md).

jamie256:
changes merged to allow newer versions of PIL and updates lock file. Slated for next whylogs release, 1.4.1

jamie256:
released in version 1.4.1

PR #1471, opened 2024-02-21
Update version ranges to allow newer Pillow and update lock file

jamie256:
## Description

Pillow 10.x has been out for a while and our use of it in the image extra works, so updating dependencies to allow version under 11 as compatible with whylogs.

## Changes

- Updated install dependency range on Pillow to allow 10.x
- Updated poetry.lock to run tests with 10.2

## Related

incidentally will get rid of vulnerabilities in detected in the poetry.lock file related to older version of Pillow and addresses #1463 


jamie256:
redoing on 1.4

PR #1475, opened 2024-03-04
Fix types and remove dead code in actor classes

naddeoa:
These changes are mostly no-ops and fixes for type issues that should have happened in a previous change. The base classes no longer needed these methods.


jamie256 (approved):
LGTM!

PR #1479, opened 2024-03-07
Update BentoML example

naddeoa:
This updates the example for BentoML to no longer use atexit or the original rolling logger, and it uses the more common (in BentoML examples) function based service interface. It also highlights logging to multiple dataset ids within a single service.

The atexit only really works for toy local examples. BentoML prefers that you use its lifecycle management hooks to do startup and shutdown related logic. The old logger couldn't gracefully handle multiple dataset ids either, it required changing the default dataset id between logging calls.


python/examples/integrations/bentoml/README.md
```diff
@@ -24,95 +25,111 @@ Next, In `service.py`, we'll get ready to serve and monitor our latest saved mod
...
 
+There are two version of the server in this example: `server_thread_logger.py` and `server_process_logger.py`. In general, the process based
+logger is better but it requires extra dependencies that only work on Mac/Linux that make the performance much better than the standard
+multiprocessing options in Python. The thread based version has a few disadvantages worth mentioning here:
 
-my_service = MyService()
-atexit.register(my_service.on_exit_callback)
+- It requires manually creating a logging instance for every dataset id you log to while the process based version handles that for you when
+  you call with `logger.log(data, dataset_id="..")`.
+- Python's GIL means that logging could contend with inference comptue.
```
jamie256:
```suggestion
- Python's GIL means that logging could contend with inference compute.
```

jamie256 (dismissed):
Looks good.

There is a reference to the whylogs[whylabs] extra which if that’s for pip install that extra is no longer required.

naddeoa:
@jamie256 updated to fix comments

jamie256 (approved):
LGTM!

PR #1481, opened 2024-03-13
ndcg refactor

### LINKED ISSUE #1480: log_batch_ranking_metrics error
## Description

You can repro using this notebook:
https://github.com/whylabs/whylogs/blob/mainline/python/examples/experimental/Writing_Ranking_Performance_Metrics_to_WhyLabs.ipynb

Hits an error in versions of whylogs after 1.3.21:
```
target_order = np.array(row_dict[target_column]).argsort()[::-1]
     99         dcg_vals = [
--> 100             (rel / math.log(i + 2, 2)) for i, rel in enumerate(np.array(row_dict[target_column])[predicted_order][:k])
    101         ]
    102         idcg_vals = [

IndexError: index 1 is out of bounds for axis 0 with size 1
```
whylogs 1.3.21 does not hit this.

FelipeAdachi:
## Description

This PR:

- Adds ndcg value expecations to the tests, not just counts
- Changes convert_non_numeric=True to be used for string columns (integers, floats in scores and bools should use the default convert_non_numeric=False )
- adds score_column support: when a score is available, like in [this example](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.ndcg_score.html), one should use score_column and target_column , and not prediction_column (see 2 last tests on the sklearn examples)
 - Changes the logic to generate target column values to make it compatible with all scenarios
- Fixes prediction and ideal relevance calculation for non numeric case
- Handles DivisionbyZero edge case when idcg=0 (ndcg set to 1 if no relevant documents exist)
- If K is not passed, metrics will be calculated according to predictions cols's length (and metrics named accordingly - k is no longer omitted in the names when k is None)

For the Numeric case:
- If predictions+target or scores+target columns are both provided, they need to be of same length.
- If prediction_column is provided with target column, prediction col contains the rank of suggested items, starting with 1
- If only prediction column is provided, it is assumed that the order encodes the ranks of recommendations: first item in the list is the first recommendation. The value in the list encodes the relevance score

### DIFF
--- a/python/tests/experimental/api/test_logger.py
+++ b/python/tests/experimental/api/test_logger.py
@@ -43,6 +43,8 @@ def test_log_batch_ranking_metrics_single_simple():
     assert pandas_summary.loc["norm_dis_cumul_gain", "counts/n"] == 1
     assert pandas_summary.loc["average_precision", "counts/n"] == 4
     assert pandas_summary.loc["norm_dis_cumul_gain", "counts/n"] == 1
+    # ndcg = [1, 0, 0.63, 0.5]
+    assert isclose(pandas_summary.loc[f"norm_dis_cumul_gain", "distribution/mean"], 0.53273, abs_tol=0.00001)
 
 
 def test_log_batch_ranking_metrics_binary_simple():
@@ -51,7 +53,7 @@ def test_log_batch_ranking_metrics_binary_simple():
     )
 
     result = log_batch_ranking_metrics(
-        data=binary_df, prediction_column="raw_predictions", k=2, convert_non_numeric=True
+        data=binary_df, prediction_column="raw_predictions", k=2
     )
     pandas_summary = result.view().to_pandas()
 
@@ -76,6 +78,8 @@ def test_log_batch_ranking_metrics_binary_simple():
     assert pandas_summary.loc["top_rank", "counts/n"] == 4
     assert pandas_summary.loc["average_precision_k_" + str(k), "counts/n"] == 4
     assert pandas_summary.loc["norm_dis_cumul_gain_k_" + str(k), "counts/n"] == 1
+    # ndcg@2 = [0.613147, 1.0, 1.0, 0.63093]
+    assert isclose(pandas_summary.loc["norm_dis_cumul_gain_k_" + str(k), "distribution/mean"], 0.81101, abs_tol=0.00001)
 
 
 def test_log_batch_ranking_metrics_multiple_simple():
@@ -121,15 +125,15 @@ def test_log_batch_ranking_metrics_multiple_simple():
     assert pandas_summary.loc["top_rank", "counts/n"] == 4
     assert pandas_summary.loc["average_precision_k_" + str(k), "counts/n"] == 4
     assert pandas_summary.loc["norm_dis_cumul_gain_k_" + str(k), "counts/n"] == 1
-
-    assert isclose(pandas_summary.loc[f"norm_dis_cumul_gain_k_{k}", "distribution/median"], 0.76244, abs_tol=0.00001)
+    # ndcg@4 = [0.9197, 0.0, 1.0, 0.386853]
+    assert isclose(pandas_summary.loc[f"norm_dis_cumul_gain_k_{k}", "distribution/median"], 0.57664, abs_tol=0.00001)
 
 
 def test_log_batch_ranking_metrics_default_target():
     multiple_df = pd.DataFrame({"raw_predictions": [[3, 2, 3, 0, 1, 2, 3, 2]]})
 
     result = log_batch_ranking_metrics(
-        data=multiple_df, prediction_column="raw_predictions", k=3, convert_non_numeric=True
+        data=multiple_df, prediction_column="raw_predictions", k=3
     )
     pandas_summary = result.view().to_pandas()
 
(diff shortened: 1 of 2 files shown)


python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -22,26 +29,40 @@ def log_batch_ranking_metrics(
...
+            raise ValueError("Either prediction_column or score+target columns must be specified")
+
     relevant_cols = [prediction_column]
+
     if target_column is None:
+        formatted_data = _convert_to_int_if_bool(formatted_data, prediction_column)
         target_column = "__targets"
-        formatted_data[target_column] = formatted_data[prediction_column].apply(lambda x: list(range(len(x)))[::-1])
+        # formatted_data[target_column] = formatted_data[prediction_column].apply(lambda x: list(range(len(x)))[::-1])
+        # formatted_data[target_column] = [[] for _ in range(len(formatted_data[prediction_column]))]
```
jamie256: lets remove these commented out lines

python/tests/experimental/api/test_logger.py
```diff
@@ -154,11 +154,13 @@ def test_log_batch_ranking_metrics_default_target():
...
     assert pandas_summary.loc["average_precision_k_" + str(k), "counts/n"] == 1
     assert pandas_summary.loc["norm_dis_cumul_gain_k_" + str(k), "counts/n"] == 1
+    # ndcg@3 = [0.9013]
+    assert isclose(pandas_summary.loc[f"norm_dis_cumul_gain_k_{k}", "distribution/median"], 0.90130, abs_tol=0.00001)
 
 
 def test_log_batch_ranking_metrics_ranking_ndcg_wikipedia():
     # From https://en.wikipedia.org/wiki/Discounted_cumulative_gain#Example
-    ranking_df = pd.DataFrame({"targets": [[3, 2, 3, 0, 1, 2, 3, 2]], "predictions": [[7, 6, 5, 4, 3, 2, 1, 0]]})
+    ranking_df = pd.DataFrame({"targets": [[3, 2, 3, 0, 1, 2, 3, 2]], "predictions": [[3, 2, 3, 0, 1, 2]]})
```
jamie256: I find the terminology confusing here, these aren't really targets and predictions, it seems what we need is a single column of relevancies for the ndcg calculation when they are numeric. Do we use both the predictions and targets here?

python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -22,26 +29,40 @@ def log_batch_ranking_metrics(
 ) -> ViewResultSet:
     formatted_data = data.copy(deep=True)  # TODO: does this have to be deep?
 
+    if prediction_column is None:
+        if score_column is not None and target_column is not None:
+            prediction_column = "__predictions"
+
+            # sort data[prediction_column] by score_column
+            def _sort_by_score(row):
```
murilommen: maybe don't create a nested function for a single line

python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -22,26 +29,40 @@ def log_batch_ranking_metrics(
...
+        else:
+            raise ValueError("Either prediction_column or score+target columns must be specified")
+
     relevant_cols = [prediction_column]
+
     if target_column is None:
+        formatted_data = _convert_to_int_if_bool(formatted_data, prediction_column)
         target_column = "__targets"
-        formatted_data[target_column] = formatted_data[prediction_column].apply(lambda x: list(range(len(x)))[::-1])
+        # formatted_data[target_column] = formatted_data[prediction_column].apply(lambda x: list(range(len(x)))[::-1])
```
murilommen: delete comments?

python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -76,41 +95,43 @@ def get_top_rank(row):
...
-        ]
-        idcg_vals = [
-            (rel / math.log(i + 2, 2)) for i, rel in enumerate(np.array(row_dict[target_column])[target_order][:k])
-        ]
+        predicted_relevances = row_dict["predicted_relevance"]
+        ideal_relevances = sorted(row_dict["ideal_relevance"], reverse=True)
+        dcg_vals = [(rel / math.log(i + 2, 2)) for i, rel in enumerate(predicted_relevances[:k])]
+        idcg_vals = [(rel / math.log(i + 2, 2)) for i, rel in enumerate(ideal_relevances[:k])]
+        if sum(idcg_vals) == 0:
+            return 1  # if there is no relevant data, not much the recommender can do
```
jamie256, line 128: returning 1 is surprising from a ranking metrics perspective, for monitoring this might make some sense, but we will need to document the convention here.

In a follow up we might consider counting the number of rows this occurs, as it could indicate a problem in the data similar to counting nulls.

python/tests/experimental/api/test_logger.py
```diff
@@ -43,16 +43,16 @@ def test_log_batch_ranking_metrics_single_simple():
     assert pandas_summary.loc["norm_dis_cumul_gain", "counts/n"] == 1
     assert pandas_summary.loc["average_precision", "counts/n"] == 4
     assert pandas_summary.loc["norm_dis_cumul_gain", "counts/n"] == 1
+    # ndcg = [1, 0, 0.63, 0.5]
```
jamie256, line 46: This is a helpful showing the expected values, thanks

jamie256 (commented):
Left a few questions and minor comments. Thanks for working on these!

python/tests/experimental/api/test_logger.py
FelipeAdachi: yes, it is confusing - Let's say we have a list of documents D_0 to D_7. The `predictions` col tells us that the recommendation system recommended [D_0,D_1,D_2,D_3,D_4,D5], and their relevance (obtained, e.g., by asking humans) is [3,2,3,0,1,2]. But we have two extra documents in our pool that were not suggested - D_6 and D_7. Those two are in the `targets` col, which contains the relevance for all documents.

So, both predictions and target are used here - in particular, the order in the predicitons col is important to calculate DCG, and targets is important because it contains the relevances for all documents, even those not recommended, which is used to calculate IDCG.

@bernease please check if that matches your understanding on how `predictions` and `targets` should be passed.

FelipeAdachi: On a more careful consideration, the api as suggested by the removed line might make more sense - the predictions being indices to the documents in the `targets` list - However, to match the example in Wikipedia, maybe `predictions` should be: `[0, 1, 2, 3, 4, 5]`?

jamie256 (approved):
Looks better, thanks!

PR #1484, opened 2024-03-15
WhyLabsWriter refactor

richard-rogers:
## Description

`Writable` represents output to be serialized by whylogs. It will write itself to 1 or more temporary files. A `Writer` takes the temporary file(s) and sends them to their intended destination. The interface allows any `Writer` to handle any `Writable`, but in practice some `Writer`s are only interested in certain types of `Writable`s.

`WhyLabsWriter` acts as the main entry point for sending data to WhyLabs. It makes use of a new `WhyLabsClient` interface for interacting with WhyLabs REST APIs. `WhyLabsWriter` has some deprecated methods that duplicate `WhyLabsClient` functionality the sake of backwards compatibility. `WhyLabsWriter` delegates to the `WhyLabsTransactionWriter`, `WhyLabsBatchWriter`, and `WhyLabsReferenceWriter` classes according to the `WhyLabsWriter::write()` use case. The 3 classes correspond to the 3 REST API endpoints for uploading profiles (transaction, batch/async, and reference).

`WhyLabsWriterBase` implements a few utility methods shared by the various `WhyLabs*Writer` classes. In particular, `_prepare_view_for_upload()` handles processing required before uploading a profile (uncompounding, custom performance metric tagging). `_send_writable_to_whylabs()` serializes a profile in either V0 or V1 format and uploads it to WhyLabs. `_upload_view()` is a convenience method that just calls `_prepare_view_for_upload()` then `_send_writable_to_whylabs()`. `WhyLabsWriter::write()` accepts a variety of data structures representing a profile: `ViewResultSet`, `ProfileResultSet`, `SegmentedResultSet`, `DatasetProfile`, `DatasetProfileView`, and `SegmentedDatasetProfileView`. `WhyLabsWriterBase::_get_view_of_writable()` converts all of those except `SegmentedResultSet` to either `DatasetProfileView` or `SegmentedDatasetProfileView`, which represent a single profile/segment to be uploaded to WhyLabs. The `WhyLabs*Writer` classes generally iterate over `SegmentedResultSet` uploading each segment.

* Transactions do not support zipped profiles or reference profiles
* Segmented batch or reference profiles can be zipped by adding `zip=True` argument to `write()`

## Changes

TODO: describe API changes

## Related

[zipped batch profiles](https://app.clickup.com/t/86azmk2t6)


python/whylogs/core/feature_weights.py
```diff
@@ -1,11 +1,11 @@
...
-from typing import Any, Dict, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
-from whylogs.api.writer import Writers
-from whylogs.api.writer.writer import Optional, Writable, Writer
+from whylogs.api.writer.writer import Writable
 from whylogs.core import Segment
 
 
+# TODO: In whlogs 2.X this should not be a Wriable, it should be sent via WhyLabsClient
```
FelipeAdachi: If we're removing this way of uploading feature weights, what is the recommended way? Can we have an integration test showing the suggested way?

Also, what is the rationale for feature weights not being a Writable? I'm not opposed to this, just curious.

python/whylogs/experimental/performance_estimation/estimation_results.py
FelipeAdachi, line 1: these changes will probably break the perf estimation example, right?

since it's in experimental, it might be ok, but we need to have a way of uploading the perf estimation results to be able to update the example

python/whylogs/api/logger/result_set.py
```diff
@@ -183,10 +178,6 @@ def reader(name: str = "local") -> "ResultSetReader":
         reader = Readers.get(name)
         return ResultSetReader(reader=reader)
 
-    def writer(self, name: str = "local") -> "ResultSetWriter":
```
FelipeAdachi, line 186: we have a number of examples where we use the `.writer("whylabs").write()`, like Segmens, Image_Logging, etc. This will be a breaking change, and we'll need to update the examples accordingly

python/whylogs/api/writer/mlflow.py
```diff
@@ -26,38 +26,50 @@ def write(
...
 
     def _get_temp_directory(self, dest: Optional[str] = None):
         tmp_dir = mkdtemp()
         output_dir = os.path.join(tmp_dir, self._file_dir)
         os.makedirs(output_dir, exist_ok=True)
         if dest:
             output = os.path.join(output_dir, dest)
-        elif self._file_name:
-            output = os.path.join(output_dir, self._file_name)
+            # elif self._file_name:
```
FelipeAdachi: lose the comented lines?

python/whylogs/api/logger/result_set.py
richard-rogers: I moved `.writer()` into the `Writable` base class and made a generic `WriterWrapper` so that we don't need a separate `*Writer` for every `Writable` subclass. This should be backwards compatible.

python/whylogs/api/writer/mlflow.py
richard-rogers:
```suggestion
```

python/whylogs/api/writer/mlflow.py
```diff
@@ -26,38 +26,50 @@ def write(
...
+            #    output = os.path.join(output_dir, self._file_name)
```
richard-rogers:
```suggestion
```

python/whylogs/api/writer/whylabs_base.py
```diff
@@ -0,0 +1,256 @@
...
+            api_key,
+            dataset_id,
+            api_client,
+            ssl_ca_cert,
+            _timeout_seconds,
+        )
+        session = default_init()  # Force an init if the user didn't do it, it's idempotent
+        config = session.config
+
+        self._reference_profile_name = config.get_whylabs_refernce_profile_name()
```
richard-rogers, line 88: Keeping these in sync with their values in `_whylabs_client` is error-prone. Keep them here and remove them from `WhyLabsClient`. Just pass them in to the methods that require them.

python/whylogs/api/writer/writer.py
```diff
@@ -1,27 +1,56 @@
...
+        Returns the default path to write to, excluding filename(s). None = use CWD
+        """
+        return None
+
     @abstractmethod
-    def write(self, path: Optional[str] = None, **kwargs: Any) -> Tuple[bool, str]:
-        pass
+    def write(
+        self, path: Optional[str] = None, filename: Optional[str] = None, **kwargs: Any
+    ) -> Tuple[bool, Union[str, List[str]]]:  # TODO: Union[str, List[Tupble[bool, str]]] ?
```
richard-rogers, line 42: The addition of a separate `filename` is a breaking change. I don't think many people are/should call[ing] `Writable::write()` directly though?

python/whylogs/api/writer/writer.py
```diff
@@ -34,15 +63,88 @@ def check_interval(self, interval_seconds: int) -> None:
...
+
     @abstractmethod
     def write(
         self,
         file: Writable,
         dest: Optional[str] = None,
         **kwargs: Any,
-    ) -> Tuple[bool, str]:
-        pass
+    ) -> Tuple[bool, Union[str, List[Tuple[bool, str]]]]:
```
richard-rogers, line 83: What do we think about the optional list of (status, message) for multi-file `write()`s?

python/whylogs/core/feature_weights.py
richard-rogers: Recommended way is `WhyLabsClient::write_feature_weights()`. I will add the integration test.

The rationale is that feature weights are specific to WhyLabs rather than being whylogs data.

python/whylogs/experimental/performance_estimation/estimation_results.py
richard-rogers: I don't think it will break. The `write()` was never called, the results were turned into a profile and that was written.

python/tests/api/writer/test_whylabs_integration.py
```diff
@@ -60,17 +62,19 @@ def test_whylabs_writer():
...
+def test_whylabs_writer_segmented(zipped: bool):
     ORG_ID = os.environ.get("WHYLABS_DEFAULT_ORG_ID")
     MODEL_ID = os.environ.get("WHYLABS_DEFAULT_DATASET_ID")
     why.init(force_local=True)
     schema = DatasetSchema(segments=segment_on_column("col1"))
     data = {"col1": [1, 2, 1, 3, 2, 2], "col2": ["foo", "bar", "wat", "foo", "baz", "wat"]}
     df = pd.DataFrame(data)
     trace_id = str(uuid4())
-    result = why.log(df, schema=schema, trace_id=trace_id)
+    profile = why.log(df, schema=schema, trace_id=trace_id)
```
murilommen, line 84: isn't log returning a `ResultSet` anymore?

python/tests/api/writer/test_whylabs_integration.py
```diff
@@ -218,6 +225,35 @@ def test_transactions():
     assert deserialized_view.get_columns().keys() == data.keys()
 
 
+@pytest.mark.load
+def test_transaction_aborted():
```
murilommen, line 279: maybe this needs to come from a separate PR?

python/tests/api/writer/test_whylabs_integration.py
```diff
@@ -262,6 +299,23 @@ def test_transaction_context():
         assert deserialized_view is not None
 
 
+@pytest.mark.load
+def test_transaction_context_aborted():
```
murilommen, line 389: same

python/whylogs/api/logger/__init__.py
```diff
@@ -261,8 +261,11 @@ def reader(name: str) -> ResultSetReader:
     return ResultSet.reader(name=name)
 
 
-def write(profile: DatasetProfile, base_dir: str) -> None:
-    return profile.write(base_dir)
+def write(profile: DatasetProfile, base_dir: Optional[str] = None, filename: Optional[str] = None) -> None:
+    if base_dir and filename:
+        return profile.write(base_dir, filename)
+    else:
+        return profile.write("", base_dir)  # backward compatibility: use base_dir as full filename
```
murilommen, line 327: this is a little confusing. why did you change the order of arguments for each write call? probably better to use kwargs

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger.py
```diff
@@ -93,7 +93,7 @@ def create_writers(self, dataset_id: str) -> List[Writer]:
 class WhyLabsWriterFactory(WriterFactory):
     def create_writers(self, dataset_id: str) -> List[Writer]:
         return [
-            Writers.get(
+            Writers.get(  # type: ignore
```
murilommen, line 96: nit: inform the pep8 you're ignoring can be helpful, i.e.: `# type: ignore E405`. the more we can avoid plain type ignores the better

python/whylogs/api/logger/result_set.py
```diff
@@ -352,6 +343,114 @@ def profile(self, segment: Optional[Segment] = None) -> Optional[Union[DatasetPr
...
+            else:
+                logger.warning(
+                    f"Found no segments in partition: {first_partition} even though we have: {self.count} segments overall"
+                )
+            logger.info(f"From list of: {self.count} SegmentedDatasetProfileViews using {len(results)}")
+        else:
+            logger.warning(
+                f"Attempt to build segmented results for writing but there are no segments in this result set: {self._segments}. returning None."
+            )
+        return results
```
murilommen, line 388: there's too many ifs on this method not to think about breaking it down into smaller ones and making `get_writables` an "orchestrator"

python/whylogs/api/logger/result_set.py
```diff
@@ -352,6 +343,114 @@ def profile(self, segment: Optional[Segment] = None) -> Optional[Union[DatasetPr
...
+            # TODO: handle more than one partition
+            if not self.partitions:
+                raise ValueError(
+                    f"Building list of: {self.count} SegmentedDatasetProfileViews in SegmentedResultSet but no partitions found: {self.partitions}."
+                )
+            if len(self.partitions) > 1:
+                logger.error(
+                    f"Building list of: {self.count} SegmentedDatasetProfileViews in SegmentedResultSet but found more than one partition: "
+                    f"{self.partitions}. Using first partition only!!"
+                )
```
murilommen, line 360: i.e.: this could be a validation method for partitions specifically

python/whylogs/api/logger/result_set.py
```diff
@@ -352,6 +343,114 @@ def profile(self, segment: Optional[Segment] = None) -> Optional[Union[DatasetPr
...
+            first_partition = self.partitions[0]
+            segments = self.segments_in_partition(first_partition)
+            if segments:
```
murilommen, line 363: a little confusing to have `self._segments` and `segments` here. what is `segments_in_partition` doing? just naming them more precisely to their actions might be enought

python/whylogs/api/logger/result_set.py
```diff
@@ -352,6 +343,114 @@ def profile(self, segment: Optional[Segment] = None) -> Optional[Union[DatasetPr
...
+        if views is None:
+            logger.warning("SegmentedResultSet contains no segments")
+            return []
+
+        times = list()
+        for view in views:
+            times.append(view.dataset_timestamp)
+        return times
+
+    def _get_default_filename(self) -> str:
```
murilommen, line 431: not necessary?

python/whylogs/api/writer/whylabs_batch_writer.py
```diff
@@ -0,0 +1,178 @@
...
+        # TODO: special handling of large number of files, handle throttling
+        for view in views:
+            dataset_timestamp_epoch = self._get_dataset_epoch(view, utc_now)
+            profile_id, upload_url = self._whylabs_client.get_upload_url_batch(dataset_timestamp_epoch)  # type: ignore
+            bool_status, message = self._upload_view(view, profile_id, upload_url, dataset_timestamp_epoch)
+            messages.append(message)
+            and_status = and_status and bool_status
+
+        logger.debug(f"Completed writing {len(views)} files!")
+        return and_status, ("; ".join(messages) if and_status else "Failed to upload all segments")
```
murilommen, line 91: drop this

python/whylogs/api/writer/writer.py
murilommen: if it's an optional argument and defaults to None, how can it break things?

python/whylogs/api/writer/writer.py
murilommen: d'accord

python/whylogs/api/writer/writer.py
richard-rogers: Changed semantics... the filename used to be included in `path`

python/tests/api/writer/test_whylabs_integration.py
richard-rogers: It returns a `ResultSet` or `SegmentedResultSet`.

python/whylogs/api/logger/__init__.py
richard-rogers: Trying to preserve backward compatibility...

python/whylogs/api/logger/result_set.py
richard-rogers: I don't disagree, but this code was just moved rather than rewritten.

FelipeAdachi (commented):
I found two examples that breaks with the changes in this PR:

- performance_estimation.ipynb

error:
`AttributeError: 'WhyLabsEstimationResultWriter' object has no attribute '_org_id'`

It looks like `_org_id` is now under `_whylabs_client._org_id`

- Writing_Reference_Profiles_to_WhyLabs.ipynb

error:
`❌ Failed to upload reference profile: cannot unpack non-iterable bool object
  File "/mnt/c/Users/felip/Documents/Projects-WhyLabs/whylogs/python/whylogs/api/whylabs/session/notebook_logger.py", line 104, in notebook_session_log
    result = session.upload_reference_profiles(profiles)
  File "/mnt/c/Users/felip/Documents/Projects-WhyLabs/whylogs/python/whylogs/api/whylabs/session/session.py", line 283, in upload_reference_profiles
    results.append(*[id for status, id in result if status])
  File "/mnt/c/Users/felip/Documents/Projects-WhyLabs/whylogs/python/whylogs/api/whylabs/session/session.py", line 283, in <listcomp>
    results.append(*[id for status, id in result if status])
TypeError: cannot unpack non-iterable bool object`

On a superficial inspection, the cause seems to be the output of ResultSet's `write()`, which used to be a list of tuples. Now the result seems to be a single tuple.

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger.py
richard-rogers, line 96:
```suggestion
            Writers.get(  # type: ignore E405
```

richard-rogers: It doesn't like that ignore syntax

python/whylogs/api/logger/result_set.py
richard-rogers: This code was just moved, not re-written.

python/whylogs/api/logger/result_set.py
richard-rogers: It's an unused abstract method, so it has to exist.

python/whylogs/api/writer/whylabs_batch_writer.py
richard-rogers: If we agree there's no use case for non-transactional segmented batch profiles...

python/whylogs/api/writer/whylabs_estimation_result_writer.py
```diff
@@ -0,0 +1,101 @@
...
+
+    @deprecated_alias(profile="file")
+    def write(
+        self, file: Writable, dest: Optional[str] = None, **kwargs: Any
+    ) -> Tuple[bool, Union[str, List[Tuple[bool, str]]]]:
+        if not isinstance(file, EstimationResult):
+            raise ValueError("I can only write EstimationResult")
+
+        if _uncompound_performance_estimation_feature_flag():
+            trace_id = kwargs.get("trace_id")
```
richard-rogers, line 83: I added `trace_id` kwarg for integration testing.

python/whylogs/api/logger/experimental/logger/actor/process_rolling_logger.py
murilommen: flake8 uses `# noqa: E{pep8 error you chose to ignore}`

richard-rogers:
> I found two examples that breaks with the changes in this PR:
> * performance_estimation.ipynb
> error: `AttributeError: 'WhyLabsEstimationResultWriter' object has no attribute '_org_id'`

This one's fixed. I also found a bug where it possible to try to write an `EstimationResult` with a `None` accuracy. Is there a use car for `None` accuracies?

> * Writing_Reference_Profiles_to_WhyLabs.ipynb
> 
> error: `❌ Failed to upload reference profile: cannot unpack non-iterable bool object ...

Also fixed.

python/examples/basic/Getting_Started.ipynb
```diff
@@ -75,7 +75,7 @@
    "outputs": [],
    "source": [
     "# Note: you may need to restart the kernel to use updated packages.\n",
-    "%pip install whylogs"
+    "%pip install whylogs==1.3.29.dev0"
```
jamie256, line 78: undo edits to the Getting Started example if not needed, we don't want to pin the version to a dev build

python/whylogs/core/feature_weights.py
jamie256: remove comment, typos

python/whylogs/api/writer/__init__.py
```diff
@@ -1,29 +1,8 @@
-from typing import Any
-
-from whylogs.api.writer.writer import Writer
-
-
-class Writers:
```
jamie256, line 6: Is this moved somewhere or did you remove this public convenience method?

jamie256 (changes requested):
Please summarize what parts of the public API are being removed and deprecated with this change. We have to work through how much this affects the public surface area of our API and decide on major/minor version bump before we can merge these changes.

python/examples/basic/Getting_Started.ipynb
richard-rogers: Will do in a follow-up PR immediately after this is merged

python/whylogs/api/writer/__init__.py
richard-rogers: I replaced all the `Writable`-specific wrapper classes with a single wrapper. The API still works.

python/whylogs/core/feature_weights.py
richard-rogers:
```suggestion
```

richard-rogers:
> Please summarize what parts of the public API are being removed and deprecated with this change. We have to work through how much this affects the public surface area of our API and decide on major/minor version bump before we can merge these changes.

The main breaking change is in the `Writable` interface, which hopefully no-one is using:
```
    def write(self, path: Optional[str] = None, **kwargs: Any) -> Tuple[bool, str]:
```
becomes
```
    def write(
        self, path: Optional[str] = None, filename: Optional[str] = None, **kwargs: Any
    ) -> Tuple[bool, Union[str, List[str]]]:  # TODO: Union[str, List[Tupble[bool, str]]] ?
```
`Writable`s were inconsistent about whether the original `path` was a directory to write to or the full filename, so now they're explicitly separate. A `Writable` can now write multiple files (e.g., `SegmentedResultSet`), so it can return a list of files written. 

The `writer()` convenience method is also moved into the `Writable` class, and returns a generic wrapper that can handle any combination of `Writable` and `Writer` rather than having a separate wrapper class for each `Writable`.

`Writer::write()` can also optionally return a list of results when it writes multiple things:
```
    def write(
        self,
        file: Writable,
        dest: Optional[str] = None,
        **kwargs: Any,
-  ) -> Tuple[bool, str]:
+  ) -> Tuple[bool, Union[str, List[Tuple[bool, str]]]]:
```

python/whylogs/api/writer/__init__.py
jamie256: ok, its moved I see into the writers.py

jamie256:
> Please summarize what parts of the public API are being removed and deprecated with this change. We have to work through how much this affects the public surface area of our API and decide on major/minor version bump before we can merge these changes.

Ok its not breaking examples or public API, but adding functionality and some behavior cleanup, I think we can release in 1.4.0rc0

jamie256 (approved):
LGTM for 1.4.0

PR #1486, opened 2024-03-15
Ranking Metrics - Mean Average Precision corrections

FelipeAdachi:
## Description

- corrected Average Precision calculation
- added handling for AP calculation for both numeric/non-numeric cases
- Removed `mean_average_precision` column, as this value is already present in `average_precision['distribution/mean']`
- Updated tests to match expected values for MAP
- If k passed is higher than predictions dimension, _max_k is used


python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -17,6 +17,48 @@ def _convert_to_int_if_bool(data: pd.core.frame.DataFrame, *columns: str) -> pd.
     return data
 
 
+def _calculate_average_precisions(formatted_data, target_column, prediction_column, convert_non_numeric: bool, k: int):
```
richard-rogers: needs type annotations

python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -62,6 +104,11 @@ def log_batch_ranking_metrics(
     _max_k = formatted_data[prediction_column].apply(len).max()
     if not k:
         k = _max_k
+    if k > _max_k:
+        diagnostic_logger.warning(
+            f"Max value of k in the dataset is {_max_k}, but k was set to {k}. Setting k to {_max_k}"
+        )
+        k = _max_k
```
richard-rogers, line 117: check k <= 0 ?

jamie256 (approved):
LGTM!

PR #1492, opened 2024-03-20
Ranking Metrics - Better Precision/Recall/MRR calculation

FelipeAdachi:
## Description

This PR:

- Makes the metrics calculation (other than NDCG and AP, which are already correct) correct for boolean and numeric scenarios
- Changes ndcg from a non-mergeable batch metric to a mergeable row-wise metric
- Corrects top_rank to return None instead of 0 when there is no relevant item (since lower is better for top_rank, assigning a 0 is not correct)
- Changes MRR from a non-mergeable batch metric to a mergeable row-wise metric (and renamed from `Mean Reciprocal Rank` to `Reciprocal Rank`)
- Refactors the code to group row-wise statistics calculations functions

After these changes, the only metric that makes the ranking metrics non-mergeable is `accuracy_k`


python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -17,6 +17,74 @@ def _convert_to_int_if_bool(data: pd.core.frame.DataFrame, *columns: str) -> pd.
     return data
 
 
+class RowWiseMetrics:
```
richard-rogers, line 20: needs type annotations

richard-rogers (commented):
Around line 143 in `__init__.py` a doc string explaining when to use which arguments would be helpful.

jamie256 (approved):
LGTM!

PR #1501, opened 2024-04-16
Add documentation for log classification/regression/ranking metrics

FelipeAdachi:
## Description

Adds and improves documentation in docstring format for log_regression/log_classification/log_ranking metrics.

### DIFF
--- a/python/whylogs/api/logger/__init__.py
+++ b/python/whylogs/api/logger/__init__.py
@@ -170,15 +170,46 @@ def log_classification_metrics(
     Function to track metrics based on validation data.
     user may also pass the associated attribute names associated with
     target, prediction, and/or score.
+
     Parameters
     ----------
-    targets : List[Union[str, bool, float, int]]
-        actual validated values
-    predictions : List[Union[str, bool, float, int]]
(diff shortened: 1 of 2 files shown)


python/whylogs/api/logger/__init__.py
```diff
@@ -170,15 +170,46 @@ def log_classification_metrics(
...
+        Dataframe with the data to log.
+    target_column : str
+        Column name for the actual validated values.
+    prediction_column : str
+        Column name for the predicted values.
+    score_column : Optional[str], optional
+        Associated scores for each inferred, all values set to 1 if None, by default None
+    schema : Optional[DatasetSchema], optional
+        Defines the schema for tracking metrics in whylogs, by default None
+    log_full_data : bool, optional
```
jamie256, line 186: Can we add a comment that typically in a production use case the ground truth or labeled data is required and isn't available at the same time the original data was generated, and to prevent double profiling the input features you should leave this as the default value of False.

jamie256 (dismissed):
LGTM!

jamie256 (approved):

PR #1503, opened 2024-04-20
ConditionCountMetric fix and docstrings

richard-rogers:
## Description

Adds docstrings & fix failed condition logging & exception bug.

### DIFF
--- a/python/whylogs/core/metrics/condition_count_metric.py
+++ b/python/whylogs/core/metrics/condition_count_metric.py
@@ -58,6 +58,27 @@ def not_relation(relation: Callable[[Any], bool]) -> Callable[[Any], bool]:
 
 @dataclass(frozen=True)
 class Condition:
+    """
+    Condition to be evaluated by the ConditionCountMetric.
+
+    Parameters
+    ----------
+    relation: Callable[[Any], bool]
+        The predicate to evaluate. The callable is passed a value from the column the
+        ConditionCountMetric is attached to, and returns True if the value satisfies
+        the condition.
+    throw_on_failure: bool
+        If throw_on_failure is true, whylogs will immediately raise a ValueError if
+        data that does not satisfy the condition is logged.
+    log_on_failure: bool
+        If log_on_failure is true, whylogs will log a warning message if data that does not
+        satisfy the conditon is logged.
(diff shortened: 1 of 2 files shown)


python/whylogs/core/metrics/condition_count_metric.py
```diff
@@ -147,12 +211,15 @@ def columnar_update(self, data: PreprocessedColumn) -> OperationResult:
...
 
         self.total.set(self.total.value + count)
-        if failed_conditions:
+        for failed in failed_conditions:
+            condition = self.conditions[failed]
             if condition.log_on_failure:
-                logger.warning(f"Condition(s) {', '.join(failed_conditions)} failed")
+                logger.warning(f"Condition {failed} failed")
 
+        for failed in failed_conditions:
```
jamie256: Can you explain why this is correct and what the bug was before? It looks like previously the error raised would contain a message for each failed condition in the set of failed conditions on that column's data, but now the error message would reference the first failed condition?

richard-rogers: The bug was that the `condition` used in the logging or throwing code was previously the last condition evaluated in the loop on line 136/200, whether it was the among the failed conditions or not.  I can restore the behavior of including all the failed throwing conditions if required.

jamie256: I think we should preserve returning the set of failures.

I don't follow that bug description, it looks like the failed_conditions is a set of the failures isn't it in the original near line 152:

`logger.warning(f"Condition(s) {', '.join(failed_conditions)} failed")`

richard-rogers: It was a scope bug... the `condition` tested in the `if`s for logging and throwing was the `condition` from the last iteration of the evaluation loop, whether that condition was configured to log or throw (or even if it matched).

I restored the `join`s on the offending condition names, though segregated for those that should log and those that should throw.

jamie256: great! ok, I see its the throws/not throws wasn't respecting the per condition setting. I thought you were claiming it ignored the failure of condition checks on earlier conditions (when they are all set to throw on error).

jamie256 (approved):
LGTM!

PR #1507, opened 2024-04-23
Ranking metrics API improvements

### LINKED ISSUE #1504: Remove batch_ranking_metrics' convert_non_numeric
## Description

In `log_batch_ranking_metrics`, if passing a list of strings to both `prediction` and `target`, `convert_non_numeric` should be set to true. Instead, this should be done automatically by checking the types of both columns. By doing this, we can remove the `convert_non_numeric` parameter, which simplifies our API.

### LINKED ISSUE #1505: Ranking Metrics - Passing the relevances through prediction in Single Column Scenario
## Description

One of the supported cases for Ranking Metrics is passing a single column, which contains the relevance values sorted by the rank:

`df_multi_single_column = pd.DataFrame({"predictions": [[3, 2, 3, 0, 1, 2, 3, 2]]})`

## Suggestions

Passing this through the `predictions` column is confusing, considering we are passing Relevance scores. This case should be supported by passing the same list through the `targets` columns instead of the `predictions` column.

## Related

richard-rogers:
## Description

Removes the `convert_non_numeric` argument and swap `prediction_column` for `target_column` when only relevance scores are passed.

## Changes

- Determines `convert_non_numeric` automatically by looking for strings in the prediction & target columns
- Swap `prediction_column` and `target_column` when only one column name is passed

Closes whylabs/whylogs#1504
Closes whylabs/whylogs#1505

### DIFF
--- a/python/whylogs/experimental/api/logger/__init__.py
+++ b/python/whylogs/experimental/api/logger/__init__.py
@@ -244,7 +244,7 @@ def log_batch_ranking_metrics(
 
         binary_single_df = pd.DataFrame(
             {
-                "raw_target": [
+                "raw_targets": [
                     [True, False, True], # First recommended item: Relevant, Second: Not relevant, Third: Relevant
                     [False, False, False], # None of the recommended items are relevant
                     [True, True, False], # First and second recommended items are relevant
--- a/python/whylogs/experimental/api/logger/__init__.py
+++ b/python/whylogs/experimental/api/logger/__init__.py
@@ -257,6 +257,9 @@ def log_batch_ranking_metrics(
     """
     formatted_data = data.copy(deep=True)  # TODO: does this have to be deep?
 
+    if score_column is not None and prediction_column is not None:
+        raise ValueError("Cannot specify both score_column and prediction_column")
+
     if prediction_column is None and score_column is None and target_column is not None:
         # https://github.com/whylabs/whylogs/issues/1505
         # The column use logic is complex, so just swapping them here for this case
--- a/python/whylogs/experimental/api/logger/__init__.py
+++ b/python/whylogs/experimental/api/logger/__init__.py
@@ -131,17 +131,38 @@ def _calculate_average_precisions(
     return averages
 
 
+def _all_strings(data: pd.Series) -> bool:
+    return all([all([isinstance(y, str) for y in x]) for x in data])
+
+
 def log_batch_ranking_metrics(
     data: pd.core.frame.DataFrame,
     prediction_column: Optional[str] = None,
     target_column: Optional[str] = None,
     score_column: Optional[str] = None,
     k: Optional[int] = None,
-    convert_non_numeric=False,
     schema: Union[DatasetSchema, None] = None,
     log_full_data: bool = False,
 ) -> ViewResultSet:
     """Log ranking metrics for a batch of data.
+
+    You can call the function several ways:
+      - Pass both prediction_column and target_column.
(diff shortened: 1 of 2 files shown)


python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -226,19 +244,25 @@ def log_batch_ranking_metrics(
...
             }
         )
 
-        result = log_batch_ranking_metrics(data=binary_single_df, prediction_column="raw_predictions", k=3)
+        result = log_batch_ranking_metrics(data=binary_single_df, target_column="raw_targets", k=3)
 
     """
     formatted_data = data.copy(deep=True)  # TODO: does this have to be deep?
 
+    if prediction_column is None and score_column is None and target_column is not None:
```
richard-rogers, line 263:
```suggestion
    if score_column is not None and prediction_column is not None:
        raise ValueError("Cannot specify both score_column and prediction_column")

    if prediction_column is None and score_column is None and target_column is not None:
```

python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -226,19 +244,25 @@ def log_batch_ranking_metrics(
...
+        # https://github.com/whylabs/whylogs/issues/1505
+        # The column use logic is complex, so just swapping them here for this case
+        # rather than unraveling all the use cases.
+        prediction_column, target_column = target_column, prediction_column
```
richard-rogers, line 267: I think this is backwards-compatible, i.e., you can still call it with only `prediction_column` and get the original behavior.

python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -131,18 +131,39 @@ def _calculate_average_precisions(
     return averages
 
 
+def _all_strings(data: pd.Series) -> bool:
```
murilommen, line 134: wondering if pandas methods could be a little more efficient, like: https://stackoverflow.com/a/43191959/8830597

murilommen (dismissed):
lgtm, just one small comment

python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -131,18 +131,39 @@ def _calculate_average_precisions(
...
+            are actually relevant. In this case, relevance is boolean.
+
+          - The prediction column contains lists of integers and the target column contains lists of numbers
+            or booleans. The value at the i-th position in the predicted list is the predicted rank of the i-th
+            element of the domain. The value at the i-th position in the target list is the true relevance score of the
+            i-th element of the domain. The score can be numeric or boolean. Higher scores indicate higher relevance.
+
+      - Pass both target_column and score_column. The value at the i-th position in the target list is the true relevance
+        score of the i-th element of the domain. The value at the i-th position in the score list is the predicted
+        relevance of the i-th element of the domain.
```
FelipeAdachi: this makes it sound like the values in the target_column and score_column should mean the same thing, but that is not true. For example, `target` could contain integers from 1-5, representing a relevance grade, and `score` could be a probability or confidence continuous value from 0-1. Instead of calling it "predicted relevance," we could use something more generic, like "model output," "decision measure," or something similar.

python/whylogs/experimental/api/logger/__init__.py
richard-rogers: There might be a better way to check, but the technique mentioned in statckflow applies to the full dataframe :(

FelipeAdachi (approved):
LGTM!

PR #1514, opened 2024-05-03
segmented ranking metrics

### LINKED ISSUE #1506: Ranking Metrics - Support segments
## Description

Currently, Ranking metrics don't support segments. We need to investigate why this isn't the case and, if possible, support it.

One reason (but possibly not the only reason) is the existence of non-mergeable metrics. Currently, the only non-mergeable metric is `accuracy_k_`.

richard-rogers:
## Description

Support segmented ranking metrics.

## Changes

- Segmented schemas are respected by the ranking metrics
- Added `sum_gain_k` to unit tests

Closes whylabs/whylogs#1506

### DIFF
--- a/python/whylogs/experimental/api/logger/__init__.py
+++ b/python/whylogs/experimental/api/logger/__init__.py
@@ -354,6 +354,7 @@ def log_batch_ranking_metrics(
     output_data["reciprocal_rank"] = mrr
 
     if schema and schema.segments:
+        original_columns = set(data.columns)
         for column in set(formatted_data.columns):
             if column not in original_columns:
                 formatted_data = formatted_data.drop(column, axis=1)
--- a/python/whylogs/experimental/api/logger/__init__.py
+++ b/python/whylogs/experimental/api/logger/__init__.py
@@ -139,7 +139,7 @@ def _get_segment_columns(schema: DatasetSchema, data: pd.DataFrame) -> List[str]
     columns: Set[str] = set()
     for partition_name, partition in schema.segments.items():
         if partition.filter:
-            return list(data.columns)  # I think filters can access any column?
+            raise ValueError("Filters are not supported for segmented ranking metrics")  # Filters are deprecated
         if partition.mapper:
             columns = columns.union(set(partition.mapper.col_names))
 
--- a/python/whylogs/experimental/api/logger/__init__.py
+++ b/python/whylogs/experimental/api/logger/__init__.py
@@ -354,7 +354,10 @@ def log_batch_ranking_metrics(
     output_data["reciprocal_rank"] = mrr
 
     if schema and schema.segments:
-        formatted_data = data.copy(deep=True)  # TODO: does this have to be deep?
+        for column in set(formatted_data.columns):
+            if column not in original_columns:
+                formatted_data = formatted_data.drop(column, axis=1)
+
         if log_full_data:
             return log(pandas=pd.concat([formatted_data, output_data], axis=1), schema=schema)
         else:
--- a/python/whylogs/experimental/api/logger/__init__.py
+++ b/python/whylogs/experimental/api/logger/__init__.py
@@ -366,14 +366,6 @@ def log_batch_ranking_metrics(
             return result
 
     result = log(pandas=output_data, schema=schema)
-    result = result.merge(
-        log(
-            row={
-                "accuracy" + ("_k_" + str(k) if k else ""): hit_ratio,
-            },
-            schema=schema,
-        )
-    )
     if log_full_data:
         result = result.merge(log(pandas=data, schema=schema))
     return result
--- a/python/whylogs/experimental/api/logger/__init__.py
+++ b/python/whylogs/experimental/api/logger/__init__.py
@@ -350,7 +350,6 @@ def log_batch_ranking_metrics(
         row_wise_functions.calculate_row_ndcg, args=(k,), axis=1
     )
     output_data[f"sum_gain_k_{k}"] = formatted_data.apply(row_wise_functions.sum_gains, args=(k,), axis=1)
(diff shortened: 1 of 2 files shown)


python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -135,6 +135,26 @@ def _all_strings(data: pd.Series) -> bool:
     return all([all([isinstance(y, str) for y in x]) for x in data])
 
 
+def _get_segment_columns(schema: DatasetSchema, data: pd.DataFrame) -> List[str]:
+    columns: Set[str] = set()
+    for partition_name, partition in schema.segments.items():
+        if partition.filter:
```
richard-rogers, line 141: It looks like filters can access any column. I'm unclear if they're actually supported, but I think this should be safe.

python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -333,6 +353,18 @@ def log_batch_ranking_metrics(
...
+        formatted_data = data.copy(deep=True)  # TODO: does this have to be deep?
+        if log_full_data:
+            return log(pandas=pd.concat([formatted_data, output_data], axis=1), schema=schema)
+        else:
+            segment_columns = _get_segment_columns(schema, formatted_data)
+            segmentable_data = formatted_data[segment_columns]
+            result = log(pandas=pd.concat([segmentable_data, output_data], axis=1), schema=schema)
+            result = _drop_non_output_columns(result, set(output_data.columns))
+            return result
+
```
richard-rogers, line 370: This skips `accuracy_k` for segmented schemas. I can drop it for the unsegmented case or in a follow-up PR.

natiska: @richard-rogers let's drop accuracy_k for both segmented and unsegmented profiles please.

python/whylogs/experimental/api/logger/__init__.py
jamie256: I think we should deprecate filter/query expression based segmentation. For ranking metrics segmentation support, since segment support in ranking metrics is new it would be better to throw an exception with message that filter segmentation is unsupported with ranking metrics, thoughts?

python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -330,18 +350,21 @@ def log_batch_ranking_metrics(
         row_wise_functions.calculate_row_ndcg, args=(k,), axis=1
     )
     output_data[f"sum_gain_k_{k}"] = formatted_data.apply(row_wise_functions.sum_gains, args=(k,), axis=1)
-    hit_ratio = formatted_data["count_at_k"].apply(lambda x: bool(x)).sum() / len(formatted_data)
     mrr = (1 / formatted_data["top_rank"]).replace([np.inf, np.nan], 0)
     output_data["reciprocal_rank"] = mrr
+
+    if schema and schema.segments:
+        formatted_data = data.copy(deep=True)  # TODO: does this have to be deep?
```
jamie256: Can you explain why are we copying the input data (again)?

python/whylogs/experimental/api/logger/__init__.py
richard-rogers: It's not hard to support them, but I'm happy to veto them if they're deprecated...

FelipeAdachi (dismissed):
LGTM!

python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -135,6 +135,26 @@ def _all_strings(data: pd.Series) -> bool:
     return all([all([isinstance(y, str) for y in x]) for x in data])
 
 
+def _get_segment_columns(schema: DatasetSchema, data: pd.DataFrame) -> List[str]:
+    columns: Set[str] = set()
+    for partition_name, partition in schema.segments.items():
+        if partition.filter:
+            return list(data.columns)  # I think filters can access any column?
```
richard-rogers:
```suggestion
            raise ValueError("Filters are not supported for segmented ranking metrics")  # Filters are deprecated
```

python/whylogs/experimental/api/logger/__init__.py
richard-rogers: The first copy has extra columns in it. Maybe I can drop them instead of copying.

richard-rogers:
```suggestion
        for column in set(formatted_data.columns):
            if column not in original_columns:
                formatted_data = formatted_data.drop(column, axis=1)

```

python/whylogs/experimental/api/logger/__init__.py
jamie256: yes, that looks better, thanks!

jamie256 (dismissed):
LGTM!

python/whylogs/experimental/api/logger/__init__.py
```diff
@@ -330,18 +350,24 @@ def log_batch_ranking_metrics(
         row_wise_functions.calculate_row_ndcg, args=(k,), axis=1
     )
     output_data[f"sum_gain_k_{k}"] = formatted_data.apply(row_wise_functions.sum_gains, args=(k,), axis=1)
-    hit_ratio = formatted_data["count_at_k"].apply(lambda x: bool(x)).sum() / len(formatted_data)
     mrr = (1 / formatted_data["top_rank"]).replace([np.inf, np.nan], 0)
     output_data["reciprocal_rank"] = mrr
+
+    if schema and schema.segments:
```
richard-rogers, line 356:
```suggestion
    if schema and schema.segments:
        original_columns = set(data.columns)
```

jamie256 (approved):
LGTM!

PR #1523, opened 2024-05-16
Thread rolling logger udf fix

naddeoa:
Apparently the logger layer is responsible for running udfs before processing data with a schema that wants segments. This updates the thread rolling logger's dataset profile abstraction to also run udfs when segments are active.


python/pyproject.toml
```diff
@@ -275,3 +277,22 @@ typeCheckingMode = "strict"
...
 reportMissingTypeArgumet = false
+
+[tool.ruff]
+line-length = 140
+indent-width = 4
+include = ["whylogs/api/logger/experimental/logger/**/*.py"]
+select = ["E", "F", "I", "W"]
+
+[tool.ruff.isort]
+known-first-party = ["whylogs", "langkit"]
```
jamie256: Do we ever reference langkit in whylogs? I think it would be a circular dependency.

jamie256 (dismissed):
Looks fine to me.

python/pyproject.toml
naddeoa: I'll get rid, was copy pasta

jamie256 (approved):
LGTM!

PR #1525, opened 2024-05-20
Fix SegmentedResultSet::set_dataset_timestamp()

richard-rogers:
## Description

`SegmentedResultSet::set_dataset_timestamp()` calls the `set_dataset_timestamp()` method on its contained segments. However, sometimes the segments are `DatasetProfileView`s rather than `DatasetProfile`s (seems to be the case with pyspark). `DatasetProfileView` is weird in that it has a `dataset_timestamp` property setter instead of a `set_dataset_timestamp()` method like everything else. This adds the method.

Perhaps we should deprecate the setter for API consistency?

## Changes

Adds `DatasetProfileView::set_dataset_timestamp()`

### DIFF
--- a/python/tests/api/logger/test_result_set.py
+++ b/python/tests/api/logger/test_result_set.py
@@ -61,7 +61,7 @@ def test_segmented_result_set_timestamp():
         assert results.view(segment).dataset_timestamp == timestamp
 
 
-def test_segmented_result_set_timestamp():
+def test_view_result_set_timestamp():
     results = ViewResultSet(DatasetProfileView(columns=dict(), dataset_timestamp=None, creation_timestamp=None))
     timestamp = datetime.now(tz=timezone.utc) - timedelta(days=1)
     results.set_dataset_timestamp(timestamp)
--- a/python/tests/api/logger/test_result_set.py
+++ b/python/tests/api/logger/test_result_set.py
@@ -39,6 +39,29 @@ def test_result_set_metadata_on_writables():
 
 
 def test_view_result_set_timestamp():
+    segment_column = "col1"
+    df = pd.DataFrame(data={segment_column: [1, 2]})
+    results: SegmentedResultSet = why.log(df, schema=DatasetSchema(segments=segment_on_column(segment_column)))
+    timestamp = datetime.now(tz=timezone.utc) - timedelta(days=1)
+    results.set_dataset_timestamp(timestamp)
+    segments = results.segments()
+    for segment in segments:
+        assert results.view(segment).dataset_timestamp == timestamp
+
+    # convert segment DatasetProfile to DatasetProfileView to test that case
+    first_partition = results.partitions[0]
+    segments = results.segments_in_partition(first_partition)
+    for segment_key in segments:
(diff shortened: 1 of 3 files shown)


python/tests/api/logger/test_result_set.py
```diff
@@ -39,6 +39,29 @@ def test_result_set_metadata_on_writables():
...
+        profile = segments[segment_key]
+        segments[segment_key] = profile.view()
+
+    results.set_dataset_timestamp(timestamp)
+    segments = results.segments()
+    for segment in segments:
+        assert results.view(segment).dataset_timestamp == timestamp
+
+
+def test_segmented_result_set_timestamp():
```
richard-rogers: Oops, I got the names backwards :)
```suggestion
def test_view_result_set_timestamp():
```

python/tests/api/logger/test_result_set.py
```diff
@@ -39,6 +39,29 @@ def test_result_set_metadata_on_writables():
 
 
 def test_view_result_set_timestamp():
```
richard-rogers:
```suggestion
def test_segmented_result_set_timestamp():
```

python/tests/api/logger/test_result_set.py
```diff
@@ -38,6 +38,29 @@ def test_result_set_metadata_on_writables():
...
+    for segment in segments:
+        assert results.view(segment).dataset_timestamp == timestamp
+
+    # convert segment DatasetProfile to DatasetProfileView to test that case
+    first_partition = results.partitions[0]
+    segments = results.segments_in_partition(first_partition)
+    for segment_key in segments:
+        profile = segments[segment_key]
+        segments[segment_key] = profile.view()
+
```
richard-rogers, line 57:
```suggestion

timestamp = datetime.now(tz=timezone.utc) - timedelta(days=2)
```

python/tests/api/logger/test_result_set.py
```diff
@@ -38,6 +38,30 @@ def test_result_set_metadata_on_writables():
...
+timestamp = datetime.now(tz=timezone.utc) - timedelta(days=2)
```
richard-rogers:
```suggestion
    timestamp = datetime.now(tz=timezone.utc) - timedelta(days=2)
```

catel22 (dismissed):
Looks good to me!

python/whylogs/core/view/dataset_profile_view.py
```diff
@@ -77,6 +77,12 @@ def dataset_timestamp(self, date: datetime) -> "DatasetProfileView":
         self._dataset_timestamp = date
         return self
 
+    def set_dataset_timestamp(self, dataset_timestamp: datetime) -> None:
+        if dataset_timestamp.tzinfo is None:
+            logger.warning("No timezone set in the datetime_timestamp object. Default to local timezone")
```
jamie256: can we call the ensure timezone function so that there is an error raised if there is no timezone?

jamie256 (dismissed):
LGTM!

catel22 (approved):
Looks good!

PR #1564, opened 2024-09-18
Add true bool count to ColumnCountsMetric

richard-rogers:
## Description

Include count of `True` bool values in `ColumnCountsMetric`

### DIFF
--- a/python/tests/core/metrics/test_column_metrics.py
+++ b/python/tests/core/metrics/test_column_metrics.py
@@ -112,3 +112,26 @@ def test_column_counts_with_and_without_nan_field():
     )
     assert full_column_counts.nan.value == 3
     assert full_column_counts.inf.value == 4
+
+
+def test_column_counts_true_count() -> None:
+    counts = ColumnCountsMetric.zero()
+    p_col = PreprocessedColumn.apply(["twelve", 12, True, False, True, "True", None])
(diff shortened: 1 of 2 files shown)


python/tests/core/metrics/test_column_metrics.py
```diff
@@ -112,3 +112,26 @@ def test_column_counts_with_and_without_nan_field():
...
+    assert deserialized.n.value == 7
+    assert deserialized.true.value == 2
+
+    # TODO: I haven't found a way to remove the counts.true attribute with the
+    #       ColumnCountsMetric frozen, but the following works at room temperature:
+    # delattr(counts, "true")
+    # buf = counts.to_protobuf()
+    # deserialized = ColumnCountsMetric.from_protobuf(buf)
+    # assert deserialized.n.value == 7
+    # assert deserialized.true.value == 0
```
jamie256: What about saving a profile/buffer file with the old code? Then you can make sure that your changes can handle the older profiles?

Remove/replace commented out code when you resolve this.

jamie256 (changes requested):
Changes for initial count tracking look good, but lets remove the commented out code, ideally replace it with a test using an artifact that contains the older code's serialized data.

jamie256 (approved):
LGTM!

PR #1566, opened 2024-09-24
Handle explicit segmentation keys

richard-rogers:
## Description

Allow `segment_key_values` in addition to segmented `DatasetSchema`

### DIFF
--- a/python/whylogs/api/logger/__init__.py
+++ b/python/whylogs/api/logger/__init__.py
@@ -165,6 +165,7 @@ def log_classification_metrics(
     schema: Optional[DatasetSchema] = None,
     log_full_data: bool = False,
     dataset_timestamp: Optional[datetime] = None,
+    # TODO: should we add segmente_key_values here?
 ) -> ResultSet:
     """
     Function to track metrics based on validation data.
--- a/python/whylogs/api/logger/__init__.py
+++ b/python/whylogs/api/logger/__init__.py
@@ -165,7 +165,7 @@ def log_classification_metrics(
     schema: Optional[DatasetSchema] = None,
     log_full_data: bool = False,
     dataset_timestamp: Optional[datetime] = None,
-    # TODO: should we add segmente_key_values here?
+    # TODO: should we add segment_key_values here?
 ) -> ResultSet:
     """
     Function to track metrics based on validation data.
--- a/python/whylogs/api/logger/__init__.py
+++ b/python/whylogs/api/logger/__init__.py
@@ -165,7 +165,6 @@ def log_classification_metrics(
     schema: Optional[DatasetSchema] = None,
     log_full_data: bool = False,
     dataset_timestamp: Optional[datetime] = None,
-    # TODO: should we add segment_key_values here?
 ) -> ResultSet:
     """
     Function to track metrics based on validation data.
--- a/python/whylogs/api/logger/logger.py
+++ b/python/whylogs/api/logger/logger.py
@@ -113,9 +113,6 @@ class Logger(ABC):
 
         # If segments are defined use segment_processing to return a SegmentedResultSet
         if active_schema and active_schema.segments:
-            if segment_key_values:
-                pass  # TODO: maybe check for duplicate keys
-
             segmented_results: SegmentedResultSet = segment_processing(
                 schema=active_schema,
                 obj=obj,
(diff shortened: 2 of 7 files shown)


python/whylogs/api/logger/logger.py
```diff
@@ -114,16 +114,15 @@ def log(
         # If segments are defined use segment_processing to return a SegmentedResultSet
         if active_schema and active_schema.segments:
             if segment_key_values:
-                raise ValueError(
-                    f"using explicit `segment_key_values` {segment_key_values} is not compatible "
-                    f"with segmentation also defined in the DatasetSchema: {active_schema.segments}"
-                )
+                pass  # TODO: maybe check for duplicate keys
```
richard-rogers: Better to do it in `_log_segment()`?

python/whylogs/api/logger/segment_processing.py
```diff
@@ -129,15 +131,19 @@ def _log_segment(
...
     segments: Dict[Segment, Any] = {}
     pandas, row = _pandas_or_dict(obj, pandas, row)
     if partition.filter:
+        # TODO: segment_key_values don't apply to filter since they're explicitely added
         pandas, row = _filter_inputs(partition.filter, pandas, row)
     if partition.simple:
         columns = partition.mapper.col_names if partition.mapper else None
         if columns:
-            _process_simple_partition(partition.id, schema, segments, columns, pandas, row, segment_cache)
+            _process_simple_partition(
```
richard-rogers, line 145: Should we allow empty columns if `segment_key_values` are specified?

python/whylogs/api/logger/logger.py
jamie256: flexible on where we check, it does seem better in `_log_segment`

python/whylogs/api/logger/segment_processing.py
jamie256: Its an edge case, not sure I understand how we get an empty column set with a simple segmentation partition, but if we do have an explicit segment via `segment_key_values` it would be good to handle this like the explicit segmentation route.

python/whylogs/api/logger/__init__.py
```diff
@@ -165,6 +165,7 @@ def log_classification_metrics(
     schema: Optional[DatasetSchema] = None,
     log_full_data: bool = False,
     dataset_timestamp: Optional[datetime] = None,
+    # TODO: should we add segment_key_values here?
```
jamie256: Yes, but can be in a separate PR.

python/whylogs/api/logger/segment_processing.py
```diff
@@ -129,15 +133,19 @@ def _log_segment(
     pandas: Optional[pd.DataFrame] = None,
     row: Optional[Mapping[str, Any]] = None,
     segment_cache: Optional[SegmentCache] = None,
+    segment_key_values: Optional[Dict[str, str]] = None,
 ) -> Dict[Segment, Any]:
     segments: Dict[Segment, Any] = {}
     pandas, row = _pandas_or_dict(obj, pandas, row)
     if partition.filter:
+        # TODO: segment_key_values don't apply to filter since they're explicitely added
```
richard-rogers: Should verify this assumption

python/whylogs/api/logger/__init__.py
richard-rogers:
```suggestion
```

python/whylogs/api/logger/segment_processing.py
richard-rogers: I knocked up some code to handle the case, but I'm inclined to think of this as a user error -- use an unsegmented schema and `segment_key_values` for this case. Maybe throw or log an error here?

jamie256: I'm ok with logging an error here, hinting on the correct usage pattern.

python/whylogs/api/logger/segment_processing.py
jamie256: The comment looks correct, can remove the TODO part of it

jamie256 (dismissed):
Looks good. Maybe cleanup some of the comments but the changes look right.

jamie256 (approved):
LGTM!

PR #1581, opened 2024-10-31
Remove size check on whylabs api keys

naddeoa:
This constraint isn't true. Keys can actually be arbitrarily long now that they include the org id as well, which is customizable.


richard-rogers (approved):

jamie256 (approved):
LGTM!

PR #1589, opened 2024-11-14
Use the v1 version of the log api

naddeoa:
This switches from the v0 to the v1 profile upload api, replacing the previous use of log_async in the whylabs client with log_profile. The apis are functionally the same but the dataset id and the org id are specified differently. The dataset id has to be supplied via a different name and ends up as a header in the request, while the org id is not specified at all because its derived from the api key.

In terms of implementation, I tried to do this all from outside of whylogs to avoid having to make this change but there are several issues that pop up when you try to pass in your own `WhylabsClient`, and especially your own `ApiClient` that made that impossible. The refactor looks daunting as well. 

Instead, this change just passes on the `WhylabsClient` to the other "sub writers" that are secretly made in the whylabs writer, and hard codes the use of the v1 api over the v0 api.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -16,7 +16,7 @@ whylogs-sketching = ">=3.4.1.dev3"
 protobuf = ">=3.19.4"
 importlib-metadata = { version = "<4.3", python = "<3.8" }
 typing-extensions = { version = ">=3.10", markers = "python_version < \"4\""}
-whylabs-client = "^0.6.5"
+whylabs-client = "^0.6.15"
 requests = "^2.27"
 backoff = "^2.2.1"
 platformdirs = "^3.5.0"
--- a/python/poetry.lock
+++ b/python/poetry.lock
@@ -4649,13 +4649,13 @@ watchdog = ["watchdog"]
 
 [[package]]
 name = "whylabs-client"
-version = "0.6.5"
+version = "0.6.15"
 description = "WhyLabs API client"
 optional = false
 python-versions = ">=3.6"
 files = [
-    {file = "whylabs-client-0.6.5.tar.gz", hash = "sha256:6a1a32847844cd45768c58441944592eb6d3ed69578dcc84cda9d7d6caea6cab"},
-    {file = "whylabs_client-0.6.5-py3-none-any.whl", hash = "sha256:6c686f6d5b7a1ae2c487a83c450c9ece46fe176784d5c5261ee840a0de9e45a5"},
+    {file = "whylabs-client-0.6.15.tar.gz", hash = "sha256:fab30b93a117ea3b2a1d84e1ed4ed4374c530a82fee791c5c7e427c76e8f0415"},
+    {file = "whylabs_client-0.6.15-py3-none-any.whl", hash = "sha256:c663cd4dcad94640c8fe96693a3c34dc30e2d9c123ff8f1c47b52c7f297190d3"},
 ]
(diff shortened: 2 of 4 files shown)


richard-rogers:
Some of the integration tests are failing. Investigating...

richard-rogers:
> Some of the integration tests are failing. Investigating...

Some just look like the mocks for injecting throttling need to be adjusted for V1.
More concerning, some fail to retrieve the uploaded profiles. That could just be the platform running slow or something more serious.

There are also some transaction test failures I haven't looked at yet.

richard-rogers:
There are integration tests failing on mainline too, so maybe some of this is a backend problem.

naddeoa:
@richard-rogers Things are all green now so I guess you reran them?

richard-rogers:
> @richard-rogers Things are all green now so I guess you reran them?

The CI doesn't run the integration tests. They're still failing here and on mainline.

richard-rogers:
> > @richard-rogers Things are all green now so I guess you reran them?
> 
> The CI doesn't run the integration tests. They're still failing here and on mainline.

Mainline CI is passing now. I updated the URI mocks, so now only a few transaction tests are failing.

naddeoa:
@richard-rogers can you post the failures or something so I can debug?

richard-rogers:
> @richard-rogers can you post the failures or something so I can debug?

```
FAILED tests/api/writer/test_whylabs_integration.py::test_transaction_aborted - AssertionError: assert '<ExceptionIn...ntextmanager>' == 'Transaction has been aborted'
FAILED tests/api/writer/test_whylabs_integration.py::test_old_transaction_context - AssertionError: assert 1 == 7
FAILED tests/api/writer/test_whylabs_integration.py::test_transaction_distributed - IndexError: list index out of range
```
I will investigate them today. You can run them by setting up the `WHYLABS_*` environment variables and running `poetry run pytest tests/api/writer/test_whylabs_integration.py --load`

richard-rogers:
> I will investigate them today. You can run them by setting up the `WHYLABS_*` environment variables and running `poetry run pytest tests/api/writer/test_whylabs_integration.py --load`

I think I figured it out. The `WhyLabsTransactionWriter` ctor used to always generate a new transaction id and stick it into its new `WhyLabsClient`. With this PR, the transaction writer is sharing the client with the parent writer, so the actual transaction id gets stomped on. I think there's a relatively easy fix, but I'd really like to clean up the writer code more in whylogs 2.0.

python/pyproject.toml
```diff
@@ -16,7 +16,7 @@ whylogs-sketching = ">=3.4.1.dev3"
 protobuf = ">=3.19.4"
 importlib-metadata = { version = "<4.3", python = "<3.8" }
 typing-extensions = { version = ">=3.10", markers = "python_version < \"4\""}
-whylabs-client = "^0.6.5"
+whylabs-client = "^0.6.15-dev0"
```
richard-rogers: Do we want to leave this as a dev dependency?

richard-rogers (dismissed):
Integration tests are green now

python/pyproject.toml
naddeoa: nope, I'll update this now and push again. Just a timing issue with the platform release

naddeoa: @richard-rogers just updated the version

richard-rogers (approved):

